PR #45, opened 2026-08-24
Remote access: a conversation that outlives the pipe, files that travel, and a road without ssh

santoshkumarradha:
## What this changes

A v3 conversation used to be married to an ssh pipe: the engine *was* `ssh … aforge engine`, so a closed laptop lid killed the turn in flight. This makes the conversation live on the machine that owns the work, with your terminal **attaching** to it.

Close the lid mid-answer, lose wifi, kill the terminal — the turn keeps running over there, the surface redials itself, and you land back in it with the gap replayed. Files you drop into the chat travel to that machine and come back. And a machine with no inbound ssh can be reached by a name and a pairing code instead.

**Testing this by hand: `docs/remote-access-testing.md`.** It is self-contained — every command, every string quoted as the code spells it, and a ranked list of what still needs work. Written for someone who has read none of the other docs.

## The UX, in one pass

**Connecting** is unchanged: `aforge chat --host devbox[:path]`. It is still the person's own ssh — their config, aliases, agent, jump hosts.

**A healthy connection says nothing about itself.** No badge, no icon, no "connected" word — the machine shows only as part of the place (`devbox:app`). That was already true and stays true.

**A dropped link** now draws one `connection` segment, and only while it is dropped:

```
reconnecting to devbox — trying for up to 5 minutes
```

It reconnects by itself and the answer continues rather than restarting. If the far machine can't keep a turn (a bare `aforge engine` on a pipe, no session host), it says so once instead of pretending:

```
devbox does not keep a turn running while nothing is attached, so the turn that was in flight did not survive the drop
```

**A question raised while nobody was attached waits**, and comes back with how long it waited: `this question has been waiting 4 hours` — and nothing at all under a minute.

**Files** — `/attach server.log`, tab-completing local paths. Bytes travel; the model is told the *path*, not the contents, so a 4MB CSV stays out of the context window until something reads it. 16MB per file, 32MB per message.

**Without ssh** — `aforge serve` on the machine prints a name and a code; `aforge chat --at otter-lamp-42` pairs once and opens thereafter. `aforge devices` lists and revokes, and revocation is always the engine machine's decision.

## How it was built

The wire was frozen first (`internal/remote/wire.go`, v1 → v2, one bump) and six lanes built against it in parallel: session host, roaming, attachments, relay, pairing, and the furrow seams. `internal/remote/loopback.go` — a real client against a real server over an in-memory pipe — is what let them start at once.

Version 2 adds exactly four things: `Frame.Seq` + `Hello.Resume` (replay the gap, not the conversation), `MethodDetach` (a closed window and a dead pipe stop being one event), file payloads both ways, and `HeldQuestion`. Every field is additive and `omitempty`; the door still refuses a mismatch, and must, because a v1 engine would silently interrupt a turn a v2 surface believed it had merely detached from.

## Testing

Beyond unit tests, `make test-remote` runs **three containers** — a scripted model, an engine with sshd, a surface — that share no path, no home and no credential, so a session that only works because both halves are one filesystem fails instead of passing by coincidence. No API key: the model is a scripted SSE server, so it costs nothing and does not flake.

| Scenario | |
|---|---|
| `HandshakeAndOneTurn` | PASS |
| `ThePathLaw` | PASS |
| `TheLinkDiesMidTurn` — ssh killed mid-reply, reply completes | PASS |
| `TwoSurfacesAtOnce` — second window gets the tail of a turn it did not start | PASS |
| `AnAttachmentLandsOverThere` | **SKIP** — `/attach` is a slash command, `--once` has no tray; wants a pty driver |

Full `internal/tui3` suite green (258s). Manual retrieval gate green three consecutive runs on the settled corpus. `internal/pair` and `internal/relay` green under `-race`.

## What needs work — please read before trusting this

1. **No relay is deployed**, and there is no default address. `--at` cannot work for a real person until ops runs one. It fails today with four distinct sentences naming what is actually wrong.
2. **The NAT claim is untested.** The relay exists so a machine with *no inbound reachability* can be reached — and the container harness cannot prove that, because containers on a bridge can reach each other. **This is the one claim in the wave still resting on design rather than a test.**
3. **`/attach` has no two-machine coverage** (scenario above skips). Both halves are unit-tested; nobody has proved the whole path end to end.
4. **`aforge serve` spawns `aforge engine` per connection** and does not use the session host, so persistence over `--at` is weaker than over `--host`. Check what `Welcome.Persistent` reports there before trusting it.
5. **No keychain, no Touch ID** — decided, not overlooked. The device key is a file at `~/.aforge/v3/remote/device.key` (0600), exactly the exposure of an ssh key with no passphrase. `pair.Keeper` is the seam; a test fails the build if the manual ever starts promising otherwise.
6. **`filippo.io/cpace` is vendored** at `internal/pair/cpace` — it was the only dependency never tagged, so it is ours now, bug included. Its header records what reading all 217 lines turned up. `flynn/noise` and `gtank/ristretto255` stay ordinary deps.
7. **`workspace_restore` writes files** and under a blanket allow would run unasked. It previews unless `confirm:true` and furrow seals state first, so it is undoable — but it is a candidate for `internal/approval`'s always-asks table. Undecided.
8. **An auxiliary call uses a hardcoded `model="nex-agi/nex-n2-mini"`** regardless of the configured model. Harmless against the stub and **not caused by this branch**, but somebody should find which call ignores the session's model.

## A prediction the build disproved

The plan expected held questions to switch harness building and adaptive runs back on over a connection. **They don't.** Those cards are announced on subscriptions this protocol has no door for, so they never cross at all — the empty room was never their blocker. Both design docs carry that correction visibly rather than quietly edited out, since the wrong guess is what put the item in the plan. Opening them means putting the card and its answer on the wire: a lane of its own, not in this wave.

## Docs

`docs/REMOTE.md` (architecture, nine decisions with rejected alternatives), `docs/remote-access-testing.md` (hands-on), `docs/remote-access-design.md` and `docs/remote-access-plan.md` (product and build order), plus five new manual pages the chat itself reads and a rewrite of `running-on-another-machine.md`, which described as impossible several things this wave ships.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HYEndv36eH44WRYyc3G6zi)_


santoshkumarradha:
## Verified by hand before merge, on linux/arm64

**What was checked, against a real model** (OpenRouter `deepseek/deepseek-v4-flash`) over a real ssh pipe (`--host localhost`, driven from tmux):

| | |
|---|---|
| headless turn over a connection | pass — engine runs over there, `pwd` returns the far path |
| a turn mid-`bash` when the link is killed | pass — `worked 1m03s · 1 tool call`, the reply completes, twice reproduced |
| the redial's landing | pass — **same** conversation, not a new one |
| the `connection` segment | pass — `reconnecting to localhost — trying for up to 5 minutes`, present only while the link is down |
| a message sent during the gap | pass — `submit failed: reconnecting to localhost — try that again in a moment`, visible, not lost |
| `/attach` over the wire | pass — tray, bytes in the far session's `attachments/`, model reads it by path |
| `aforge devices` / `--at` / `serve` with no relay | pass — all four refusals word-for-word as documented |
| `make test-remote` | 4 pass, 1 skip, 37s |
| unit suites + `internal/tui3` + `cmd/aforge` + `internal/session` | green |

### Three defects found and fixed on the merge

1. **The container harness could not run on this machine.** `buildStatic` pinned `GOARCH=amd64`, so on an arm64 host docker started arm64 containers and the scripted model died with `Exec format error` — a failure that reads like a missing file. It follows the host now (`AFORGE_E2E_ARCH` overrides). The suite went from FAIL to 4 pass / 1 skip.
2. **A refused handshake said its sentence twice.** `aforge chat --host devbox:/nowhere` printed `error: engine: open /nowhere…` once from the engine's stderr (which ssh puts on the person's terminal) and once from the surface reading the refusal off the wire. `remote.Refusal` is typed now and the engine door exits quietly on one; exit code still 1.
3. **The usage text advertised v2's `--session id|new`.** Typing that against v3 writes a transcript into a file literally called `new` in the workspace. The line now says what the flag takes.

### Merge conflict

One: `internal/manual/chat/screen.md`'s segment table — this branch added `connection`, the base added `crew`. Both kept, renumbered against the real order in `render.go`'s `hudSeg`, count corrected to twelve.

### One regression the merge itself caused

Two manual retrieval probes broke on the **merge**, not on either side alone — the wave's five new pages out-scored the pages "can I run this on my dev box over ssh" and "nothing happens for a minute after a run starts" belong to. Fixed per the manual law: the headings take the asker's words, never the test. Gate green three consecutive runs.

### Docs

`docs/remote-access-testing.md` gains **§3.0 — without a second machine**: the `--host localhost` recipe, the tmux driver, how to kill the link by pid, and what that run actually proved, with the honest limit (localhost shares a disk, so it cannot prove the bytes had to travel). `internal/manual/chat/running-on-another-machine.md` learns the same thing, so the chat can answer it. `CLAUDE.md` gains the remote suites.

The "what still needs work" list is updated rather than trimmed: no relay deployed, the NAT claim still untested, and `/attach` still has no automated two-machine coverage.

PR #93, opened 2026-08-31
fixes: a hint is a command with evidence behind it, or it is silence

santoshkumarradha:
## What was measured

On a real session on 2026-08-31 the project's fix store
(`~/.aforge/v3/projects/…/tasks/fixes.json`) read `asked: 17, found: 3,
worked: 0` — and the three things it had found were these:

| the failure | the line the model was handed |
| --- | --- |
| `grep` — `Path not found: …/test_execution_logger.py` | `what ran next and it went away: /\/+$` |
| `bash` — `npm error Missing script: build` | `what ran next and it went away: git log --oneline -5` |
| `bash` — `<path>: no such file or directory` | `what ran next and it went away: gh` |

The first is not a command at all. It is the **regex the next grep searched
with**, filed as a remedy because a pattern is what says what a grep call did
(`glossField`). The second and third are commands, and they fixed nothing —
the model looked at the log, then reached for `gh`, and the store watched two
adjacent events and wrote down a cause. A wrong hint costs a model a step
chasing it, which is strictly worse than the silence it replaced.

## The gate

Two questions, in a new `internal/session/fixremedy.go`, asked of every entry
on the way out (`fixEntry.worthSaying`) **and** on the way in
(`episode.noteSuccess`), so junk stops entering the two files rather than being
filtered out of them forever:

- **Could anybody run this?** The first word has to name a program this machine
  has, resolved through `exec.LookPath` — the same question the shell asks
  first. A word with a slash in it is resolved as a path and must be an
  executable file, so `/\/+$` and `internal/session/loop.go` both fail. Known
  costs, both silence rather than a wrong line: a patch opening with a shell
  builtin (`cd /tmp/x && npm test`) or an environment assignment
  (`CGO_ENABLED=0 go build`) is not offered.
- **Did the world say it twice?** `fixMinPairings = 2`, or
  `fixMinWorkedOffers = 1` — a patch that was offered, taken, and made the
  error go away needs no second sighting, because that observation is a cure
  rather than an adjacency. Both are named constants carrying their reasoning.

`commandLike` and `approval.Vouchable` are deliberately **not** reused: both
refuse a composed line, because both answer a safety question (may a standing
approval speak for this, may an auditor run it from clean), and
`make clean && make build` is the archetypal fix this whole sidecar exists for.
What is asked is the smaller question those readings both open with.

## grep and find lose their lanes

Their defining argument is a search pattern, which is the same objection that
already keeps `read`, `write`, `edit` and `ls` off the belt one step along: a
pattern is what the model was LOOKING FOR, not a thing to do about a failure.
No amount of counting the `/\/+$` entry better could have made it right. `bash`
is what is left, and its argument is a command.

## What I found about the worked signal

**It is wired, end to end, and nothing was built for it here.** `noteFailure`
blames the offered patch when the same signature comes back; `noteSuccess`
calls `confirmAdvised` and `store.outcome(true)` when the patch that succeeded
is the exact patch that was offered; `fixShelf.blame` and `fixStore.outcome`
carry both halves into the entry and into the store's own counters.

`worked: 0` is therefore not a missing half. It is the loop reporting honestly
that **nothing it ever offered was taken** — which is exactly what an offer of
`/\/+$` earns. The strictness is real (a `worked` needs byte-identical text),
but loosening it would be inventing a similarity judgement, which is a project
of its own and not this one.

## Tests

Every row of the measured store is now a test (`fixremedy_test.go`):
`/\/+$` planted 9 times is never offered; `git log --oneline -5` on the npm
error is silent at one sighting and produces the hint **verbatim in the
existing wording** at two; a program this machine lacks is never offered; a
store already full of junk stops saying it through the live belt.

Existing tests that recorded a pairing once and expected the line now watch it
twice, and the placeholder patches in them (`the fix`, `the project answer`)
became real commands. **The wording of both lines is untouched** — what changes
is how often either is said at all.

```
go build ./... && go vet ./...          clean
go test ./internal/session/ -count=1    ok  105.352s
go test ./internal/manual/ -count=1     ok  (the corpus gate)
go test ./internal/tui3/ -run Manual    ok
```

## Manual

`internal/manual/chat/what-i-remember.md` gains a section — "When does it say
nothing, and why did a suggestion stop appearing?" — carrying both gates and
both measured lines, and the claim that `grep` and `find` are remembered is
gone. The pack is rebuilt.

## Seam

`fixCleanPatch` is untouched, so the one-line `redact.Secrets` call coming from
`feat/task-ground` (#84) lands on it without a conflict.


santoshkumarradha:
`origin/dev` merged in (#90 landed while this was in flight). The only conflict was `internal/manual/chat.pack.gz`, resolved with `go generate ./internal/manual`. `fixCleanPatch` came across untouched, so the `redact.Secrets` call and this gate sit side by side with no overlap.

Re-run on the merged tree:

```
go build ./... && go vet ./internal/session/ ./internal/manual/   clean
go run ./cmd/aforge-changes check    docs/changes/unreleased: 2 entries, all well formed.
go test ./internal/manual/ -count=1  ok  0.427s
go test ./internal/session/ -count=1 FAIL — TestAWokenTurnIsMeteredExactlyLikeATypedOne only
```

That one failure is **not from this branch**: it fails identically at `origin/dev` in a clean detached worktree (`TempDir RemoveAll cleanup: … directory not empty`, and a task-checkpoint rename returning `invalid argument`). It arrived with #90. Every other test in the package, including all of `fixremedy_test.go`, passes.

PR #105, opened 2026-08-31
A conversation switcher: ctrl+k over a dimmed screen, and families folded on the task page

### LINKED ISSUE #94: A conversation ring: alt+tab for the eight conversations this process already holds
## What is missing

This process already holds up to eight conversations alive at once (`internal/tui3/keeper.go`), but the only way between them is `tab`, which goes to **one** — the last one — and says nothing about the other six. To reach a third conversation you leave the chat for home, read a list, and come back. That is a page transition for a gesture people fire fifty times a day.

## What to build

**A conversation ring: a hold-and-cycle overlay, in the shape of alt+tab, over whatever you are looking at.**

- A card in the middle of the frame, with the surface behind it **dimmed** rather than covered — the depth ladder this surface already has (`depthfade.go`'s `palette.fade`), which is exactly how the composer layer behaves (`composerlayer.go`, SCREEN 2e). You never lose your place.
- One row per **open conversation**, most-recently-in-front first, with the one you are standing in printed last and marked `you are here`, so the ring has a visible seam.
- Every row says **what changed since you last looked** — `it is asking you something`, `3 tasks running`, `it finished while you were away`, `nothing new` — rather than what the conversation is. That makes the switcher double as the catch-up, which is the thing you actually want after twenty minutes in one chat.
- It must be **snappy**: built entirely from what this process already holds in memory — the keeper's map and the agents' own predicates. No disk, no world scan, no presence read, no remote call on the keystroke.

### Only main sessions in the ring

The ring lists **conversations**, never the tasks running inside them. A conversation with six nodes under it is one row, and the count of what it is running is a clause on that row. Home already reads this way and stays that way.

### Child tasks belong on the tasks page, folded

The tasks place (`ctrl+.`) draws every row flat today, so a family of eight arrives as eight peers of everything else. It should draw the **family root** and fold its children under it, expanded on demand — the shape the roster column already uses (`task.go`'s `railGrow`), with the same connectors. Collapsed is the default; a page that opens with every child expanded is the clutter this is fixing.

## The key

The chord has to be free in a terminal, unclaimed by the OS or the emulator, and mean something.

- **`ctrl+k` is the binding.** It has a legacy encoding, so it arrives in every terminal with no protocol negotiation at all; no OS steals it; no common emulator binds it; and "jump to a conversation" is what `ctrl+k`/`cmd+k` already means in Slack, VS Code, Linear and Notion. Nothing on this surface binds it except the harness-design approval row, which is a modal that owns the keyboard while it is up.
- **`ctrl+tab` is its alias, and only where the terminal says it can spell it** (`app.keysDisambiguated`) — the same law `ctrl+1`…`ctrl+7` already follows in `chords.go`. `ctrl+tab` is the muscle memory, but it has no legacy encoding, and WezTerm and Windows Terminal both bind it to their own tabs by default, so it can never be the only way in.
- **`alt+tab` is not available at any price** — the window manager takes it on Windows and on most Linux desktops.

Inside the ring: `tab`/`↓` down, `shift+tab`/`↑` up, `enter` goes, `esc` puts it away and leaves you exactly where you were. `tab` on an empty box keeps its current meaning — straight to the last conversation, no overlay — so the two-conversation flick costs one key as it does today.

## Done means

- The ring opens in the conversation and on every place, from state already in memory.
- The surface behind is dimmed, not replaced.
- The tasks page folds families.
- `internal/manual/chat/` says all of it — the keys page, the chord table, and the tasks page's own account of the fold — and the three manual gates pass.
- A change entry in `docs/changes/unreleased/`.

santoshkumarradha:
Closes #94.

## What this is

`ctrl+k` draws SCREEN 3b's card in the middle of whatever you are looking at, with the screen behind it dimmed. It lists **the conversations this terminal has open**; everything else on the machine is behind the fold at its foot.

```
╭──────────────────────────────────────────────────────────────────────────────╮
│  open              3 of 12 · tab down · shift+tab up · enter go · esc back   │
│                                                                              │
│  1 ? harness dry run on one pub…  asking you something      aforge-v2    4m  │
│  2 ◐ openrouter price scrape      2 tasks running            research    1d  │
│  3 ○ Refactor the rail scope mo…  you are here              aforge-v2   40m  │
│  ▸ 9 more on this machine · → reach them                                     │
╰──────────────────────────────────────────────────────────────────────────────╯
```

| Key | |
| --- | --- |
| `ctrl+k` / `ctrl+tab` | open, and step down |
| `ctrl+shift+k` / `ctrl+shift+tab` | step up — where the terminal can spell them; `shift+tab` does it everywhere |
| `tab` `shift+tab` `↑` `↓` `1`…`9` | walk and take |
| `→` `←` | open and shut the fold |
| `ctrl+w` | close that conversation |
| `enter` `esc` | go, or leave exactly where you were |

Verified in a real terminal against `make demo-home` at every step, not only in the rig.

**Read the commits in order — each one is a defect the running build showed:**

1. the switcher and the tasks-page family fold;
2. it listed only what was already open, so on a fresh launch the key did nothing and was named nowhere — the feature was invisible until you had learned the thing it exists for;
3. it was a band of text rather than the drawing's card, open and closed rows were indistinguishable, and there was no way to close anything.

## The three decisions worth reviewing

**1. `ctrl+k`, with `ctrl+tab` as an alias only where the terminal can spell it.**

`ctrl+tab` is the muscle memory and it cannot be the way in: it has no legacy encoding, so it arrives as a bare `tab` unless the terminal took the kitty keyboard protocol's disambiguation flag — and WezTerm and Windows Terminal both bind it to their own tabs by default, so it never reaches us there at all. It is bound where `app.keysDisambiguated` is true and **never named where it would not be delivered**, which is the law `ctrl+1`…`ctrl+7` already follows in `chords.go`. `alt+tab` belongs to the window manager on Windows and most Linux desktops and was never available.

`ctrl+k` has a legacy encoding (0x0B), no OS or emulator wants it, and it is already "jump to a conversation" in Slack and the switcher in VS Code, Linear and Notion. The one other binding on this surface is the harness design's approval row, which is a modal that owns the keyboard while it is up and is read first — the same arrangement `ctrl+.` already has.

**2. The open half comes from memory; the rest is one reading on the keystroke.**

Above the rule: the keeper's map, the sidecar each detach left, and three predicates asked of agent pointers this process already holds — no disk at all. Below it: one world reading, taken **on the keystroke and never on a frame**, which is affordable because this gesture *replaces* pressing `space space` — home takes the same reading and then draws a whole page with it, so the card can be no slower than what people do today to answer the same question.

`worldSeam` is that reading's three inputs handed in, so a command can take it off the loop without a second copy of `worldOf`'s branching. `hopKnown` — the count the legend reads on every frame to decide whether to name the key — is refreshed by a command at boot, never by a walk on the paint path.

The key's guard (`hopMayOpen`) is deliberately looser than the advertisement's (`hopAvailable`): the count lands a moment after boot, and a key gated on it would do nothing for the first frames of a session whose card would plainly have had rows. It errs the way the capability law cares about — a key that works slightly before it is named, never the reverse.

The rows are **frozen the moment the card opens**. A conversation finishing a turn stirs the surface, and a list that re-ranked on that stir would move the row under the cursor between the keystroke that aimed at it and the `enter` that took it.

**3. The dim is the layer, and the box is the one outlined thing here.**

The body behind is repainted at the faintest stop of the depth ladder through `composerFade`, the same function SCREEN 2e's composer layer uses, so the two layers can never end up at two different depths. The roster column dims with it. `hopOver` replaces rows in a body that was going to be laid out anyway, so the frame is the same height it was before the key was pressed and no place, room or transcript has a line about being underneath one.

The border is a **ruled exception** to the house rule that nothing on this surface is outlined. It earns it by being the only panel that floats in the middle of a page rather than taking the frame or hanging off an edge: without an edge of its own the eye reads it as text that happens to be brighter. Six columns of inset each side, two of air inside, and the fixed subject column is why it scans.

**4. `ctrl+w` closes, and closing is not switching.**

The grooming called it "put away without stopping it"; that is not true of this engine, so the card does not say it. Closing ends the agent and work inside it stops. A quiet conversation closes on one press; a busy one takes two, with `2 tasks running · ctrl+w again to close it anyway` in between; walking the cursor disarms it. The card stays up and says `closed · <name>`, because tidying up is something people do two or three of in a row.

## Only main sessions in the switcher, families on the tasks page

The card lists **conversations**, never the tasks inside them — a conversation with six nodes under it is one row and the count is a clause on it.

The tasks place (`ctrl+.`) now catches up with the roster column: a run that split into eight workers was eight peers of everything else on the page, and is one root with the family folded under it — **shut**, saying `+3 under`, opened with `→` and shut with `←`. A second `→` on an open family falls through to the row's verbs, so a running root keeps its `stop it`. The two-cell family column appears only where there is a family on the page, so nothing moves sideways on a machine that has never split work up. A worker whose root is filed in another section stands on its own rather than hiding under a parent that is not there.

## Manual and tests

`internal/manual/chat/keys.md` gains four sections (the switcher, what each row says, why `ctrl+k` and not `ctrl+tab`, and `tab` unchanged); `tasks.md` gains the family fold. The keys page said `ctrl+k` was **"Not bound anywhere else"** — that denial is removed in this change, per the manual law's hunt-the-old-denial clause. Eleven new probes in `internal/manual/chat_test.go` in people's own words ("alt tab between conversations", "what did my other chats do while I was away", "what does +3 under mean").

Sixteen new tests: the reading and its order, the walk and the wrap, `esc` changing nothing, the dim, the freeze, the note column, the capability law on both the key and its advertisement, and the family fold shut, open, and with a child whose root is elsewhere.

`go build ./...`, `go vet ./...`, `go test ./internal/tui3/ ./internal/manual/` all green.

## Try it

```sh
git fetch origin && git checkout feat/conversation-ring && make build
make demo-home DEMO_HOME=/tmp/aforge-demo
```

Then just press **`ctrl+k`** — from the first frame, in the chat or on any place. The legend above the box names it whenever it would act. `→` for the rest of the machine, `ctrl+w` to close one.

https://claude.ai/code/session_01EsHoUNFe8RxMiiSvMcjTwi


santoshkumarradha:
**Quick switch landed** (06787d70) — answering the release-to-commit question from grooming.

True `release ctrl → go` needs kitty-protocol key-release reporting: absent behind tmux, iTerm2, Terminal.app, Windows Terminal, and already ruled against in app.go. Instead this ships the Chrome model — **each press switches immediately**, so there is nothing left to commit on release:

- `ctrl+k` lands you in the previous conversation at once; keep tapping to go older; `ctrl+shift+k` cycles back (with the card down it enters at the ring's far end)
- the card is a receipt that fades ~1s after the last press; `esc` takes the burst back; `tab` afterwards returns to where the burst started (MRU resealed)
- typing rides straight through the fading card into the draft — zero lost keystrokes
- any non-chord key (arrows, fold, `ctrl+w`) converts it to the holding card, which never fades and waits for `enter`
- setting: `quick switch` (interface, `ui.quick_switch`), **default on**; off = the holding card everywhere

Verified in a real tmux terminal: instant switch, marker + note handoff, self-fade, wrap cycling, esc undo, type-through. Full tui3 + config + manual suites green.

PR #167, opened 2026-09-01
UI feature branch: chat chrome — top bar, bottom fold, and the ongoing UI wave

### LINKED ISSUE #126: chat chrome: one top bar for chat and task, and the bottom folded to two lines
# The chat chrome: one top bar for chat and task, and the bottom folded to two lines

This is a full redesign of the v3 chat's frame — the rows above and below the transcript.
It is one issue on purpose: the top bar and the bottom fold are one geometry change and
splitting them would leave the frame half-moved. Everything here is `internal/tui3`.

## The doctrine (read this before the spec)

The physics of a chat TUI: text streams in at the bottom, the cursor lives at the bottom,
so **foveal vision is the bottom third and the top line is peripheral**. The top is read at
three moments only — arriving, switching into a task, and glancing up to ask "which chat is
this?" The bottom is seen on every keystroke.

So the two bars split by tempo, and this is the law of the whole change:

- **Top bar = slow facts.** Things true of the whole conversation that change only by a
  deliberate act: where you are (the breadcrumb), the model and its effort, the branch, the
  host, the safety posture. A change up there IS a navigation event, so the bar doubles as
  confirmation that the navigation happened.
- **Bottom line = fast facts.** Things that tick: cost accruing, context filling, the
  open/want-you count, the working clock. Plus the modal hint ladder, because a hint's job
  is to inform the next keystroke and the next keystroke happens at the composer.

**Nothing that ticks goes on top; nothing permanent goes on the bottom.**

Three display strengths, and every fact is assigned exactly one:

1. **Ambient** (the two bars) — the few numbers that create continuous trust.
2. **Receipts** (rows in the transcript at landing moments) — where the impressive numbers
   live. A number in a static bar becomes wallpaper; a number attached to a thing that just
   finished is evidence. (Receipts already exist; this issue does not add new ones — it
   demotes ambient trivia so the receipts can carry the story.)
3. **Sheets** (`/status`, `/cost`, the status sheet, the Spending tab) — everything, deep,
   one press away. Every ambient number is a **door** to its sheet; the bar is an index,
   not a report.

## What exists today (the inventory being reorganized)

- The chat has **no top line at all** — the conversation starts at row 0 (`view.go:856`).
- Bottom chrome is two full-width rows: the **legend** (input's top border,
  `render.go:2658`: `─ host · branch* ──── hint/keys ─`, with a 25-branch modal hint ladder
  at `render.go:2975`) and the **status row** (`render.go:1435`: identity cluster
  `session · model:level · via server · tok/s` left; telemetry right in fixed order —
  crew, open, jobs/watches, keeping-an-eye, delta, cost, context+sparkline, cache, burn,
  compaction ETA, YOLO, host link, state word — with `dropOrder` at `render.go:2007`).
- A **task strip** of chips appears above the transcript while work runs (`taskstrip.go`).
- Inside a task the room pins an accent header (`roomHead`, `room.go:2458`):
  `─ ⠙ main ▸ title · state · clock · $ · model · thinking high ─── esc/← main · ✕ ─`,
  plus up to 3 dim kin rows (`roomKinRows`, `room.go:2628`).
- Phone tier (<60 cols) draws a 2-row **deck** (`statusdeck.go`) and a fullscreen sheet.
- Doors today: model name → picker (`modelSpan`, `app.go:5468`; in a room it retargets that
  node, `room.go:188`), `$` → Spending tab (`moneydoor.go`), `◦ keeping an eye on N` →
  standing page (`standdoor.go`). YOLO has **no** door today.

### Top bar — one row + hairline rule, always present (chat and task alike)

```
· aforge-v2 › fix the parser                      glm-5.3:high · main* · devbox · YOLO
──────────────────────────────────────────────────────────────────────────────────────
```

Inside a task the same row deepens — the room header stops being a separate design and
becomes the top bar's task form:

```
⠙ aforge-v2 › fix the parser › port the lexer #2 · working · 2m12s · $0.04   esc/← back · ✕
──────────────────────────────────────────────────────────────────────────────────────
```

**Left cluster (the crumb):**
- Lead glyph column, 2ch: `·` in a chat at rest; in a task, the node's state glyph
  (spinner / ✓ / ✗ / queued — reuse `roomMark`, `room.go:2724`), unpainted, exactly as
  the room header draws it today.
- Crumb segments joined by ` › `. Parents in **dim**, the current segment in **ink**.
  In a chat: `project › chat name`. In a task: `project › chat name › task title #N`.
  The task's handle (`#N`, the roster's number) rides in dim after the title.
- In a task, after the crumb: ` · state word · clock · $spend` — the room header's
  existing facts (`roomStateWord` `room.go:2778`, `roomClock`, `roomSpend`), each dropped
  when unpublished, exactly the current emptiness behavior. The node's model does NOT
  repeat here — the right cluster already shows the model in force, and inside a room
  that IS the node's model (see below).
- The whole left cluster paints **accent** while a room is open — the existing law
  (`render.go:1456`) moves up with it. In a chat at rest it is ink/dim, never accent.

**Right cluster (the terms):**
- `model[:effort]` — the conversation's model, basename via `modelBase`, with the
  reasoning rider spliced exactly as today (`view.go:656`). Inside a room this becomes
  `task <node model basename>` and retargets that node on press, preserving
  `roomModelMovable` (`room.go:164`): unpressable once the node can no longer be moved.
- `branch*` — `branchWord` (`render.go:2797`), dirty star included. Host branch when
  hosted, as the legend does today.
- `host` — only when the session is hosted (`--host`); the reconnect note
  (`reconnecting to devbox — trying for up to 5 minutes`) **outranks everything else in
  the right cluster** and may evict all of it, as `segLink` does today (`hostlink.go:191`).
- `YOLO` — only when approvals are `allow` (`render.go:2392`), painted `bad`. Never
  dropped by the width ladder.
- In a task, the right cluster ends with ` · esc/← back` and the ` · ✕` stop mark —
  the room's existing back door and stop door move here unchanged (`roomBackWord`,
  `stop.go:574`), with the ✕ outliving the microcopy at narrow widths as it does now
  (`room.go:2483`).

**What the top bar never carries:** cost of the chat, context %, counts, rates, state
spinners for the chat itself — those tick, and ticking facts live at the bottom. (A task's
own clock/spend in the task form is the one exception: that cluster is the room header's
existing contract, it is scoped to the thing you are looking at, and it dies with the room.)

### Bottom — exactly two rows plus the composer

```
── <modal hint ladder, unchanged> ────────────────────────────────────────────────────
› say what you want done                                                          high
2 jobs · ◦ keeping an eye on 2        8 open · 2 want you · $1.04 · 8% · ⠹ working · 4s
```

- **The composer's top rule keeps the legend's hint machinery whole.** The 25-branch
  `hintWord` ladder (`render.go:2975`) — consent `y allow · n deny · a always`, task
  proposal, pickers, rewind, copy mode — is the interface at those moments, not
  documentation. It does not move to `?`. What LEAVES the rule: `host · branch` (moved
  to the top bar). The generic key advertising (`space space home · tab last · / commands`)
  stays but decays: shown while the earned-tips machinery would show it, quiet after.
- **The status row keeps only what ticks**, right-aligned cluster in fixed slot order:
  1. `N open · M want you` — the parallelism fact, `segOpen`'s successor; the `want you`
     count paints `warn`-adjacent per the existing ask/warn rules.
  2. `$X.XX` — this chat's spend. **The `$0.00` stability exception dies**: with the slow
     facts gone from this line, segments no longer jump, so zero renders as nothing
     (emptiness law regains the line; update the law's stated exception in CLAUDE.md and
     the manual page in the same change).
  3. `N%` — context as percent only. The full `82.5k/1M` form appears automatically at
     ≥80% and in the sheet//status; pressing the span prints `/status` into the transcript.
     Sparkline (`render.go:2234`) is demoted to the sheet only.
  4. `compaction in ~N turns` — unchanged condition (`render.go:2350`); splices in only
     when ≤5 turns out.
  5. State word, last, never dropped: `⠹ working · 4s` while working, and the exceptional
     words (`waiting · your call`, `stopping`, `interrupted`, copy-mode words, drag
     receipts). **`idle` is deleted** — an idle chat's status row simply ends at the
     numbers. The working clock is the engine being audible; idle said nothing.
  - Left, dim, the presence clauses nothing else shows: `N jobs · N watch` (`segAmbient`)
    and `◦ keeping an eye on N` (`segKeeping`, still breathing, still a door).
- **Demoted to `/status` + the sheet only** (delete from the ambient row): cache savings
  (`segCache`), burn rate (`segBurn`) — note: the state word's clock stays, the tok/s
  figure goes — delta (`segDelta`), the served rider (`via server · tok/s`,
  `render.go:1786`), and crew (`segCrew`). `/status` and `/cost` output is unchanged.
- **The task strip is removed** (`taskstrip.go` and its tests): the top bar signals the
  place, the roster/rail carries the running set, the strip was the old top-of-screen
  presence and is now redundant. The `+N` roster door it provided is already covered by
  `ctrl+t`/`ctrl+g` and the rail.
- **Kin rows are removed** (`roomKinRows`): the crumb carries the parent; spawned
  children are visible in the room body and the rail.
- `statusQuiet` (welcome open) now means: top bar shows the crumb only (no right
  cluster), bottom status row is empty. Same spirit as today (`render.go:1941`).

## Interaction audit — every door, its hover, its press

The hover law (`hover.go`) governs all of these: **the set that lights is exactly the set
the press acts on**, each door lights at the size of the thing (its span, not its row),
and a span that answers to nothing does not react. Every door below registers a
`hudSpan` + row mark at layout time, the `moneydoor.go`/`standdoor.go` pattern.

| Element | Hover | Press | Notes |
| --- | --- | --- | --- |
| model segment (top bar) | brightens one step (existing `paintIdentity` behavior) | opens the `/model` picker | keyboard parity: `/model`. In a room: retargets that node; **unpressable (no span, no hover) when `roomModelMovable` is false** |
| YOLO | brightens | opens Settings on the **Safety** tab | NEW door. Keyboard parity: `/settings`. This is a safety affordance: seeing the gate open must offer the way to close it |
| `$` (bottom) | brightens | Spending tab | existing `moneydoor.go`, span moves rows |
| `N%` ctx | brightens | prints `/status` into the transcript | NEW door; the terminal translation of "tokens on hover" |
| `◦ keeping an eye on N` | accent on hover (existing) | standing page | existing `standdoor.go` |
| `N open · M want you` | brightens | opens the conversations list (the `tab`-door's list form / home's open band) | NEW door; keyboard parity: `tab` |
| crumb: `project` | brightens | home (`space space` parity) | |
| crumb: chat name (in a task) | brightens | leaves the room — `closeRoom`, scroll restored | replaces today's whole-header back press with a precise span; the ` esc/← back` word remains a second, wider target |
| crumb: current/last segment | no hover, no press | — | you are already there; a door to nowhere must not light |
| `✕` (task form) | brightens | stop the work | existing `stopMarkPress`; claimed one rung before the back word, as today (`stop.go:551`) |
| deck rows (phone) | unchanged | unchanged | the deck IS the phone form of these bars |

Keyboard: nothing changes. `esc` in a room leaves (never stops), `←`/`→`/`←←` navigation,
`ctrl+t`/`ctrl+g` roster, `ctrl+v` rung, `ctrl+l` jump, `x` stop, `space space`, `tab` —
all exactly as today. Every door above must have the keyboard path listed beside it,
because the pointer is optional on this surface.

## Breadcrumbs at depth

`roomPath` (`room.go:2769`) is one level deep today but written over a path; the run page
already draws `main ▸ goal ▸ n8 ▸ transcript` (`roomorch.go:1900`). One shared truncation
function for both, with this give-way ladder (applied before the right cluster's ladder —
the crumb is the thing being confirmed, it yields last):

1. Full: `project › chat name › parent task › this task #2.1`.
2. Middle-collapse: keep the first and the last two segments, fold the middle to `…`:
   `project › … › parent task › this task #2.1`.
3. Drop the project: `… › parent task › this task #2.1`.
4. Drop the parent: `… › this task #2.1`.
5. Truncate the title itself with `…`, floor 12ch (today's `roomHeadFloor`).

The current segment is never sacrificed for a parent. Handles stay attached to their
titles through every rung. `esc` climbs exactly one crumb level (a sub-task's esc goes to
its parent task, not the chat) — matching the run page's existing behavior.

## The responsive ladder

Width tiers are the existing four (`layoutTier`, `view.go:1141`): wide ≥120,
standard ≥80, narrow ≥60, phone <60. Heights: the top bar (like today's room header)
requires `breathingRows() > 0`; on brutally short windows the transcript wins and the top
bar is the first chrome to go — the bottom two rows are last.

**Top bar right cluster, drop order (first dropped → last):**
1. host latency detail (name survives) → 2. branch → 3. `:effort` rider → 4. model.
YOLO and a reconnect note are never dropped. In the task form, the state/clock/$ cluster
drops right-to-left ($ first, glyph last) before the right cluster gives anything up —
the crumb outranks its own trimmings.

**Bottom status row, drop order:** delta of today's `dropOrder` re-scoped —
1. jobs/watches → 2. keeping → 3. open·want-you → 4. compaction ETA → 5. ctx% → 6. `$`.
The state word is never dropped. Below `hudWrap` (<100) the row may wrap to its own
right-aligned second row exactly as today (`render.go:1571`).

**Phone (<60):** the top bar is not drawn; the deck's top row (`deckTopRow`,
`statusdeck.go:122`) takes its crumb (`project › chat` or the room chip) — the two
geometries converge instead of forking. Deck row 2 and the sheet are unchanged except:
the sheet gains rows for anything demoted (cache, rate, changes, served) — it already
has all of them (`deckItems`, `statusdeck.go:387`).

## Theme, spacing, paint — the existing laws apply unchanged

- **Palette roles only** (`styles.go`): crumb parents `dim`, current segment `ink`,
  task-form left cluster `accent` while the room is open (ONE LIT ELEMENT — the accent
  budget is not grown by this change; when a consent question is up, the question's ask
  hue wins and the bars fall to dim exactly as `paintPart` collapses fades today).
  YOLO `bad`. Money keeps its current paint (ink, `warn` at 4/5 of the conversation
  rail). Telemetry keeps the age-fade ramp (`fadeSeg`, `render.go:2154`).
- **No new hues, no italics, no underline.** Bold only where the design law already
  allows it.
- **Spacing:** separators are ` · ` (`dotted`, `render.go:2853`); minimum 3-column gap
  between left and right clusters (`hudGap`); glyph column 2ch; the top bar's rule is
  the existing `legendLine` hairline. No half-gaps, no new constants without a name.
- **Emptiness law everywhere:** a missing figure is an absent segment, never a
  placeholder. The one historical exception ($0.00 on the live line) is deleted — see
  above; the doc stating the exception updates in the same change.
- **A capability that cannot work is absent:** an unpressable model span registers no
  span; a chat with no host shows no host; a YOLO that is off is nothing.

## The manual law (same change, not a follow-up)

- Update `internal/manual/chat/` pages that describe: the status line and its segments,
  the top of the screen, the task room header, the task strip (remove its page/claims),
  where cost/context/model are shown, what pressing each does, and the keys. Write the
  asker's vocabulary into headings ("where is the cost", "what is the bar at the top",
  "how do I change the model").
- Hunt down and rewrite every page that says the chat has no header, that the strip
  exists, or that `idle` is a word the status row says.
- Gates that will catch omissions: `internal/tui3/manual_test.go`,
  `internal/manual/chat_test.go` probes, `internal/manual/packed_test.go` (rebuild via
  `make build`).
- `internal/session/prompts/system.md` is unaffected (no toolbelt change).

## Acceptance

- `make check` green (vet, tests, build, size ratchet — the strip's removal should pay
  for the top bar's addition; if the binary grows past `SIZE-BUDGET`, say so in the PR).
- `go test ./internal/tui3/` (~150s) green apart from the known-red ledger
  (`.github/known-red.txt`).
- Existing tests that pin the old geometry (`bottomchrome_test.go`, `chrome_test.go`,
  `designlanguage_test.go`, strip tests, room header tests, deck tests) updated to pin
  the new one — moved, not deleted: every law above that a test can state should have
  one.
- `make demo-home` and a bare `bin/aforge` launch verified by eye at 200, 120, 80, and
  50 columns: top bar present/absent per tier, doors light under the pointer, model
  press opens the picker in chat and retargets in a room, YOLO press lands on Safety,
  esc climbs the crumb.
- A change entry in `docs/changes/unreleased/` (`make changelog-new`) whose
  `invalidates:` names what stopped being true: the chat had no header; the strip
  existed; `$0.00` was the stated exception; `idle` was a status word; model/branch/host
  lived at the bottom.

## Out of scope (separate issues, do not start them here)

- The ▁▄█ three-tier model system and a `ctrl+m` tiers picker (a crew redesign).
- The ctrl+tab conversation switcher.
- Rail promote-don't-nest, worst-child glyphs, `#1.2.1` handles beyond what the roster
  already numbers.
- The catch-up seam ("you left here · 14h") and away rows.
- New landing receipts (cache-savings-per-turn lines in the transcript).

santoshkumarradha:
Closes #126.

The chat chrome was two designs wearing one frame: a task got a pinned header and a strip of running work, a conversation at rest got neither, and the model, branch and host were a line at the bottom that had to be found before it could be read. Now there is one top bar for both, and the bottom is two rows plus the composer.

## Where this came from

The implementation is **a glm-5.3 dogfood run's**, salvaged onto `feat/chat-chrome-glm` and finished here. It is a large and genuinely good piece of work — `topbar.go` in particular arrived nearly whole, with the crumb's give-way ladder, the span/paint single-walk bargain, and the doctrine comment at the head of the file all its own. What it did not finish was the last mile: a test package that did not compile, four claims that were documented but not implemented, and the hover half of its own stated law. That is what this branch is.

It is merged up to current `dev` twice — nine commits arrived under the dogfood run (places, the `ctrl+k` switcher, home's enter, spend live at the tree) and nine more while this branch was being finished (the grading loop, the ground law, resource policy, the leash). Conflicts were prose and one test file; each merge commit says how they were resolved.

## What glm built

- **`internal/tui3/topbar.go`** — one row plus a hairline rule above the transcript, chat and task alike. Crumb left, terms right; the crumb's five-rung give-way ladder (`fitTrail`), shared with the run page's own trail; the span-and-paint single walk that keeps a door's columns and its ink from disagreeing about which of them was dropped.
- **`rowdoors.go`** — the meter and the count as doors.
- **`taskchip.go`** — what survives `taskstrip.go`: the live set, the parent seam, the pause.
- The room header, its kin rows and the task strip deleted; the status row folded to what ticks; eighteen manual pages rewritten.

## What I fixed, against the review's numbered findings

**1. `$0.00` was claimed and not implemented.** `CLAUDE.md`, the changelog and `render.go`'s own comment all said zero renders as nothing; `dollars()` still spelled the zero and the row still drew it. The refusal is now the row's (`costSegment`), not the formatter's — a gauge with a tank and a ceiling somebody typed are figures where `$0.00` is the true reading. Four comments that still described the exception now say where it went. The `CLAUDE.md` edit is true and kept.

**2. A stale `roomStop` span put a hitbox over the transcript.** `topBarRows` returned before the span-clearing, so opening a room wide and dragging under sixty columns left a `✕` live over the body — with a three-row hit box, at the phone tier. Fixed twice over: every path out of the bar clears the spans, and `stopMarkAt` refuses on a frame with no bar at all. Test: `TestABarThatIsNotDrawnLeavesNoDoorsBehindIt`.

**3. The test package did not compile.** Fixed: `statusdeck_test` named a field that does not exist, `taskchip_test` was written against six symbols that were never in the salvaged file (rewritten to pin what does exist), four tests still called `roomHeadWord`, and `taskphone_test`'s two phone-strip tests went with the strip. Then ~40 tests that pinned the old geometry were **moved, not deleted** — every one names where its fact went, with the old law kept in its comment. `gofmt` clean.

**4. Doors that did not light.** At review time only the model span brightened. Every top-bar and status-row door does now, at the size of its own span: home, the chat crumb step, the model, YOLO, the back word, the `✕`, the meter, the count. The step is the room header's own — ink inside the accent cluster, accent in a chat — and YOLO takes the band instead, because both lifts would paint an open gate in the hue of something safe. **The chat crumb step and the back word are two hovers because they are two presses**: one leaves the room however deep the trail is, the other climbs exactly one level; at one level down they land in the same place and at two they do not.

**5. The ctx segment.** Now a percent alone, the full `12.4k/128k · 10%` back automatically past `ctxCrowded()` (80% of the compaction threshold — the same one-bit question the paint has always asked), and the sparkline demoted to `/status` and the sheet entirely. `telemetrySheet` is where both spellings meet, so there is no second meter to keep in step.

**6. Segments dropped by switch-omission.** `statusLayout` had a switch with no default and silently discarded six segments, including `segLink`, whose own comment said it was never dropped. `hudLaneOf` routes every segment by name into one of four lanes and **panics on a new one**; `TestEverySegmentHasALane` walks every constant so that can never reach a person. The fade clock is stamped from what the row draws. `segYolo` and `segLink` are routed to the bar, and the reconnect note surfaces there and is never dropped by its ladder.

**7. The deck did not carry the crumb.** `deckTitle` computed `crumbSegments` and returned the last step, with dead locals on the room path. It now walks the same `fitTrail` ladder the bar's crumb does — `project › chat` at rest, the room's chip in a room.

**8. Cleanups.** The orphaned `roomHead`/`roomTrail` prose; **every** `[app.X]` doc link across twenty files that named a deleted symbol; the double-painted rule; the hardcoded `" · "` (and `rowdoors`' magic `at += 3`, now `legendJoin`'s own width, so the separator's spelling and its width come off one constant); the per-frame parent map (one `nodesByKey` for three climbs) and its deref-safe-by-accident walk; `quietParts`, `roomStopSep`, the doubled `dropOrder` doc; **B3** — `clampCells` makes the crumb yield rather than `fit` cutting the `✕` and YOLO off the right end (`TestTheCrumbYieldsBeforeTheSafetyPostureAndTheMark` also proves no two spans ever share a column); **B4** — the count registers no span where this build cannot open a conversation; **B7** — the model rests in its cluster's hue, because ink amid accent was one term unlike its neighbours for no fact about it *and* invisible under a pointer whose lift raises a door to ink.

**9. Manual.** Eighteen pages verified claim by claim against the finished code, and the false ones fixed. `internal/manual/chat_test.go`'s probes pass; the pack is regenerated.

**10. Changelog.** Renumbered to this PR. Four clauses the salvage's entry did not carry (the meter, the deck's crumb, the kin rows, the bar's height yield) and two it carried wrongly (the crumb's two doors do different things; the decaying advertisement names the switcher now).

## Defects I found beyond the punch list

- **The crumb collapsed in a room.** An untitled conversation took the early return that exists for "there is no chat step to draw" and took the *task's* step down with it: the bar in a room said `lab` and nothing about the task the person was standing in. The chat's step is conditional now; the room's is not.
- **The lead glyph column was one cell**, so the bar read `·ui-audit`. It is two.
- **The bar was charged on a frame too short to pay it.** Sixty by twelve under the greeting came to thirteen rows for twelve, and the tear came out of the body. The bar is the first chrome to go, as the spec says: it stands only where the frame keeps a row of the conversation.
- **The run page's word was built at the full width and then clamped**, which cut a trail that has a ladder of its own. It is asked for the columns it will actually get.
- **`taskphone.go`'s `stripPhoneDoor` / `stripPhoneState` / `stripPhonePress`** had no caller left — the strip's phone form. Deleted with it.
- **`/model`'s own description** still told a person to press the model's name in the status line.
- **A whole-life test was deleted as collateral.** `TestATaskRoomWalksItsWholeLifeWithoutAStaleWord` walks one node from queued through the check, a repair round, back to work and into landing, asserting at every step that nothing from the stretch before survives — the one thing none of the per-word tests can check. `roomlife_test.go` was rewritten wholesale for the bar and took it with it. Restored, asking the top bar where it used to ask the room header.

## From the owner's run of the binary

**The room's segment was drawn twice.** Reported off a real session: two stacked `⠑ budget back bed ›` rows under the legend, the first carrying the placeholder and the second bare with the cursor. The cause is not the composer and not the chrome rework — it is older than both. The tray takes the input block's **first** row whenever there is one (a picture attached, a harness picked, or **the thinking dial**, which is what the owner had up), and the two lanes that splice a placeholder onto the box — `redirectLane` and `roomSteerLaneRows` — wrote `out[0]` blind. So the placeholder landed on the tray: the chips were **overwritten and gone**, and the segment was left drawn on two rows.

Fixed with one source of truth — `input.go`'s `draftHeadRow(width)`, which answers which row of the block the draft is on. Both lanes ask it; so do `draftclick.go` and `pastechip.go`, which were each working the same offset out for themselves. Pinned by `TestTheRoomsSegmentIsDrawnOnceWhateverIsOnTheTray`, which covers the empty box, the box with a tray, a typed draft over a tray, and a wrapped draft over a tray (a continuation row must not repeat the chip), and separately asserts the tray survives the placeholder. Verified the test fails on the old `out[0]` and passes on the fix.

**The placeholder's parenthetical is dropped** — `Steer this task… (esc: main)` is now `Steer this task…`. The coordinator left this to judgment; the case for dropping it is that the way out is already on screen, one row above the box, at **every** width a room can be drawn at (`─ room · esc/← back ─`, checked at 200/120/100/80/70/60/50/40/30), and again on the top bar wherever the bar is drawn. Three spellings of one gesture, one of them naming the destination `main` where the other two call it `back`. The placeholder's own job is who is listening, and it now does only that. `TestTheSteerPlaceholderDoesNotRepeatTheWayOut` pins both halves: no `esc` in the box, and the legend still saying it.

**The key advertising was starved, not decayed.** Second report: on a fresh home the legend's right end shows only `/ shows every command`, and `space space home` is never advertised even though `space space` works. I traced it and the gate is fine — `tipsLive` reads correctly on a brand-new home, and the ladder is the one the issue describes (state hint → earned tip → the resting advertising). The starvation is one rung up: the `menu-after-first-turn` notice arms the moment the first answer lands and is retired only when somebody presses `/`. It occupies `slotHint`, which outranks the rest state, so from a person's very first answer until they happen to open the menu the slot says one key and the three doors below it are named nowhere at all. On a fresh home that is every frame there is.

It was also a downgrade on its own terms: the rest state already ends in `/ commands`, so the tip displaced a line naming four doors with a longer sentence about one of them. **Removed**, with the law written where it was: a tip has to teach something the frame is not already saying. The other seven tips still outrank the rest state, which is what the earned machinery is for, and the decay is unchanged — the slot goes quiet only when every tip has retired.

Two tests: `TestTheKeyAdvertisingSurvivesTheFirstTurn` (a fresh home's slot before and after the first answer, and that an armed tip still wins) and `TestTheIdleSlotNamesEveryDoorThatWouldAct` (all four clauses in order once each gesture would act). Verified the first fails with the notice restored: `after the first answer the slot reads "/ shows every command", want "space space home · / commands"`.

One correction to the ask: on a genuinely fresh home the line is `space space home · / commands`, not all three doors. `tab last` has no conversation to go back to and `ctrl+k switch` has nothing to list — each clause is under its own condition, which is this surface's own law that a key which cannot act is not advertised. The three-door line appears as soon as a second conversation exists, and the second test pins that.

Verified in the built binary on a home that has never run aforge, at 120 columns — and the tray row (`⠿ high`, the thinking dial) is up in that capture, which is the same row that triggered the duplicate segment above:
```
──────────────────────────────────────────────────────── space space home · / commands ─

                                                                                 ⠿ high
 ›
                                                                                     1%
```

While in that table I found a second false claim in the manual: it said a room's legend left end is *"empty at rest — the way back is named in the top bar"*. It is not empty — it reads `room · esc/← back`, which is the whole reason the parenthetical is redundant. Corrected.

## From the design-language audit

The current crumb step is **bold in its own hue** — the crumb is read by finding the end of it, and hue alone (ink against dim) is a step a person scanning a peripheral row does not reliably make; weight is spent on where-you-are, which is one of the three things this surface spends it on. The right cluster **rests dim**: it is the quiet "about" voice, and loudness was never how this surface says pressable — the pointer is. The two terms that stay loud stay loud because of what they *mean*: YOLO in `bad`, the redial sentence in the accent (which is the ruling the status row already had for it).

## Judgment calls

- **`dollars()` keeps its zero.** The emptiness law belongs to the row that decides whether to draw a figure, not to the function that formats one — `roomorch`'s fuel gauge, `settingspend`'s ceiling and the composer's cap all read `$0.00` truthfully. Twenty-eight callers, one of which wanted the change.
- **An untitled conversation has a one-step crumb.** `project › project` would be a step that says nothing and a door beside its twin. The consequence is that such a conversation's pointer-exit from a room is the back word alone, which is what the back word is for.
- **The context segment is absent below 1%**, which follows from "a percent alone" plus the meter's own standing law that the percentage is dropped below 1% rather than shown as `0%`. `/status` and the sheet still carry the fraction, inserted at its place in the order when the row dropped it entirely.
- **The spine (a tinted lead column down a room's body) is NOT in this PR.** The coordinator raised it as a judgment call: include it if contained. It is not — a room's body is drawn by the same renderers as the conversation, at the full body width, so a one-cell lead changes the wrap and every x-coordinate the hit-testing, the link spans, copy mode and drag-select resolve against. It needs a `bodyLeft()` the way the rail needed a `bodyWidth()`. **Follow-up issue.**
- **The scroll allocation ceiling rises from 220 to 260, and `PERF.md` moves with it in the same commit** (that file's own rule). The law being pinned is unchanged and still holds: scrolling four thousand lines re-wraps none of them, and zero unchanged entries re-render. What rose is the cost of composing ONE frame, which now carries a whole extra row — a crumb walked from the task tree, two cell clusters with their paint, and the spans that make their words pressable. Measured at 239 with the bar in. I did not cache the bar's row: every fact on it is live, and a cache keyed on all of them is more machinery than nineteen allocations is worth.
- **Nothing ambiently shows a running sub-harness at wide width** any more; the strip's harness chip was the surface that did. The transcript's unreturned tool row and its spinner survive, and `/harness` opens the panel. Flagging rather than fixing: the issue did not mention it, and the honest fix is a decision about the bar's glyph column.

## Where the spec is wrong, and I did not silently follow it

- The issue's interaction table calls the `esc/← back` word "a second, wider target" for the crumb's chat step. **They are different doors**: the back word climbs one crumb level (which the same paragraph says three sentences later — "esc climbs exactly one crumb level"), and the chat step leaves the room whatever the depth. I implemented them as two doors with two hovers and said so on the page.
- The issue says the right cluster's model is a door "at the size of the span" but does not say what it rests in. The design audit's answer (dim, brightening under the pointer) is the one I took; the issue's `paintIdentity` reference points the same way.

## The evidence

`make build` in this worktree, then a private tmux socket, resized live.

**200 columns** — bar present, crumb and terms:
```
· agent-a88ca51964f45f877                                                            deepseek-v4-flash · feat/chat-chrome-finish*
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
...
─────────────────────────────────────────────────────────────────────────────────────────────────── / shows every command ─
                                                                                                                    ⠿ high
 ›
                                                                                                              $0.0019 · 2%
```

**120 columns** — same shape, terms intact:
```
· agent-a88ca51964f45f877                                                   deepseek-v4-flash · feat/chat-chrome-finish*
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                            $0.0019 · 2%
```

**80 columns**:
```
· agent-a88ca51964f45f877           deepseek-v4-flash · feat/chat-chrome-finish*
────────────────────────────────────────────────────────────────────────────────
                                                                    $0.0019 · 2%
```

**50 columns** — no bar; the deck's row 1 carries the crumb:
```
──────────────────────────────────────────────────
                                            ⠿ high
 ›
 agent-a88ca51964f45f877            $0.0019 · 2% ▸
 deepseek-v4-flash
```

**A session that has spent nothing** — no `$` at all, and no `idle`:
```
· agent-a88ca51964f45f877                                                   deepseek-v4-flash · feat/chat-chrome-finish*
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
──────────────────────────────────────────────────────────────────────── space space home · ctrl+k switch · / commands ─
 ›
                                                                                                                      1%
```

**A first launch, on a home that has never run aforge** — the greeting quiets the bar to its crumb, and there is no `$0.00` and no `idle` anywhere:
```
· agent-a88ca51964f45f877
────────────────────────────────────────────────────────────────────────────────
```

**60 × 12 — the frame that used to tear.** The bar is the first chrome to go when the frame cannot pay for it; here it can, and the whole thing comes to eight rows of twelve:
```
· agent-a88ca51964f45f877
────────────────────────────────────────────────────────────
  ┌─┐ ┌─┐ ┌─┐ ┌─┐ ┌─┐ ┌─
  │ │ ├─┘ ├─  │ │ ├─┤ ├─
  └─┘ │   └─┘ │ │ └─┘ │
  ~deepseek/deepseek-v4-flash-latest

  ›
  /task <brief> starts work · / shows commands
```

The room form is pinned by test rather than by hand (`TestARoomPinsAFocusHeader` asserts the crumb, the state word, the rail's clock spelling, the back word, the accent and `headHeight() == bodyTop() == 2` on row 0).

## Checks

- `gofmt -l internal/tui3/` empty; `go vet ./...` clean; `go build ./...` clean.
- `go test ./internal/tui3/` **green, whole package** (423s).
- `go test ./internal/session/ ./internal/manual/ ./internal/packed/` green — all three manual gates and the pack/folder agreement.
- `make build` clean, and the binary is **54.0M against a 59,515,000-byte ratchet**: the strip's removal paid for the bar.
- Binary **56,586,354 bytes** against a `SIZE-BUDGET` of 59,515,000. The strip's removal more than paid for the bar.

https://claude.ai/code/session_0165rrkCBfR6aqZRdCuqULGG


santoshkumarradha:
Waiting on UX check by me wait until my approval

santoshkumarradha:
Closed by the owner 2026-09-04: superseded by the lens stack (#337/#351/#355) and the polish wave (#518); branch archived by tag.

PR #206, opened 2026-09-01
chat: connect OpenRouter through PKCE

AbirAbbas:
## What changed

- Add an S256 PKCE flow for the built-in OpenRouter provider, using a random 127.0.0.1 callback and the documented auth-code exchange.
- Make the missing-provider screen available on fresh, existing, and resumed local interactive chats. Enter opens the browser; pasting a key remains available.
- Preserve an unsent draft when the provider connection interrupts the first model message, and hand the returned key to every open conversation immediately.
- Keep headless, hosted, and custom OpenAI-compatible endpoints on their explicit-key path.
- Update the compiled chat manual and provider settings copy.

## How it was checked

- make build: passed; rebuilt bin/aforge and the packed manual.
- go vet ./...: passed.
- go test ./internal/config/ ./internal/openrouterauth/ -count=1: passed.
- go test ./internal/manual/ and both manual-law package checks: passed.
- Focused browser-flow, TUI, and command seam tests: passed.
- Full internal/tui3 suite: one failure, TestAPathInsideACorrectionIsADoor; reproduced unchanged on clean origin/dev.
- Full cmd/aforge suite: existing tick tests failed for a missing key; both reproduced unchanged on clean origin/dev.
- make size: local x64 binary is over the arm64 budget; clean origin/dev is already 1.10 MB over on the same toolchain. This branch adds about 36 KB and does not change SIZE-BUDGET.

## Checklist

- [x] A change entry is included in docs/changes/unreleased/206-openrouter-pkce.md.
- [x] The manual knows about the changed default and refusal.
- [x] No new line in .github/known-red.txt.
- [x] Only this branch paths were staged.


AbirAbbas:
Live test found and fixed a late-key handoff bug: the profile write succeeded, but Agent.SetAPIKey checked the session request wrapper rather than the provider beneath it, leaving the open session keyless. Commit 8d018f41 unwraps the real provider before applying the key and adds a regression test that verifies the first HTTP request after handoff carries the new Bearer credential. Full internal/session and internal/provider suites pass, along with the focused TUI and command handoff tests.

PR #220, opened 2026-09-01
transfer: the move is told where the eye is, and an idle window lets go at once

### LINKED ISSUE #211: home: the move-it-here story is a dim footer whisper, and a pending transfer renders nothing
## What happens today

On the home page, a conversation held by another window shows `~ another window` on its row. With the cursor on it, the ONLY guidance — "open in another window — enter again to move it here (it moves when that window's reply ends; its tasks resume here)" — is a dim line at the very BOTTOM of the frame, while the eye is on the row and the right-hand detail card. Worse: after pressing enter to claim it, **nothing anywhere changes** while the transfer is pending. No "waiting for that window", no spinner, no elapsed, no arrival confirmation. The surface goes silent exactly when the person most needs it to speak. And the move itself is slow even when the far window appears idle — nothing explains why.

Screenshot evidence (2026-09-01, owner's session): card shows `…4f4e12f/work · open in another window`, `touched 1 file · spent $0.83`, `1.2M tokens · last active 11h` — and the transfer story is two dim lines at the bottom of a 45-row frame.

## Why it's wrong

- The information architecture puts the answer far from the question. The person is reading the card; the card says nothing about how to get the conversation back or what state the claim is in.
- A pending claim that renders nothing is indistinguishable from a broken key. This violates the house's own doctrine (a running thing must be visibly running).
- If the handoff only lands at the far window's reply boundary *even while that window is idle*, that is a latency defect, not a UX one — an idle window should let go near-instantly. Root-cause it: the claim check may ride a slow beat (look at the presence/keeper machinery from #137, e.g. the presence desk's beat cadence).

## What to build

1. **The facts move to the point of attention.** Selected row's right end + the detail card carry the door and the live state; the bottom line becomes at most an echo.
   - At rest: card says `in another window · enter brings it here`.
   - Claim pending: row's `~ another window` becomes a live state (`⠙ coming here…`); card's top states honestly what is happening — `asked that window to let go — it is mid-reply; it moves the moment the reply ends`, or plain `moving…` when idle. Elapsed appears once it exceeds a beat (emptiness law: never `0s`). esc cancels, and the card says so.
   - Arrival: row flips to here; card confirms. Failure/timeout: an honest state in the same place (`that window did not answer — still there`), never a silent revert.
2. **Fix the latency if mechanical** (poll interval / missed wake); if the wait is genuinely "far window mid-reply", the UI says that.
3. Tests pinning each state (rest / pending-busy / pending-idle / arrived / failed), the esc-cancel, and a no-silent-pending law. Manual pages updated in the same change (grep the corpus for the old footer's words).

House laws apply: one accent, dim machine-voice, emptiness law, no machinery vocabulary, keyboard-only reachability.

santoshkumarradha:
Closes #211

rides the UI wave (#167)

## Root cause of the latency — mechanical, confirmed, fixed

`Agent.drainTakeover` — the look for a `takeover.json` another window left — ran **only on the presence desk's `ticker.C`**, and that ticker is `presenceHeartbeat = 5 * time.Second` (`internal/session/taskpresence.go`, `presenceDesk.beat`). So an **idle** far window took up to five seconds to so much as *notice* a request there was nothing at all to wait for. The asking window was already beating at the flock every 200ms; it was watching a door nobody had been told to open.

The probe's guess was right and the diagnosis is narrower than "it only releases at reply boundaries": a window mid-reply correctly waits for the turn to end (that law stands — a reply is never cut), and an idle window *also* waited, for the heartbeat.

The doorstep now has a beat of its own — `takeoverDoorstep = 250ms` — while presence writes and `drainAnswers` keep the five-second heartbeat. The cost is one `os.ReadFile` of a path that is almost always absent, four times a second per live session; that is cheaper per second than the temp-write-and-rename the heartbeat already does.

**Measured on two real `./bin/aforge` processes sharing one isolated `HOME`, in a private tmux socket** (request written into a live session's folder, timed to its deletion by the holder; five samples each, far window idle):

| | pickup |
| --- | --- |
| before | 4.84s · 3.00s · 2.99s · 3.00s · 3.00s |
| after | 0.21s · 0.25s · 0.24s · 0.25s · 0.25s |

End to end through the surface in the same rig, `enter` to the conversation on screen: **3.48s before, 0.24s after**.

## The redesign

The whole account of this door lived on the foot line, and while a claim was actually out it lived nowhere. Three places now, in decreasing width, and never twice:

**the row** — the right margin stops saying `another window` and says `coming here`, and the row takes the page's **one spinner**. That is the one-spinner law's own rule rather than an exception to it: it belongs to the most recently started thing, and nothing on the machine started more recently than the keystroke that asked.

**the card** — under the place line that has just said `open in another window`, in the same band, so it is the same subject with no blank row between:

```
…/ws · main, 1 file dirty · open in another window     at rest
enter brings it here

…/ws · main, 1 file dirty · open in another window     armed
enter again moves it here
its reply finishes first · its tasks come here

…/ws · main, 1 file dirty · open in another window     claimed, far window mid-reply
coming here · 14s
that window is mid-reply — it comes the moment the reply ends
esc stops waiting
```

An idle far window gets no reason line — there is nothing to explain, and it lets go in about a quarter of a second. Past fifteen seconds still quiet, the card says `that window has not answered yet` and invents no cause: a wedged window, a sleeping machine and an older build look identical from here. The clock appears only past two seconds (the emptiness law — never `0s`).

**the foot** — keeps the armed sentence, because that press ends another window and must not be reached by leaning on a key. Otherwise an echo: it speaks only where the card cannot, below `homeCardMin` where there is no second column, or with the cursor walked off the claimed row.

## Two states that had no words because they had no code

- **A request that ages out.** `session.TakeoverStale` is ten minutes and past it the holder deletes the request unread — so the asking window was beating at a lock nothing would ever release, for ever, in silence. It now stops, withdraws the dead request, and the card says `that window did not answer — it still has it` / `enter asks again`. That is what a window killed with the conversation open looks like.
- **A conversation that came free while nobody was on home.** Nothing is opened under somebody who walked away — that rule stands — but the ending was dropped on the floor with it. It is kept now: `it came free — enter opens it`.

`esc` is the one ending that still says nothing, and that is right: the person withdrew the question themselves, so the screen going back is the answer.

## Tests

`internal/tui3/takeover_test.go` pins every state — rest, armed, pending-idle, pending-mid-reply, aged-out, came-free — plus the one-spinner claim, the elapsed clock's emptiness, `esc` cancel, the retry key, the no-duplication rule (foot quiet while the card speaks), and **a no-silent-pending law** that walks every state and fails if the card says nothing. `internal/session/takeover_test.go` picks up the exported `TakeoverStale`.

`go test ./internal/tui3/` green. `go test ./internal/session/` green apart from `TestOnlyADesignsOwnThreadCarriesTheReviseVerb`, which is in `.github/known-red.txt` as a full-suite-load flake and passes alone (verified). `gofmt`/`go vet` clean on every touched file.

## Manual

`internal/manual/chat/home.md` — the transfer section is rewritten and split, with a new `## Why is moving a conversation slow — it says coming here and nothing happens, how do I cancel the move, that window did not answer` carrying the asker's own vocabulary. Five new probes in `internal/manual/chat_test.go`, all reaching `home`.

`internal/manual/chat/running-on-another-machine.md` said `enter` on an `another window` row "will not open a second copy … go to that terminal, or start a new conversation here" — a denial that stopped being true when the move shipped. Removed.


santoshkumarradha:
**The margin now carries the door, on the row the cursor is on.**

```
○ The Other Terminal                        alpha another window · enter brings it here
○ A Third Window                                              alpha another window
```

- **Selected held row grows**, unselected ones keep `another window`. Only the cursor's row — never the pointer's — because the card follows a hovered row *without* moving the cursor, and a promise about `enter` on a row `enter` would not act on is a lie the selection band makes look true.
- **Built from the two spellings that already exist** (`homeHeldShort` + the card's `takeoverDoorWord`), so the row and the card cannot drift.
- **Give-way is dropping the clause, never cutting it.** It is offered *after* the row's ranked facts have dropped, so it spends only cells nothing else wanted — and the test is the conversation's **whole name** still fitting beside it, not the eight-cell floor the drops settle for. This list's job is choosing between conversations; `another window · enter brings i` would have bought half an instruction with the name's last four letters. Below that width the short word comes back whole and the card still says `enter brings it here`.
- **A row with no door is never offered one.** Over `--host` the holder is a window on this laptop and the journal is on the far machine; a conversation with no folder of its own has nowhere to leave a request. Both are `another window`, neither has a way back.
- **Pending / arrived / failed are untouched** — `coming here`, the card's reason, clock and `esc`, `that window did not answer`, `it came free` all as they were.
- Two tiers deliberately keep the short word: a **project card's** conversation list has no cursor on it, and the **phone** tier draws the tail as a second line under the name — a note that grew on the selected row would push that row from one to two lines and shove the column under it down on every `↓`.

**On the footer — kept, and here is the one sentence.** The armed line is the only place on the screen that states the *cost* of a key already half-pressed, and below 160 columns there is no card at all, so it is the only thing that speaks — while the *wait's* echo (`moving it here — esc stops waiting`) was already silent whenever the card is up (`takeoverCarded`), and this change adds no new duplicate at rest, since the resting foot never mentioned the door in the first place.

**Pinned:** `TestTheHeldRowUnderTheCursorNamesTheDoorAndItsSiblingsDoNot` (selected grows, siblings don't, the sentence follows the cursor), `TestTheDoorOnTheMarginGivesWayToTheNameRatherThanCutAWord` (narrow width drops the whole clause, no half-sentence prefix survives, card still carries it — verified failing under the old floor rule), `TestNoDoorOnTheMarginWhereEnterWouldRefuse` (`--host`, and a folderless conversation).

**Manual:** `chat/home.md` gains the row's own words under *Continue a conversation from another terminal*; `chat/running-on-another-machine.md` no longer says the row says only `another window`; two new probes (`the row says another window, how do I get it back`, `how do I bring a conversation back to this window`). Pack rebuilt. Changelog entry updated with the sixth `invalidates` clause.

**Tests:** `go test ./internal/tui3/` green (421s). `internal/manual` green. `internal/session` — only `TestOnlyADesignsOwnThreadCarriesTheReviseVerb`, the documented full-suite flake; green alone. gofmt/vet clean on everything touched.

Issue #233, opened 2026-09-01
task families: uniform part-brief composition — harness composes context, worker authors scope


santoshkumarradha:
## Uniform part-brief composition: harness composes family context, worker authors scope

**Parent:** #228 · **Depends on:** #231 · **Last in sequence** — the adaptability layer that stands on the structural spine.

## Problem

What a part is told varies by road, and the weakest road is the one a cheap worker drives:

- **Sketch parts** (`divideFromSketch`) get the parent's *entire* brief plus sibling boundaries (`sketchBrief`) — the harness composes it.
- **`depends_on` nodes** get prerequisite reports appended (`briefLocked`).
- **Worker-written `divide_work` parts** get *only* the per-part prose the worker wrote. The parent's assembled brief is **not** prepended. The single reviewer pass can patch this up, but it fails open — so a cheap worker that under-restates context ships parts that can't see the family's findings.

This is the same class of bug as the structural ones: context that should be *composed by the harness* is instead *re-authored by the cheapest model on the road*.

## Decision

**Split every part's world into what the harness composes and what the worker authors — uniformly, one composer.**

- **Family context** (harness, identical for all parts): the person's verbatim ask, the parent's assembled brief, the sibling map, the landing shape.
- **Scope** (worker, per part): what it owns, the evidence, the acceptance.

`divide_work`'s `brief` field becomes *scope*; `divideOnce` composes the family context around it — the same composer the sketch road already uses. `sketchBrief` stops being special. The reviewer reviews the composed product, as it already does.

This kills the reliance on a cheap worker restating context N times, by construction.

## Acceptance

- A `divide_work` part's opening message contains the parent's assembled brief and the sibling map, identical in shape to a sketch part's.
- The worker's per-part `brief` text appears as the scope section, not the whole brief.
- No double-printing: the person's ask appears once (it is already in the parent's assembled brief via `composeBrief`).
- `prompts/divide.md` updated: the `brief` field is documented as *scope* — what this part owns — with family context stated as composed by the harness.

## Notes for the implementer

- The composer seam is `divideOnce`'s part-spec construction (task_divide.go:~815, where `spec.brief = part.Brief` today). Wrap `part.Brief` with the family-context block there.
- Factor the family-context block out of `sketchBrief` (task_divide_sketch.go) into one function both roads call — one composer, two callers, no drift.
- The sibling map already exists for sketch parts ("which letters somebody else is holding"); generalize it to "which scopes somebody else owns" using the settled parts list.
[...]

PR #236, opened 2026-09-01
task families: scope-ownership enforcement at division admission

### LINKED ISSUE #231: task families: scope-ownership enforcement at division admission
## Scope-ownership enforcement at division admission

**Parent:** #228 · **Depends on:** #229 · **Blocks:** #TBD5

## Problem

The ledger defines what ships, and `stageTaskWork` stages only the ledger. If two parts of one division write the same path, the parent's ledger contains that path twice — staged once — so **one version silently destroys the other before git ever sees a conflict**. A merge conflict here is a symptom; the disease is overlapping scope.

Today overlap is only *advised* against — `divide.md` says "Two parts that edit the same file are not independent" — and the reviewer *may* catch it. Both are soft. The failure mode is silent data loss in the deliverable.

## Decision

**Enforce path ownership at admission.** When a division is admitted, the parts' declared scopes (the files/paths each names in its brief/acceptance) are checked for overlap. Two parts claiming the same path is a refusal, returned through the existing division-refusal road (`divisionNotAsWritten`-style: ordinary tool result, worker carries on and re-scopes). The reviewer keeps its role of *sharpening* boundaries; this is the hard floor beneath it.

This replaces any notion of merge-conflict resolution between parts: with ownership enforced, same-path writes cannot exist, so there is nothing to resolve.

## Acceptance

- A division whose parts name overlapping paths is refused at admission with a sentence naming the colliding path(s); nothing is admitted, nothing spent.
- Disjoint scopes admit normally (no false positives on parts that merely work in the same directory tree but different files).
- The refusal is journaled with the existing `journalDivision` record.
- The reviewer's brief is updated to state that overlapping scope is refused outright (so it stops spending effort merging overlapping parts).

## Notes for the implementer

- The check belongs in `divideOnce` (task_divide.go:605), after the reviewer's parts settle and before slots are claimed — the same place the fan cap is enforced.
- Scope extraction: parts name paths in `brief`/`acceptance` as prose. Reuse `pathTokens`/`groundHolds` (taskstands.go) to pull candidate paths, normalized against the family tree. Keep it conservative — only *exact same normalized path* claimed by two parts is a collision; directory-level sharing is fine.
- Do **not** try to be cleverer than exact-path overlap. Prefix/subtree claims ("part A owns `src/`, part B owns `src/foo.go`") are a follow-up only if real usage demands it; exact-path covers the measured data-loss case.

<details>
<summary><b>Verification recipe</b></summary>

1. Scripted nest; worker calls `divide_work` with two parts whose briefs both name `report.md`.
2. Assert refusal, the colliding path named in the result, zero nodes admitted.
3. Two parts naming `a.md` and `b.md` under the same directory → admitted.
</details>

santoshkumarradha:
Closes #231.

## What was true

The ledger is the contract of what ships and `stageTaskWork` stages it once. Two parts of one division writing the same path put it in the parent's ledger twice, staged once — **one version silently destroys the other before git ever sees a conflict**. Overlap was only *advised* against, in `prompts/divide.md` ("Two parts that edit the same file are not independent") and in the reviewer's brief. Both are soft.

## What is true now

`divideOnce` refuses a division whose parts claim the same path — after the reviewer's parts settle, before any hand is claimed, on the same line the fan cap stands on.

- New file `internal/session/task_divide_scope.go`: `scopeCollisions(parts, tree)` reads each part's own scope with `pathTokens`, keeps only tokens that `groundHolds` says land under the family tree, normalises with `resolvePath`, and collides on the exact same normalised path claimed by two parts. One pass with a set, not a comparison of every part against every other.
- One call site in `divideOnce`, four lines, above the slot claims.
- The refusal is the ordinary tool-result road (`divisionScopesOverlap`, in `divisionNotAsWritten`'s shape): the colliding path named, nothing admitted, nothing spent, and a worker that can redraw the boundary and ask again.
- Journalled with the existing `journalDivision` record as `refused:scope` — its own word because it is the one refusal here saying the division was *right* and its boundaries were wrong.

**Deliberately dim.** Only the exact same normalised path collides. Parts sharing a directory, or a part owning a folder while another owns a file inside it, are what a good division looks like; a prefix rule over prose would refuse them.

**What a part CLAIMS is its own scope, not the family context around it.** `sketchBrief` composes the *parent's whole brief* above every part and its *siblings' scopes* below it. Reading either as this part's claim makes the harness-drawn road collide with itself on the first path the parent ever mentioned. `partScope` cuts to the harness's own markers (`divisionThisPart` / `divisionOtherParts`); a brief a worker wrote itself carries neither and is its own scope whole. This is the seam #233 formalises.

## Prompts and manual

- `divideReviewBrief`: the overlap bullet now says an overlap left standing is refused outright and nothing is handed out; and the brief-sharpening bullet asks the reviewer to say *whose* something nearby is rather than listing a sibling's files under this part (that instruction was a false-positive generator).
- `divideDescription` and `prompts/divide.md`: a new paragraph, "EVERY PART OWNS ITS OWN FILES, AND THIS ONE IS ENFORCED", with what to do about it and the explicit note that two parts in one directory on different files are fine.
- `internal/manual/chat/tasks.md`, *When a task turns out to be too wide for one worker*: the refusal, the wording the worker is told, why sharing cannot work, and what does **not** count. One new probe in `internal/manual/chat_test.go`.

## Verification recipe

- `TestTwoPartsClaimingOneFileAreRefusedBeforeAnyPartExists` — two parts both naming `report.md`: refused, path named, zero nodes admitted, no lane held, record says `refused:scope` with `admitted=0`.
- `TestPartsThatShareADirectoryButNoFileAreAdmitted` — `reports/a.md` and `reports/b.md`: split into 2 parts.
- `TestOneFileNamedTwoWaysIsStillOneFileAndProseIsNotAClaim` — relative vs absolute spelling is one file; one part naming its own file twice is not a collision; prose is not a claim; a path outside the family tree is not this division's to own; the done-condition claims too.

https://claude.ai/code/session_01UnNCCsBvRnXDoPzFs88zQL


santoshkumarradha:
### Review fix: the refusal no longer claims a cost that was already paid

The defect was real. `reviewDivision` makes and charges the reviewer call before the check ran, so `"nothing is cancelled and nothing is spent."` was false on every scope refusal. Fixed at cause rather than in the wording.

**The check is now asked twice, through one function.**

- **Gate three, free** — `a.scopeRefusal(parsed.Parts, scopeSpentNothing)` sits directly under the free-hands gate, above `node.takeTiebreak()` and above `reviewDivision`. It reads the parts the *worker* wrote, which are the only parts that exist yet. A worker that drew its own boundaries badly — the common case — is refused having spent nothing: no adjudication taken, no model call made. The sentence saying so is now true.
- **After the review** — `a.scopeRefusal(parsed.Parts, scopeSpentTheRead)` stays exactly where it was, above the slot claims. It has to: the reviewer may sharpen a brief onto a file its sibling already owns, and a rule enforced only on the asked-for shape is a rule the settled shape walks around.

Both call sites are one call to `(*Agent).scopeRefusal`. What differs is the ending and nothing else — `scopeSpentNothing` / `scopeSpentTheRead`, spelled once as constants. The post-review ending is `"nothing is cancelled."` and says nothing about spend at all rather than naming a figure: the emptiness law, and the worker cannot act on the number anyway.

`prompts/divide.md` and the manual page now say the same thing — checked twice, the first one free — instead of the old flat "nothing is spent".

**Tests**

- `TestTwoPartsClaimingOneFileAreRefusedBeforeAnyPartExists` now runs against a scripted `divideReviewer` and asserts `reviewer.reads() == 0` plus the free ending — the zero-model-call test asked for.
- `TestAReviewerThatSharpensTwoPartsOntoOneFileIsRefusedToo` (new): worker parts own `north.md` / `south.md`, reviewer rewrites both onto `report.md`. Asserts the reading happened once, the division is refused with `report.md` named, zero nodes admitted, and the answer does **not** contain `"nothing is spent"`.

No other refactor of `divideOnce`, and no prefix/subtree collisions.

Issue #252, opened 2026-09-01
The lens: one transcript grammar for chat and the task room


santoshkumarradha:
## Status (checked 2026-09-12)

**Landed**
- **L1 — feed extraction:** [PR #259](https://github.com/Agent-Field/aforge-v2/pull/259) — `internal/tui3/feed.go`; chat on the feed. Change entry: `docs/changes/unreleased/259-lens-l1-feed.md`.

**Still open**
- **L2** — room adopts feed; delete `room.go` mirrors.
- **L3** — one shaper + elbows.
- **L4** — lens + altitude.

(Lane checkboxes in the body below still list L1 open — treat L1 as done; L2–L4 remain.)

---

The chat and the task room were meant to be one conversation surface pointed at two speakers. The renderer is genuinely shared (`deckRows` has exactly three callers), but everything upstream drifted: `room.go` mirrors eleven event-ingestion functions from `app.go` and is the copy that fell behind — a room never learns a tool finished (`EventToolFinished` unhandled, so no "took 4s"), never sees retries or notices, pairs tool-ends by a different rule than chat, parses raw JSONL itself with hand-copied caps, and renders a steer as an ordinary new question so that on replay a correction is indistinguishable from the brief.

The full design — philosophy, decisions, rejected alternatives, and the owner's four rulings (2026-09-01) — is `docs/design/lens/DESIGN.md` (lands with lane L1).

**The one-sentence design:** one reducer and one renderer produce every transcript; a page differs from another only by a declared lens — the participant's (chat: dialogue at full fidelity, work folded) or the overseer's (task: trajectory at a glance, machinery one gesture down) — and a correction typed into running work is an elbow on both.

**Rulings, settled:**
1. The task page folds settled work into phase chips by default; the machinery stays one keypress away (reverses the written no-fold law in `workfold.go:26-52` and its three pinning tests).
2. A room steer is an elbow (`└ `, drawn where said, opens no turn); the node journal gains a steer mark so replay stops lying.
3. Args-match tool-end pairing on both surfaces (chat drops oldest-of-name).
4. Every task steer confirms delivery: a short fading `· delivered` clause (the woke-a-parked-node sentence when that is the truer fact).

**Lanes:**
- [ ] **L1 — feed extraction.** A `feed` type owns event → `[]entry` for chat; no behavior change; tests stay green. Ships with the design doc.
[...]

PR #259, opened 2026-09-01
lens L1: one feed reduces events to entries (#252)

santoshkumarradha:
Lane L1 of the lens design (docs/design/lens/DESIGN.md, Decision 1). A pure
refactor: no behavior changes, no manual pages touched, because nothing a
person can see moved.

## What moved

Live event ingestion comes off `*app` and onto a new `feed` type in
`internal/tui3/feed.go` that owns `event → []entry` and knows nothing about
how any of it is drawn:

- `formTool`, `announceTool`, `beginTool`, `claimAnnounced`, `closeTool`,
  `finishTool`, `settleCompaction`, `closeLive`, `resolveUnfinished`
- `claimForming`, `claimFormed`, `formingPreview`, `formingPreviewField`
- the thought settling from `thinking.go` — `settleThought`, `collapseThought`
- `openCompaction`, which is the `EventCompacting` arm the switch used to
  spell out inline
- `ingest`, the one switch a new event kind now gets wired into

The law is written at the top of the file:

> **ONE REDUCER PRODUCES EVERY TRANSCRIPT'S ENTRIES; A VIEW MAY INSTALL
> HOOKS, NEVER A COPY.**

`room.go`'s eleven mirrors are the standing proof of what the other way costs
— being the copy is why a room still does not learn a tool *finished* — and
they are deleted in L2, not here.

## The seam

`feed` is **embedded** in `app` rather than held beside it, so every existing
reader of `a.entries`, `a.live`, `a.think` and `a.turn` is spelled exactly as
it always was: the fields moved house, not name. That is what keeps a
mechanical extraction from becoming a seven-hundred-site rename.

The three acts that belong to the chat alone become `feedHooks` the chat
installs in `newApp` (`app.feedHooks`), so the reducer stays view-agnostic:

| hook | what the chat does with it |
| --- | --- |
| `forming` | draws the `propose_task` spawn card from the same fragments |
| `closing` | reads a result-with-a-forming-card as a refusal |
| `closed` | learns a backgrounded `bash`, drops the ambient-count cache |

Plus `now` / `follow` / `touch`, which are what any surface with a screen owes
a reducer. Every hook is optional — a `feed` with none installed still reduces,
which is the room's configuration in L2.

## What explicitly did NOT change

- **Tool-end pairing.** Chat keeps oldest-of-name (`closeTool`'s own walk).
  Ruling 3 — args-match-then-oldest on both surfaces — lands in L2.
- **`room.go`'s mirrors.** Untouched except for doc-comment link targets that
  now point at `[feed.X]` instead of `[app.X]`. No room code path moved.
- **Rendering, folding, steering.** Nothing in `render.go`, `workfold.go`,
  `steerelbow.go` or `deckRows` was touched.
- **The event pump.** `apply` keeps everything that is not entry work:
  `EventToolEnd` still speculatively prefetches the written file,
  `EventCompacted` still rebases the scrollback and re-reads the meter, the
  wait-clock anchors and the winding-down gate are where they were.
- **`pendingReplyTags`.** Left on `app` in this lane — `appendText` consumes
  it, and splitting its ownership across two types for no gain is how a seam
  goes wrong. It joins the feed in L2 with the rest of Decision 4's events.

## How no-behavior-change was verified

- The moved functions were moved **verbatim** — receiver renamed, `a.` →
  `f.`, and three call sites turned into hooks. `git diff` on `app.go` is
  506 deletions against 72 insertions, and the insertions are the hook
  installation and the rewired switch arms.
- The one ordering question in the move: `hudStale = true` used to be set a
  few statements after `learnBackground` inside `closeTool`'s matched branch,
  and now both run in the `closed` hook at `learnBackground`'s old position.
  Nothing between the two reads `hudStale`, so the two orders are
  indistinguishable.
- `go build ./...` and `go vet ./internal/tui3/` clean; `gofmt` clean on every
  file touched.
- `go test ./internal/tui3/` green (the whole ~470s suite).
- `make build` succeeds, packed manual and all.
- `go test ./internal/session/` — one failure, `TestTheFixedPrefixStaysUnderItsBudget`,
  which **fails identically at `origin/dev`** in a clean detached worktree (the
  fixed prompt prefix is 758 bytes over its 48000 budget). Not this branch's,
  and not in `.github/known-red.txt` yet — worth its own issue.

## New tests

`feed_test.go` pins the law from both sides: one call becomes one row from the
first fragment to the result with **no view behind the reducer at all**, and
the chat's own three acts arrive through hooks rather than through a copy.

Refs #252.


santoshkumarradha:
## Review — high effort, all findings verifier-confirmed

**Headline: the zero-behavior-change claim holds.** Three independent byte-level audits found no semantic drift between the old `app.go`/`thinking.go` ingestion and the new `feed` — twelve of fourteen moved bodies are byte-identical after receiver normalization; the two that differ do so only at the hook seams, whose contents and call positions match the old inline code. Build/vet/targeted tests pass at head. No manual page is owed (nothing person-facing moved) and the change entry is correct.

Ten confirmed findings, ranked. 1–2 undercut the PR's own one-reducer law; 3–4 are the only correctness-grade items; the rest are seams L2 will trip over.

1. **`app.go:4194` — apply's grouped case re-spells `feed.ingest`'s event-kind list.** A future event wired in `ingest` works in tests and the room but never fires in chat. `ingest` no-ops on unknown kinds, so apply can dispatch unconditionally and leave one authoritative list.
2. **`app.go:4251` — `EventRetrying`'s reduction stays hand-rolled in the pump** (`dropLive`, `dropRetryingFormingTools`, `resolveUnfinished`, `dropRetryingFormingCard` all mutate feed-owned state from outside), though DESIGN.md Decision 4 lists it as feed-handled. As written, L2's room still won't see retries — the defect the design names.
3. **`feed.go:371,410` — `announceTool`/`beginTool` adopt a forming row without copying `ev.CallID`**, discarding the id at the one moment session offers it. Two concurrent same-name calls can get each other's "took Ns". Pre-existing, moved verbatim — but not covered by ruling 3's L2 deferral (that covers tool-END pairing); copying the id here is free.
4. **`feed.go:487` — latent pointer invalidation**: `closeTool` keeps writing through `e := &f.entries[i]` after `hooks.closed(e, ev)`, with no stated no-append invariant — and the forming hook already appends today (`formTask`). A future appending closed-hook silently defeats FAILURE OPENS ITSELF. Re-resolve the index after the hook or state the invariant in `feedHooks`' doc.
5. **`feed.go:37` — the zero value is wrong** (`live`/`think` = 0 point at entry 0), the `-1` sentinels are hand-spelled at four sites, and hooks install by post-construction mutation with every hook nil-guarded — a mis-built feed degrades silently. A `newFeed(hooks)` constructor closes all of it before L2 builds a second feed.
6. **`feed.go:583` — `settleCompaction` doesn't `touch()`/`follow()`**; chat repaints only because apply compensates (`app.go:4228-4229`), contradicting `feed.go:123`'s "ingest is the whole of what an event does". A bare-feed consumer gets a divider that doesn't paint.
7. **`feed.go:623` — `resolveUnfinished`'s forming-row arm is dead** at every current call site because `app.dropForming` runs first — two spellings of one law whose outcome depends on caller order.
8. **`thinking.go:70` — `appendThought` stayed on app** while its lifecycle partners moved to feed, so the reducer can't cover `roomThink` when L2 deletes the mirrors — that hoist should happen before, not during, L2.
9. **`feed.go:469` — full-slice scans per tool event** (`closeTool`, `finishTool`, `claimAnnounced`, `claimFormed`, `resolveUnfinished`), O(n²) over a session, now canonicalized in the reducer every surface will run. A settled-floor index makes each claim O(live rows); the extraction is the natural moment.
10. **`room.go:1125` — new comment points at the wrong file**: "app.go's [feed.formTool]" — it lives in feed.go now.

Recommended before merge: 1, 2, 3, 6 (cheap, and 1/2/6 are the one-reducer law itself). 4, 5, 7, 8 are fine as immediate follow-ups but ideally land before L2 starts. 9 optional, 10 one word.

https://claude.ai/code/session_0165rrkCBfR6aqZRdCuqULGG

Issue #255, opened 2026-09-01
task families: a landing that could not save the work merges anyway, and then deletes it


santoshkumarradha:
## A landing that could not save the work merges anyway, and then deletes it

**Parent:** #228 follow-up

## Problem

`comeHome` (`internal/session/task_run.go`, near 6017) is the one road every merging landing takes. It commits the node's ledger and throws the answer away:

```go
_ = commitTaskWork(t.dir, title, wrote)
```

`commitTaskWork` (near 6294) can answer `nil` three ways and none of them reaches a caller: `stageTaskWork` said the directory is not a worktree, nothing ended up staged, or `git commit` itself failed — the commit's own error is dropped on the floor (`_, _ = git(dir, …, "commit", …)`). `stageTaskWork` (near 6374) has the same shape one level down: it returns `false` only for a non-worktree, and ignores the error from `git add --all --` and from every per-path retry under it.

`comeHome` then carries on regardless. It merges a branch that holds nothing — which succeeds, because there is nothing there to conflict with — and `releaseLanded` (`internal/session/groundladder.go`, near 765) removes the worktree, with `--force` when the plain removal refuses. The node settles `TaskDone`.

**Evidence** at `dev@c1776b76`. One repository ground, one file the worker wrote, and the worktree's git directory made unwritable — a full disk, a read-only mount, a permission somebody changed:

```
merge="merged" detail="it left files it did not write, and they went with its working copy rather than onto your branch: parser.py"
parser.py on the person's branch: false · still in the worktree: false
```

Two things are in that line. The work is gone from both places. And `leftBehind` read the node's own unstaged deliverable as somebody else's droppings, so the landing's own sentence names the file it has just destroyed — as junk.

Since #237 the ledger composes, so at a family's landing this is not one node's slice: the parent's list carries every landed part's paths, and the one discarded commit is every generation's work.

## Decision

**A landing that could not save the work does not merge, does not release anything, and does not say done.**

- `stageTaskWork` and `commitTaskWork` answer WHAT WENT WRONG instead of a bare `nil`/`false`: git's own first line for the add, the commit, or the reset that failed.
- `comeHome` stops at the first failure. It does not merge, it does not `releaseLanded`, and the branch and the worktree both stay exactly where they are — the work is on disk and that is the only copy of it.
- The outcome is the one that already exists for a branch that never came home, `mergeAborted`, carrying a sentence in the shape `unreachedSentence` already uses:

[...]

PR #237, opened 2026-09-01
task families: ledger absorption — part writes compose into the parent

santoshkumarradha:
Closes #229.

## What was true

A landing carries the **ledger** — the paths the node wrote — and nothing else: `stageTaskWork` stages that list on a repository ground and `landMirror` lays it back over a folder one. A node that handed work out had a hole in it. Its parts wrote into ITS tree while their paths sat on THEIR ledgers. On a repository ground git closed the hole underneath, because a part's work arrives as commits on the very tree the parent merges. On a folder ground nothing closed it: the mirror held every file the family made, the parent's ledger named its own, and the person's folder got that one. The task said done and the check passed against the mirror.

## What is true now

The ledger composes. `absorbedLedger` (new `internal/session/task_ledger.go`) folds every landed part's paths into the node's own **at the landing** — one line on each of the three roads that reach `comeHome`: the audit-off road, the checked done road, and `landStopped`'s verified road. `comeHome` keeps reading exactly one list; what changed is that the list is complete.

Because a node settles holding what it absorbed, the fold is recursive by construction: what a part settles with is what its parent ships.

It is folded **after** the check and never before, because the checker's packet has to keep the two halves apart — `Files it wrote:` must stay a true claim about this node, and `And the parts it handed out wrote, into the same tree:` is the parts' own sentence (`task_audit.go`'s `auditQuestion`).

`absorbedLedger` is `landingFilesFor(...).all()`, so the emptiness law and the landed-only law come from the one existing source of truth: a part that wrote nothing absorbs nothing, and a part that never landed is not on the ledger at all. Ordered dedup of an identical path is `mergePaths`/`alsoChanged`'s, unchanged — scope-overlap refusal at admission is #231's and is untouched here.

## Tests

`internal/session/task_nest_test.go`:

- `TestAFolderFamilyLandsWhatEveryPartWrote` — the issue's recipe end to end: a mirror-mode parent on a `t.TempDir()` ground writes one file, `divide_work`s into two parts, each part writes a distinct file, the family lands, and the person's folder holds all three with no `files:` re-declaration. It also asserts the ledger the family settled with names all three, which is the half the level above reads. **Verified failing without the fix** (`first.md: no such file or directory`).
- `TestAFamilyLedgerFoldsAPartsWholeSubtree` — the fold composes downward: a part whose own ledger already carries a nested path is folded whole and lands whole through the real mirror.
- `TestAPartThatWroteNothingAddsNothingToTheLedger` — the emptiness law at the seam.

[...]

Issue #230, opened 2026-09-01
task families: folder family tree — parts cut worktrees off the mirror


santoshkumarradha:
## Folder family tree: parts of folder/mirror parents cut worktrees off the mirror

**Parent:** #228 · **Depends on:** #229 · **Blocks:** #TBD4

## Problem

`divide.md` and `prompts/task.md` both promise a dividing worker that each part runs "in a copy of the repository taken from yours." For a parent whose ground is a git repository this is true — parts cut worktrees off the parent's worktree HEAD and merge back. For a parent whose ground is a plain folder (`folder`/`in place`/`mirror`), `prepareTaskTreeAt` falls through to `dir: workspace` — **the part works in the parent's own directory**, concurrently with its siblings, with no isolation and no merge semantics.

This is the largest class of general agentic work: research folders, document sweeps, data directories, multi-section reports. Today those families run their parts on top of each other.

## Decision

**The mirror is the family tree; parts cut real worktrees off it.** When a folder/mirror parent divides, each part gets `git worktree add` off the mirror's HEAD, in its own directory under session `trees/`, merging back into the mirror on landing. The mirror is harness-owned space, so this is a real git worktree with real isolation — not a copy, not a shared directory.

- The mirror already holds the ground's content (it is the byte-for-byte copy made at family start).
- Parts merging into the mirror use the *same* `comeHome` machinery as repo parts — one road.
- The family's final landing into the person's folder is unchanged: `landMirror` lays the (now complete, post-#229) ledger by name.

This deletes the "parts share the parent's directory" hazard and makes the prompt's promise true for folders.

## Acceptance

- A folder/mirror parent that divides gives each part its own directory; two parts writing different files never touch each other's tree.
- A part's work merges into the mirror and is present there before the family lands.
- The audit (`restoreFromFolder`) judges the integrated mirror.
- `in place` families are explicitly out of scope for worktree isolation (the person said "here") — parts of an in-place parent follow the add-only rule (#TBD4's landing shape) and this issue documents that boundary.
- Prompt lies removed: `prompts/task.md` and `divide.md` no longer claim "copy of the repository" for folder families; they describe the family tree.

## Notes for the implementer

- The seam is `prepareTaskTreeOn` (task_run.go:5260): the `TaskModeMirror`/`TaskModeFolder` arms currently return a shared `dir`. For a node with `spec.parent != 0` whose parent's tree is a mirror, cut a worktree off the mirror instead.
[...]

PR #270, opened 2026-09-01
task families: a folder family writes over the person's own edits without a word

### LINKED ISSUE #258: task families: a folder family writes over the person's own edits without a word
## A folder family writes over the person's own edits without a word

**Parent:** #228 follow-up

## Problem

A mirror family copies the person's folder at the start (`mirrorGround`, `internal/session/task_run.go` near 5855), works in that copy for as long as the work takes, and lays its ledger back over the folder at the end (`landMirror` near 6089 → `layWork`, `internal/session/task_audit.go` near 1581). `layWork` is a blind overwrite:

```go
_ = os.RemoveAll(target)
if err := copyPath(source, target); err != nil { … }
```

Nothing anywhere records what those files looked like when the copy was made, so nothing can notice that the person changed one while the work ran. `mirrorGround` copies bytes and keeps nothing.

**Evidence** at `dev@c1776b76` — a mirror family over a folder holding `notes.md`; the family rewrites it in the mirror, and the person edits their own copy while it runs:

```
merge="inplace" detail=""
the person's folder now holds: "the line the task wrote\n"
the node settled as done with report "tidied the notes"
```

The person's line is gone. The detail is empty, the state is `done`, and there is nothing in the report, the card or the job log that mentions it happened.

**And the window is now much wider than one run.** #237 made the late roads real: a family that settled needing a look is laid over the folder when somebody types `accept` (`acceptTask`, `task_audit.go` near 2011) or when a late verdict arrives (near 2180) — hours later, over a folder somebody has been working in all morning, on a ledger recorded before lunch.

**This is not what `groundShift` catches.** `taskground.go` asks a different question — did another *aforge task or window* land in these paths — off the project index and the live claims. A person editing their own file in their own editor appears in neither. It is also asked on only three roads (`task_run.go` near 3383, 3457, 3535) and on neither of the late ones, so an accept has no manners check of any kind.

Related, and the same missing baseline: `restoreFromFolder` (`task_audit.go` near 1479) re-runs `mirrorGround` **at check time**, so a family whose ground has moved is already being checked against material that no longer matches the copy it worked in.

## Decision

**A landing does not write over a file that changed under it. It says which ones, and lays nothing.**

- **A baseline is recorded when the mirror is made.** `mirrorGround` is already reading every byte it copies; it records a digest per path as it goes.
- **The baseline lives with the tree, in its own private corner** — `<tree.dir>/.aforge-v3/ground-baseline.json`, written and read by the pair `rememberLeftBehind`/`rememberedLeftBehind` (`task_run.go` near 6207) already models. It survives a dead process, an accept the next morning and a re-audit, without a byte reaching the checkpoint. A folder with no record is a folder from an older build and lands exactly as it lands today — additive, absence is ordinary.
- **`landMirror` compares before it lays.** For every path in the ledger: what the ground holds now against what the baseline says it held. Equal — including *absent then, absent now* — and the file is laid. Different, in either direction, and it is named.
- **The refusal is `landConflicted`'s, in the folder's words.** Nothing is laid, the mirror is kept whole, the outcome is `mergeAborted`, and the node settles **needing your look**:

  > finished, but needs your look — its work is in `<dir>` and was not laid over `<ground>`: `notes.md` changed there while this ran

  The names come through `namedFew(…, conflictNamesShown)`, the way a conflict's do.
- **All five roads, one check.** The two in `workTaskNode`, `landStopped`'s, and both late roads in `task_audit.go` reach `landMirror`, so they all get it by construction. This is the folder ground's half of what git does for a repository ground — where a person's own edit to a file the branch touches is exactly what makes a merge refuse.

<details>
<summary><b>Rejected alternatives</b></summary>

- **The baseline on the checkpoint** (`taskRecord`, beside `Rung`/`Seal`/`Base`/`Universe`). A mirror is bounded at `auditRestoreEntries` = 20000 paths and the record is rewritten as the graph moves; a landing manner should not cost the checkpoint a megabyte.
- **The family tree's baseline commit** (`openFamilyRepository`, `task_tree_mirror.go` near 109). It is a byte-exact baseline and it is free — but only where the mirror could be opened as a repository. The family whose `git init` failed is the degraded one whose parts already share a directory, and it is the last one to leave without a manners check. One road beats a cheaper road that skips the hard case.
- **Merging the two versions.** There is nothing to merge with: the ground is a plain folder and neither side is a commit. Naming the file and standing back is the honest answer, and it is what a person can act on.
</details>

## Acceptance

- **e2e:** a mirror family lands `notes.md`; the ground's `notes.md` is edited after the family started; the landing settles `TaskUnverified`, the report leads with `finished, but needs your look — ` and names `notes.md`, **the person's edit is still in their folder**, and the family's own version is still in the mirror directory the sentence names.
- **e2e:** the control — the same family over an untouched folder lands every file, settles `TaskDone`, `Merge` is `inplace`, and the report gains no sentence (the emptiness law).
- **e2e:** the late road — a family that settled needing a look, whose ground is then edited, and which is then accepted by hand, refuses the same way rather than laying the stale ledger.
- **Unit:** a path that was absent at mirror time and exists in the ground now is refused (a person created the file the task was going to create); a path the task wrote and then deleted, untouched in the ground, is still removed; a folder with no baseline record lays exactly as it does today.
- The manual page on folder landings (`internal/manual/chat/how-tasks-run.md`) states the refusal and quotes it, and the change entry's `invalidates` says the landing used to write over a person's own edits in silence.

<details>
<summary><b>Verification recipe</b></summary>

Deterministic, no model — `internal/session`, the shape `task_nest_test.go` already uses for folder families. This is the run that produced the evidence above:

```go
ground := t.TempDir()
writeFile(t, filepath.Join(ground, "notes.md"), "the original line\n")

tree, _ := prepareTaskTreeOn(ctx, Place{}, workspace, "dddd4444dddd4444", 11,
    "tidy the notes", taskStand{dir: ground, mode: TaskModeMirror})
writeFile(t, filepath.Join(tree.dir, "notes.md"), "the line the task wrote\n")

// The person edits their own copy while the work runs.
writeFile(t, filepath.Join(ground, "notes.md"), "the line the person typed\n")

ledger, merge, detail := landHome(node, tree, []string{"notes.md"})
node.finish("tidied the notes", ledger, "", merge)
```

Today: `notes.md` in the folder reads `"the line the task wrote\n"`, `merge` is `inplace`, `detail` is empty and the node is `done`. Wanted: the person's line untouched, `mergeAborted`, and the file named.

**Field (real models).** Open `bin/aforge` in an empty directory and ask for work whose `where` is a plain folder somewhere else that it will write into — that is the ladder's mirror mode (`groundMode`, `taskstands.go`: a non-repository ground that is not the workspace, with writes named under it). While the task runs, open one of the files it is going to write and change a line. Needs `OPENROUTER_API_KEY`, a few cents, a few minutes. Today the change disappears at landing with no sentence anywhere.
</details>

santoshkumarradha:
Closes #258

## What was true

`layWork` is `RemoveAll` + copy and `mirrorGround` recorded nothing, so a mirror family's
landing wrote straight over a file the person had edited in their own folder while the work
ran — no sentence in the report, the card or the job log, state `done`, `merge` `inplace`.
Since #237 the same happens on the two late roads, hours later, from a ledger recorded
before lunch.

## What is true now

- **A baseline is recorded when the mirror is made.** `rememberGroundBaseline` walks the
  ground once at `prepareTaskTreeOn`'s mirror case and writes a digest per path to
  `<tree.dir>/.aforge-v3/ground-baseline.json` — the private corner and the shape
  `rememberLeftBehind`/`rememberedLeftBehind` already model. It survives a dead process, an
  accept the next morning and a re-audit, and not a byte of it reaches the checkpoint. It is
  written where the copy is made and nowhere else, so a **resumed** node keeps the baseline
  its first run recorded.
- **`landMirror` compares before it lays.** For every path in the ledger: what the folder
  holds now, what the baseline says it held, and what the family would put there. Equal to
  the baseline (absent-then/absent-now included) → laid. Equal instead to what would be laid
  → laid, because the lay changes nothing — this is what keeps the accept road idempotent
  after a landing that already happened. Otherwise → named, and **nothing is laid**.
- **The person reads a `landConflicted` sentence in the folder's words**, settling
  `TaskUnverified`:
  `finished, but needs your look — its work is in <dir> and was not laid over <ground>: notes.md changed there while this ran`
- **All five roads, one check**, by construction.

## The two decisions the issue asked to be justified

**The family tree's baseline commit is not used as the baseline** (#230's
`openFamilyRepository`), and the file comment says why: that commit exists only where
`git init` succeeded, and the family whose init failed is the degraded one — whose parts
already share a directory — and the last one that should also leave without a manners
check. One road that covers every family beats a cheaper one that skips the hard case.

**The mark is `mergeConflicted`, not `mergeAborted`.** The issue proposed `aborted`; every
one of the five roads already tests for `conflicted` and routes it to a needs-your-look
settlement with the names in front of the person, so `conflicted` delivers the issue's whole
decision — including "all five roads, one check" — with no change to any road, and it is
also what this *is*: the same file changed on both sides. It leaves `aborted` to mean what
#255/#256 are giving it (a landing that could not save the work), and it stays correct
whichever way that lane rewrites the road tests. Two person-facing lines assumed a
conflicted node has a branch name; both now leave the branch clause out when there is none
(the emptiness law) — `taskSettleNote` and the settled card's tail.

## A file the person deleted is refused

Deleting is an edit somebody made on purpose; laying the family's version over it would put
back the one file they had just thrown away. They get the name and the directory the other
version is in. A file the *family* wrote and then deleted, untouched in the ground, is still
removed — that is the unchanged case.

## Tests

`internal/session/task_mirror_manners_test.go`, eleven of them: the issue's own recipe, the
control, the run road end to end (a scripted family whose person types into the folder
mid-run), the late accept road, a file created by the person, a file deleted by the person,
a file the family removed, the already-laid folder, a folder with no record, and the entry
cap.

https://claude.ai/code/session_01UnNCCsBvRnXDoPzFs88zQL


santoshkumarradha:
**Independent review round** (codex, high effort, against `origin/dev`) — seven findings, four fixed in `99543448`:

- **Fixed (high):** the baseline walked the person's folder *after* the copy was taken, so a save made during the copy itself would have been recorded as the original and then written over — the defect, in the seconds nobody watches. It walks the copy now, which holds the bytes the family was actually given.
- **Fixed (high):** a directory the walk could not open folded in as "empty", which would have let a landing `RemoveAll` files it had never seen. The trouble is carried out of the walk and answered as `digestUnknown`, which is never equal to anything. Pinned by a test that chmods a directory to `0o000`.
- **Fixed (low):** a ledger path naming a whole directory scanned every recorded path. The record's keys are sorted once, on the first such path, and binary-searched.
- **Fixed (low):** added the run-road **control** e2e — the same scripted task over an untouched folder settles `TaskDone`, `Merge` `inplace`, lays its file and gains no sentence.

Not taken, with reasons now in the file's own comments:

- **Fail closed when the record is missing or unwritable.** The issue rules the other way, and it is the right ruling: absence is ordinary (an older build, a folder that could not be read in full), and failing closed would make every family carried across an upgrade need somebody's look. The record lives in the harness's own corner beside `left-behind.json`, not in the person's folder.
- **Put the permission bits in the digest.** They do not survive the copy: `copyPath` creates the mirror's file with the source mode *through the umask*, so an ordinary `0664` file is copied as `0644` on most machines and every such file would be refused as "changed" with nobody near it. Contents are what a person edits.
- **The compare-then-lay window.** Real and inherent: a plain folder has nothing to lock, and git reads its index then writes the working tree the same way. What this changes is the size of the window — it was the whole of a run, and a whole morning on the late roads. Stated as a limit in the file header rather than papered over.

Issue #232, opened 2026-09-01
task families: WIP commit at divide time so parts see the parent's on-disk work


santoshkumarradha:
## WIP commit at divide time so parts see the parent's on-disk work

**Parent:** #228 · **Depends on:** #230, #231

## Problem

Workers are told never to commit ("you never run `git add`"); the harness commits only at landing (`commitTaskWork`). So when a parent divides **mid-run**, each part's worktree is cut from the *base* commit — the HEAD the family started from. Anything the parent has written to disk by then (a repro, a failing test, scraped material, draft sections, a half-built fixture) is **not on disk for any part**.

The parent-stays law means the parent is the coordinator holding the findings — but the parts can't see the findings' *artifacts*, only whatever prose the parent restates in their briefs. For code work this is the difference between a part that can run the failing test and one that has to be told about it.

The sketch road (`divideFromSketch`) doesn't suffer — division happens before the first request, nothing is written yet. Mid-run `divide_work` after the parent wrote anything does.

## Decision

**Commit the parent's work-in-progress to the family branch at divide time.** Before cutting part worktrees, the harness stages the parent's current ledger and commits it to the family branch — the harness committing, not the worker, so "you never run `git add`" stays true. Parts then branch from a HEAD that contains the parent's real files.

- No-op for the sketch road (nothing written yet — the commit is empty and skipped).
- Applies uniformly to repo families (worktree off parent worktree) and folder families (worktree off the mirror, post-#230).
- The WIP commit lands on the *family* branch, which the person never sees; `comeHome` merges the family branch home as one operation. No squash needed.

## Acceptance

- A parent that wrote `repro.txt` then divides: each part's worktree contains `repro.txt` on disk at start.
- Sketch-road division (nothing written) produces no empty commit.
- The family's final landing is unchanged — one merge home, WIP commits internal to the family branch.
- The parent's own subsequent writes after the WIP commit are still staged/committed at landing as today (the WIP commit is additive, not a replacement for `commitTaskWork`).

## Notes for the implementer

- The commit point is in `divideOnce`'s admission path (post-#231 scope check, before part worktrees are cut) — but parts' trees are prepared lazily at node start (`workTaskNode` → `prepareTaskTreeOn`), so the WIP commit must happen before the *first* part's tree is prepared, not at `divideOnce` return. The clean seam is where the parent's tree is known and the parts are about to start.
[...]

Issue #229, opened 2026-09-01
task families: ledger absorption — part writes compose into the parent


santoshkumarradha:
## Ledger absorption: part writes compose into the parent

**Parent:** #228 · **Blocks:** #TBD3, #TBD4 · **First in sequence** — kills the data-loss bug and proves the invariant the rest stands on.

## Problem

A task family's product is defined by its **ledger** — the list of paths the family wrote, which `stageTaskWork` stages and `landMirror` lays back into the ground. Today the ledger is per-node and does not compose.

Concretely: a `mirror`-mode parent (plain-folder ground) that divides work gets parts that write into the parent's mirror directory. At landing, `landMirror` copies home **only the parent's own `wrote` list** (`task_run.go:5530`). A part's files never reach the person's folder unless the parent happened to re-declare them on its `files:` line. That is silent work loss — the task reports done, the check passes against the mirror, and the person's folder is missing the parts' output.

The same gap exists one level down: at depth 2, a grandchild's writes must reach the root ledger or they strand in an intermediate tree.

## Decision

**Absorb part ledgers into the parent on landing, recursively.** When a part lands, its shipped paths append to the parent's ledger. The rule:

> The ledger decides what ships; the repository decides how it merges.

Repo grounds already work this way (`stageTaskWork` stages the ledger, not the directory). This change gives folder families the same law, so `landMirror` ships the *family's* product, not the parent's slice.

## Acceptance

- A mirror-mode parent that divides work lands its parts' files into the person's folder without any `files:` re-declaration.
- Depth-2: a grandchild's writes reach the root ledger and ship.
- A part file the parent *also* wrote (same path) is refused at admission — see #TBD3 (this issue must not mask overlap by absorbing it).
- `restoreFromFolder` (audit) sees the family's full product in the mirror.
- No change to repo-worktree behavior (ledger already composes there via merge).

## Notes for the implementer

- The absorption point is the part-landing path in `workTaskNode` / the completion fold, not `landMirror` itself — `landMirror` should keep reading one ledger; the fix is that the ledger is now complete.
- `changed` in `runTaskChild` is the per-node list; the parent's effective ledger is `changed ∪ (landed parts' changed)`. Find where the parent's `changed` is finalized before `comeHome`/`landMirror` and absorb there.
- `mergePaths` (task_run.go:5996) is the existing ordered-dedup helper — reuse it.
- Emptiness law: a part that wrote nothing absorbs nothing.

<details>
[...]

PR #280, opened 2026-09-01
task families: a real-model end-to-end lane on both grounds

santoshkumarradha:
Part of #228. A real-model end-to-end suite, `internal/e2e/families_e2e_test.go` (build tag `e2e`), that proves task families work on a git repository **and** on a plain folder. Every assertion is on disk — the person's own directory, the family tree's git history, the session's task records, the division records in the workers' own journals. Nothing reads the model's prose.

## The four scenarios

| Subtest | What it proves | Issues |
| --- | --- | --- |
| a three-section report on a folder ground | the mirror is a repository of its own under the session's `trees/` with a baseline commit; each part in a worktree cut off it, on a `task/` branch, never in the parent's; each merged part's file arrives on its own commit in that tree; the family's ledger holds what the parts landed and every file on it is in the person's folder; the folder never gains a `.git` | #230, #229 |
| a two-part write-up on a repository ground | the same isolation and ledger laws where git was always underneath, plus **one** landing onto the person's branch, plus the divide-time checkpoint | #229, #232 |
| two parts that claim one file | the division is refused at admission with `refused:scope` and zero parts admitted, the refusal names `report.md` in the worker's own journal and says nothing was spent, no part is born, and the worker carries on to a settled task with the deliverable written | #231 |
| an edit made in the folder while it ran | the person saves `a.md` in their own folder after the family took its copy; the landing stands back — `conflicted`, their bytes untouched, no sibling laid either, the family's own version in the directory the refusal names, and the sentence names the file | #258 / #270 |

## How #232 is read

Off the parts' own commits, never out of their prose: the commit a part's file arrived on is the tip of the branch that part worked on, so `git show <commit>:NOTES.md` asks exactly the issue's question — was the parent's on-disk work under the part when it started. A token minted per run makes the answer unguessable. **Both halves are required**: the family history must hold the checkpoint (matched on the invariant half of its subject, `before its parts were handed out`) *and* the plan file must stand in each part's commit tree — a checkpoint that carried nothing would pass the first alone, and a part that wrote the plan itself would pass the second alone.

## How "one landing" is read

By the person's branch's own **reflog**, not by merge commits. A three-part family showed why counting merges is wrong twice over: a part's merge into the *family's* branch becomes reachable from the person's HEAD the moment the family lands, so more parts read as more landings — while a landing that fast-forwards writes no merge commit at all. The claim is that their ref moved once, and a reflog is the reading of exactly that.

## Pinning and cost

Every model row is pinned at `deepseek/deepseek-v4-flash` — the four tiers, `task.model`, eighteen text roles, `models.fallbacks` — and the pin is **checked**: each scenario reads the machine's own usage ledger back and fails if a second model answered. The four capability-qualified media roles are left alone, which is `--one-model`'s own reasoning.

Per scenario the suite logs wall time, cost (summed off `usage.jsonl`, auxiliary calls and every worker at every depth included), the models that answered, and which reader armed the split.

```
--- PASS: TestFamilies (581.31s)
    --- PASS: .../a_three-section_report_on_a_folder_ground (166.08s)  $0.027
    --- PASS: .../a_two-part_write-up_on_a_repository_ground (258.14s) $0.048
    --- PASS: .../two_parts_that_claim_one_file (52.03s)               $0.003
    --- PASS: .../an_edit_made_in_the_folder_while_it_ran (105.04s)    $0.012
```

About ten minutes and nine cents for the whole file.

## What is a finding rather than a red

A worker is told never to commit. One run answered with a part's file arriving on `y.md covering the last four topics (5-8)` — the worker had run `git commit` inside its own worktree. It had still worked in isolation and still merged back, which is everything the isolation and landing laws claim, so the lane logs it and holds its reds for the claims the wave actually makes. Likewise a part whose branch conflicts is named in the log; what must never happen — a file that came back and then did not ship — stays hard.

## Two departures from a person's own launch, both deliberate

- **`AFORGE_SPLITGATE=0`.** internal/splitgate refuses a division whose evidence names fewer than six separate items — a finding about whether handing work out *pays*, which this lane is not asking. Three sections and two write-ups are the smallest families that can prove isolation and landing. It is the switch `cmd/aforge`'s own tests already use.
- **`TaskAudit` off.** What this lane measures is where a part works and how its work comes home. Leaving the verified frontier on would put a cheap checker's opinion between a finished part and the person's folder, so a red here would say nothing about isolation or landing.

## Re-run

```sh
go test -tags e2e -count=1 -timeout 45m -v -run TestFamilies ./internal/e2e/
```

Needs `OPENROUTER_API_KEY`; skips green without it, as the rest of the tagged package does.


santoshkumarradha:
Rebased onto `dev` at ea25a27a — which now holds #229, #230, #231 and #233's part-brief composition (#240) plus the #251 wording scrub. Re-ran the suite against it, green a third time:

```
--- PASS: TestFamilies (312.97s)
    --- PASS: .../a_three-section_report_on_a_folder_ground (104.05s)  $0.030  3 parts, all 3 merged back into the family tree
    --- PASS: .../a_two-part_write-up_on_a_repository_ground (154.09s) $0.026  2 parts, both merged, one landing
    --- PASS: .../two_parts_that_claim_one_file (54.82s)               $0.002  refused:scope, 0 admitted, worker carried on
```

#232's divide-time checkpoint is still not in (it is #262), so scenario 2 logs `PENDING #232` for both parts rather than failing — a part cut off the ground's HEAD has no `NOTES.md` in the tree of its own commit. That assertion turns hard by itself the moment a `wip:` checkpoint appears in the family history; nothing here needs changing when #262 lands, only re-running.

**Heads-up, not this PR:** CI's packed-corpus job is red on `dev` itself and reproduces with this branch checked out away —

```
--- FAIL: TestTheChatManualAnswersTheQuestionsPeopleAsk
    chat_test.go:1409: "what decides whether work gets split" should reach tasks;
      it reached [how-tasks-run standing-orders adaptive-runs keeping-an-eye]
```

It is not in `.github/known-red.txt`, and it will fail every PR opened against `dev` until whoever moved that wording (#240 / #251) puts it back within reach of the `tasks` page.

santoshkumarradha:
Rebased onto `dev` at a5877146 (#262 and #270 both in) and re-run. Head is `a9dbf560`. Green twice on this exact tree, four scenarios:

```
--- PASS: TestFamilies (581s)              |  --- PASS: TestFamilies (661s)
  a three-section report on a folder    166s $0.027  |  196s $0.047
  a two-part write-up on a repository   258s $0.048  |  274s $0.018 (2nd attempt)
  two parts that claim one file          52s $0.003  |   54s $0.002
  an edit made in the folder while it ran 105s $0.012 |  137s $0.010
```

**#232 is live and hard now.** The PENDING branch is gone. The repository scenario requires both halves and gets both:

```
the family history holds the divide-time checkpoint "the work so far on plan notes, before its parts were handed out"
the commit x.md arrived on carries the parent's NOTES.md with its plan token
the commit y.md arrived on carries the parent's NOTES.md with its plan token
```

**#270 has a scenario of its own** — the person saves `a.md` in their folder after the family took its copy (`awaitFamilyTree` waits for `ground-baseline.json` beside the tree, which is the line an edit has to fall after to mean anything). The landing stands back: `merge=conflicted`, their bytes untouched, no sibling laid either, the family's own version in the directory the refusal names, and the sentence naming `a.md`. Scenario 1 stays green with the manners check in, as it must — nothing edits that folder mid-run.

Two failures on the way, both autopsied and neither a harness fault:

- **Test fault** — "one landing" counted merge commits. A three-part family exposed it: a part's merge into the *family's* branch becomes reachable from the person's HEAD the moment the family lands, so more parts read as more landings, while a fast-forward landing writes no merge commit at all. It is the branch's own reflog now: their ref moved once. (`main moved 2 time(s): [merge task/…: Fast-forward, commit (initial): the material]`)
- **Model fault** — a part's file arrived on `y.md covering the last four topics (5-8)`: the worker ran `git commit` inside its own worktree, which the prompts tell it never to do. It had still worked in isolation and still merged back, so the subject line is logged as a finding and the reds stay on the claims the wave makes.

PR #282, opened 2026-09-01
manual: the section on splitting work carries the question people ask it with

santoshkumarradha:
Follow-up to #270, which turned `dev` red on the manual gate the moment it landed.

## What is wrong on `dev` today

`go test ./internal/manual/ -count=1` at `c41545aa`:

```
chat_test.go:1409: "what decides whether work gets split" should reach tasks;
    it reached [how-tasks-run standing-orders adaptive-runs keeping-an-eye]
```

**It is not that #270's words are a magnet.** The section it added is nowhere in the top
eight for that question. What happened is that the probe was never really answered: on clean
`dev` the `tasks` page reached that question **fourth of four**, a tie at the last slot, and
the section that actually answers it — *When a task turns out to be too wide for one worker*,
the one holding the two tests a split has to pass — sat **sixth**. BM25 scores against the
corpus average length and each term's rarity, so adding a section *anywhere* moves every
score a little, and a fourth-place tie is a thing that falls over.

## What this does

The page, never the probe. The asker's own words go into the `## ` heading of the section
that answers them — headings are the search index, and they count three times:

```
## When a task turns out to be too wide for one worker — what decides whether work gets
   split, a task that splits itself, dividing work, parts of a task
```

That question now reaches it **first**, ahead of the `how-tasks-run` section that used to
take the slot, by a margin that does not depend on what else the corpus happens to hold. The
three cross-references to the section quote the phrase before the em dash and are unchanged.

## And the packed corpus, which #270 owed and did not pay

`internal/manual/chat.pack.gz` is committed here (it is tracked, despite the `.gitignore`
line, and every manual PR carries it). #270 edited `how-tasks-run.md` without regenerating
it, so **`dev` currently ships a binary whose manual does not contain the folder-landing
section at all** — `make embed` on a clean `c41545aa` checkout dirties the pack, which is
the proof. Regenerated here; `make test-packed-manual` passes.

## Tests

```
go test ./internal/manual/... -count=1        ok
make test-packed-manual                       ok
go test ./internal/tui3/ -run Manual          ok
go test ./internal/session/ -count=1          ok
```

https://claude.ai/code/session_01P1Ez7trmDzB8mGoWvuqwJZ


santoshkumarradha:
Overtaken: this was opened when #270's landing left the manual gate red ('what decides whether work gets split' stopped reaching tasks). On current dev (32fd51cf) go test ./internal/manual/ is green — the corpus reshaping from #245 and #251 restored the probe — so the premise is gone and this branch's chat.pack.gz is now stale. Closing; if the probe regresses, the +15 section here is the ready-made fix.

PR #296, opened 2026-09-02
surface: a selection covers every cell it sweeps, code spans included (#294)

### LINKED ISSUE #294: The selection highlight stops at the first inline code span — a sweep lights 4 cells of 64
## What happened

Observed 2026-09-01 at `dev@701d6d12`, in `bin/aforge` on a TrueColor terminal.

Sweeping the mouse across an assistant answer highlights only the text **before**
the first inline code span. Everything from the code span to the end of the row
draws with no highlight at all. Across a multi-line sweep the selection reads as
a set of disconnected bands with holes in it — the status line says
`copied · 5 lines` while roughly a line and a half appears lit.

A sweep across this row — every cell of it swept, all 64 selected and all 64
copied — paints **4 cells**:

```
The internal/swepro/ directory was deleted, along with the rest.
^^^^  <- the only cells that wear the selection
```

The bytes the terminal is handed, from `a.frame()`:

```
\x1b[48;2;67;76;94m                          <- the mark's ground opens
\x1b[38;2;198;205;218mThe \x1b[39m
\x1b[48;2;27;27;37m                          <- the code span sets ITS OWN ground
\x1b[38;2;198;205;218minternal/swepro/\x1b[49;39m   <- and CLEARS the ground entirely
\x1b[38;2;198;205;218m directory was deleted, along with the rest.\x1b[39m
                \x1b[49m
```

Two separate failures in one row:

1. The inline code span **sets** its own background, so those cells wear the
   code-span plane instead of the selection — the span reads as a hole the
   selection skipped.
2. The code span **clears** the background on its way out (`\x1b[49;39m`), which
   switches the ground off for the whole rest of the row. This is the worse of
   the two: a row that *opens* with a code span lights **0 of 20 cells** — a
   selection that is completely invisible.

The clipboard is correct throughout. `dragYank` reads `ansi.Strip`ped plain
text, so what lands is exactly what was swept. **Only the paint lies** — which
is the whole damage, because the person has nothing but the paint to tell them
what they are about to copy, and the manual promises them "What is highlighted
is exactly what is copied, to the character" (`internal/manual/chat/keys.md`).

Not limited to the sweep. Every ground step composes the same way, so the same
hole appears in:

- **copy mode** (`ctrl+s`) — `copymode.go` marks whole rows: 4 of 42 cells lit
  on a row with one code span.
- **`palette.selected`** — the pointer/selected step on every list that draws
  rows carrying chips or code spans (`place_memory.go`, `place_spend.go`,
  `place_search.go`, `rewind.go`, `task.go`, `crew.go`, `effortchip.go`).
- **`pastechip.go`**, which nests the two steps outright:
  `a.pal.mark(a.pal.chip(token), …)`.

## Replication

**Deterministic (no model).** Nothing but the repo and Go:

```sh
go test ./internal/tui3/ -run 'TestAGroundStepCoversEveryCellItIsGiven|TestASweepOverACodeSpanLightsEveryCellItCopies' -count=1 -v
```

At `dev@701d6d12` (before the fix) this prints, for a real sweep driven through
`Update` and read back off `a.frame()`:

```
the sweep lit 4 of the row's 80 cells: the highlight stops at the inline code span
```

and, for the palette step directly:

```
a row with one inline code span / mark: 9 of 27 cells wear the ground
a row that OPENS with a code span / mark: 0 of 20 cells wear the ground
a row with two code spans / mark: 0 of 31 cells wear the ground
```

The tests build the row the same way prose does — `palette.chip` for the code
span — and count cells with a small SGR walker (`litCells`) that tracks which
background is actually standing, because the sequence at fault is the
**compound** `\x1b[49;39m` and a naive scan for `\x1b[49m` never sees it.

**Field (real models).** No key and no money needed — this is a pure render
path. By hand: `make build && bin/aforge`, ask anything whose answer contains a
backticked path, and sweep across that line. The failure is on the first frame.

## Where

- `palette.background` in `internal/tui3/styles.go` — the root. It composes one
  background code, the text, one reset, which is only true of text carrying no
  background of its own.
- Its three callers on the ground ladder in the same file: `palette.mark`,
  `palette.selected`, `palette.cursor`.
- `app.markCells` in `internal/tui3/dragselect.go` (search `markCells`) — the
  sweep's painter, and the door `view.go` calls.
- `copymode.go`, search `a.pal.mark(text, width)`.
- The inner ground comes from `renderer.codeSpan` in
  `internal/tui2/prose/inline.go` (search `next.ground = true`), emitted by
  `tokens.ResetBg` / the `layerBg` arms of `internal/tui2/tokens/profile.go`.

## The fix

**A ground step owns every cell it covers.** Rewrite every background parameter
inside the wrapped span to the step's own: an inner set becomes the step's
ground, an inner clear becomes the step's ground, and a full reset keeps its
reset and re-lays the ground behind it. Inks and attributes are untouched, so a
marked code span keeps its colour, its bold and its italic — only the plane
under it is made one.

**The selection wins over the chip**, and that is the ruling rather than a
side effect. The mark step is the loudest rung of the ground ladder; a raised
plane inside it says "this cell is a different kind of thing" at exactly the
moment the person is being told "these cells are the ones". Every terminal's own
selection replaces the background it sweeps over, and this one should read the
same.

Every form the styler can emit has to be answered, not just the obvious one:
`48;5;n`, `48;2;r;g;b`, the sixteen-colour `40`–`47` and `100`–`107`, the bare
`49`, the compound `49;39` that prose actually emits, and `0` (which a terminal
also spells as no parameters at all).

## Acceptance

- **e2e:** `TestASweepOverACodeSpanLightsEveryCellItCopies` drives a real press
  and sweep through `app.Update` and reads the row back out of `a.frame()` — the
  bytes the terminal is handed. It asserts every cell of the swept row wears the
  selection's ground, padding included: `litCells(row) == ansi.StringWidth(ansi.Strip(row))`,
  80 of 80 where today it is 4.
- **e2e:** `TestTheSweepCopiesExactlyWhatItLit` releases the same sweep and
  asserts the OSC 52 payload is
  `The internal/swepro/ directory was deleted, along with the rest.` — the
  control, so a future change cannot make the highlight honest by making the
  copy wrong.
- **Unit:** `TestAGroundStepCoversEveryCellItIsGiven` holds both `mark` and
  `selected`, at TrueColor and ANSI256, over five row shapes (one code span, a
  row opening with one, two spans, a row that is only a span, and a plain
  control) to full coverage. `TestRegroundAnswersEveryBackgroundForm` pins every
  SGR form above, including `49;39` and the empty-parameter reset.
  `TestHoldGroundLeavesPlainTextAlone` and `TestAGroundStepLeavesTheInkAlone`
  hold the two things that must NOT change.
- `internal/manual/chat/keys.md` gains the ruling under "Selecting text with
  your mouse", and the change entry's `invalidates` names what people believed
  before: that the sweep's highlight already covered what it copied.

santoshkumarradha:
Closes #294.

## What was wrong

A ground step — `palette.mark`, `palette.selected`, `palette.cursor` — was one
background code, the text, and one reset. That is only true of text carrying no
background of its own, and the transcript is full of text that does: an inline
code span sits on the surface's raised plane, and it both **sets** a ground and
**clears** it on the way out with the compound `\x1b[49;39m`.

Nested inside a selection, the set turned the span into a hole the highlight
visibly skipped, and the clear — the worse of the two — switched the ground off
for the whole rest of the row. Measured at the door, off the bytes `a.frame()`
hands the terminal:

| Row swept | Cells lit, before | After |
| --- | --- | --- |
| `The internal/swepro/ directory was deleted, along with the rest.` | **4 of 80** | 80 of 80 |
| a row that *opens* with a code span | **0 of 20** | 20 of 20 |
| a row with two code spans | **0 of 31** | 31 of 31 |
| copy mode (`ctrl+s`) over a row with one span | **4 of 42** | 42 of 42 |

The clipboard was always right — `dragYank` reads `ansi.Strip`ped text — so this
was the paint lying about a copy that was correct. That is the whole damage: the
paint is all a person has to tell them what they are about to take, and
`keys.md` promised them "What is highlighted is exactly what is copied, to the
character."

## The change

`palette.background` rewrites every background parameter inside the span it
wraps to its own ground: an inner set becomes the step's ground, an inner clear
becomes the step's ground, and a full reset keeps its reset and re-lays the
ground behind it. Every form the styler can emit is answered — `48;5;n`,
`48;2;r;g;b`, the sixteen-colour `40`–`47` and `100`–`107`, the bare `49`, the
compound `49;39` prose actually emits, and `0` spelled as no parameters at all.

Inks and attributes are untouched, so a marked code span keeps its colour, its
bold and its italic; only the plane under it is made one.

**THE SELECTION WINS OVER THE CHIP**, and that is the ruling rather than an
accident of implementation. The mark is the loudest rung of the ground ladder; a
raised plane inside it says "this cell is a different kind of thing" at exactly
the moment the person is being told "these cells are the ones". Every terminal's
own selection replaces the background it sweeps over, and this one now reads the
same.

Because it lands on the shared step, everything drawn with it comes back at
once: the sweep, copy mode's span, the pointer's row on every list carrying
chips or code spans (memory, spend, search, rewind, task, crew, effortchip), and
`pastechip`'s `mark(chip(token))`, which nested the two outright.

Text with no escape in it returns untouched and unallocated, which is most of
what this is handed.

## Tests

`internal/tui3/holdground_test.go`, new:

- **e2e** `TestASweepOverACodeSpanLightsEveryCellItCopies` — drives a real press
  and sweep through `app.Update`, reads the row back out of `a.frame()`, and
  asserts every cell wears the selection's ground, padding included.
- **e2e** `TestTheSweepCopiesExactlyWhatItLit` — the control: releases the same
  sweep and asserts the OSC 52 payload, so a future change cannot make the
  highlight honest by making the copy wrong.
- **Unit** `TestAGroundStepCoversEveryCellItIsGiven` — `mark` and `selected`, at
  TrueColor and ANSI256, over five row shapes to full coverage.
- **Unit** `TestRegroundAnswersEveryBackgroundForm` — every SGR form above.
- **Unit** `TestHoldGroundLeavesPlainTextAlone`, `TestAGroundStepLeavesTheInkAlone`,
  `TestAGroundStepPaintsItsPaddingToo` — the things that must NOT change.

The cell counter (`litCells`) walks SGR the way a terminal does, because the
sequence at fault is compound and a naive scan for `\x1b[49m` never sees it —
which is why this survived review the first time.

```
go test ./internal/tui3/ -count=1     ok  483.357s
go vet ./internal/tui3/               ok
go build ./...                        ok
```

Full `internal/tui3` suite green, so no snapshot asserting the old escape stream
broke.

## Manual

`internal/manual/chat/keys.md` gains "The highlight covers inline code too."
under *Selecting text with your mouse*, stating the ruling in the person's own
words.

https://claude.ai/code/session_01B25XERftiFQW619ctN4kRg


santoshkumarradha:
Superseded by #303, which fixes the same root cause correctly where this branch does not.

Both branches rewrite inner background parameters in `palette.background`, and the ruling is the same. The difference is sub-parameter handling: my `reground` walks a parameter list flatly, so a true-colour **foreground** whose channels collide with background parameter numbers gets mangled — `reground("38;2;0;49;100", "48;5;7")` returns `38;2;0;48;5;7;48;5;7;48;5;7`. That is not an edge case: it hits every colour with a zero channel. #303's `regroundParams` consumes `38`/`58`'s sub-parameters and gets it right.

#303 also already carries every test named in #294's acceptance, under the same names, including the two end-to-end ones through `a.frame()` and the OSC 52 control. I have left the one case it does not yet pin as a comment there.

Closing unmerged; nothing here is worth carrying over. Issue #294 stays open for #303.

PR #303, opened 2026-09-02
surface: a ground step owns every cell it covers — a sweep lights code spans and chips (#294)

### LINKED ISSUE #294: The selection highlight stops at the first inline code span — a sweep lights 4 cells of 64

santoshkumarradha:
## What this fixes

A sweep across a row carrying an inline code span lit the cells before the span and nothing after it — 4 of 64 cells on the row the field failure was seen on — while the clipboard still took the whole row. The paint lied about the copy.

`palette.background` was one background code, the text, one reset. That is true only of text carrying no background of its own. An inline code span sets its own ground (`48;2;…` on the raised plane) and closes with the compound `49;39`, so the span's ground took the plane off the step, and the compound clear switched the step's ground off for the rest of the row. A row that opens with a code span lit nothing at all.

## The fix

`holdGround` walks the wrapped text and re-lays the step's ground under every cell: an inner background set becomes the step's ground, an inner background clear becomes the step's ground, and a full reset keeps its reset and re-lays the ground behind it. Every form the styler can emit is answered — `48;5;n`, `48;2;r;g;b`, the sixteen-colour `40`–`47` and `100`–`107`, the bare `49`, the compound `49;39` prose actually emits, and `0`, which a terminal also spells as no parameters at all. Inks and attributes are untouched, so a marked code span keeps its colour, its bold and its italic.

THE SELECTION WINS OVER THE CHIP, and that is the ruling rather than a side effect. The mark step is the loudest rung of the ground ladder; every terminal's own selection replaces the background it sweeps over, and this one reads the same.

One root, three callers healed at once: the sweep's mark (`palette.mark`), copy mode's selection (`palette.selected`), and the pointer's row (`palette.cursor`).

## Tests

`internal/tui3/ground_test.go`:

- `TestASweepOverACodeSpanLightsEveryCellItCopies` — the e2e: a real press-and-sweep through `app.Update`, read back off `a.frame()`.
- `TestTheSweepCopiesExactlyWhatItLit` — the control: the OSC 52 payload is exactly what was lit, so the highlight cannot be made honest by making the copy wrong.
- `TestAGroundStepCoversEveryCellItIsGiven` — mark and selected, TrueColor and ANSI256, over five row shapes (one span, leading span, two spans, only a span, plain control).
- `TestRegroundAnswersEveryBackgroundForm` — every SGR form above.
- `TestHoldGroundLeavesPlainTextAlone` — no escape, no rewrite, no growth.
- `TestAGroundStepLeavesTheInkAlone` — foreground, underline and attribute codes keep their codes and their resets.

The manual ruling lands in `internal/manual/chat/keys.md` under "Selecting text with your mouse", and the change entry's `invalidates` names what people believed before: that the sweep's highlight already covered what it copied.

Closes #294


santoshkumarradha:
I had a parallel fix for #294 open as #296 — same root cause, same ruling. Yours is better and I've closed mine in favour of it, so this is just the one thing I found that your branch does correctly but does not yet pin.

**A colour's sub-parameters are not parameters.** `38;2;0;49;100` is the foreground `rgb(0,49,100)`: its channels collide with the full reset (`0`), the background clear (`49`) and a bright sixteen-colour background (`100`). A walker that reads them as parameters corrupts the ink of any text painted in such a colour — and that includes **every colour with a zero channel**, which is common rather than exotic. My version had exactly this bug; `reground("38;2;0;49;100", …)` returned `38;2;0;48;5;7;48;5;7;48;5;7`.

`regroundParams` gets this right because it consumes `38`/`58`'s sub-parameters via `colourSubs`. I ran this against `origin/fix/294` and it passes — it is coverage, not a fix:

```go
// A colour's SUB-PARAMETERS are not parameters. 38;2;0;49;100 is the
// foreground rgb(0,49,100): its channels collide with the full reset (0), the
// background clear (49) and a bright sixteen-colour background (100), and a
// walker that reads them as parameters corrupts the ink of any text painted in
// a colour with one of those channel values — which includes every colour with
// a zero channel.
func TestAColoursSubParametersAreNotParameters(t *testing.T) {
	const ground = "48;2;67;76;94"
	for _, tc := range []struct{ in, want string }{
		{"a\x1b[38;2;0;49;100mb", "a\x1b[38;2;0;49;100mb"},
		{"a\x1b[38;5;49mb", "a\x1b[38;5;49mb"},
		{"a\x1b[58;2;0;0;0mb", "a\x1b[58;2;0;0;0mb"},
		{"a\x1b[1;38;2;0;40;47mb", "a\x1b[1;38;2;0;40;47mb"},
	} {
		if got := holdGround(tc.in, ground); got != tc.want {
			t.Errorf("holdGround(%q) = %q, want %q", tc.in, got, tc.want)
		}
	}
}
```

Worth adding to `ground_test.go` next to `TestRegroundAnswersEveryBackgroundForm` — it is the case that separates a correct walker from a plausible one.

Two other things from my branch you may want, both already covered by yours as far as I can tell: the full `internal/tui3` suite is green with this change (483s, no snapshot asserted the old escape stream), and `internal/manual/chat/keys.md` needs the ruling — which your diff already has.

Issue #321, opened 2026-09-02
manual: whether the model opens the manual at all is the dominant term — it swings up to nine of twenty-five plain questions per run


santoshkumarradha:
## Finding first (2026-09-02, from the measured cue-prepend experiment on draft PR #324 — read this before anything below)

**Shown that an answer exists, the cheap model does not go and get it, and the clause invites it to pretend it did.** With the top matching manual section titles placed beside the person's question, the model opened the manual no more often than without them (16, 25, 17 of 25 across three passes against a same-night control of 21), and on turns where it did not open the manual it sometimes answered as if it had — "That's the manual you found", "Yes. The manual covers memory in three sections:" — with no lookup made: four fabricated lookups in the first pass, two in each of the others, none in the control.

A mechanism that makes a model claim it consulted a source it never opened is worse than the gap it was meant to close. This is a finding about cue-style prompting in general, not about this corpus: anyone reaching for the same shape elsewhere in this product — surfacing evidence beside a message in the hope the model acts on it — needs to meet that sentence first. The cue is not landed; the branch and its measurement are kept as evidence.

Two facts from the same tables that stand on their own: reachability on turns that did open the manual is 100% both with and without the cue, so the retrieval work of #293/#307/#313 is finished and that measure has stopped discriminating; and the same-night control sits above two of the three cued passes.

**Open question, parked for a cost decision (not a lane):** whether a stronger model opens the manual reliably on the same corpus and questions. If yes, this is a model-choice problem and no mechanism fixes it for the cheap one; if no, the gap is structural. Both remaining candidates below assume that answer without measuring it.

---

Follow-up to #293 and #307. Measured on the wire by the #307 lane (PR #313), deepseek/deepseek-v4-flash pinned on every role, five passes of the 25 plain questions and the 22 held-out questions through the live chat surface.

## The measurement

| pass | opened the manual (25-set) | opened (held-out) |
|---|---|---|
| without #313, run 1 / 2 | 19 / 17 | 18 / 17 |
| with #313, run 1 / 2 / 3 | 19 / 20 / 22 | 17 / 13 / 17 |

[...]

PR #324, opened 2026-09-02
manual: a question about aforge arrives with the manual's own titles beside it

santoshkumarradha:
**Refs #321. The measurement is a NULL and the acceptance is not met: cue-prepend did not move the open rate.** This is not ready to merge on its numbers; it is here so the mechanism, the measurement and the reason are all one thing somebody can act on.

## What was built

`internal/session/manual_cue.go`. When the person's message reaches for the manual's own vocabulary, the top four matching section titles — heading plus an opening clause each — are attached beside their message for that turn's requests. Evidence, not an instruction: the tool description and `prompts/system.md`'s tool policy are **untouched**, because both already ask and the measurement says asking is not what is missing. What the model cannot know is that an answer **exists** — this corpus is a few dozen sections nothing in its training data has seen.

### The design decisions, and why

**Where it rides: beside the person's message, on the COPY the requests are made from.** Three candidates — a system-side turn preface, an ephemeral line beside the message, the tool's own result slot. The result slot is out by construction: it needs the tool call this whole issue is about. The system preface is out because `message[0]` is the cached prefix (`a.system + a.standingText + a.memoryText`), and a cue there would disturb it on every cued turn of every conversation. So it goes beside the message, at the moment the decision is made — and on the copy rather than in the transcript, because a hint about a question already answered is prompt nobody should buy twice.

**It goes through the one turn seam.** `Agent.cueTurnLocked` is called from `startTurnLocked`, the single place a turn starts, so every road a person's words take is cued once and there is no second answer to what a turn is carrying. `Agent.withManualCue` attaches it where `snapshotWithReasoning` already hands the loop its own copy, and refuses on any doubt — a message no longer at that index, no longer the person's, or no longer the sentence the block was matched against is a transcript a compaction rewrote under this turn.

**Nothing else moves.** `TestTheFixedPrefixStaysUnderItsBudget` reads **47,843 bytes before and after** — the same figure #313 landed. The transcript, the journal, a resume, an export and `Agent.taskRequest` are all exactly what they were, and `TestTheCueNeverReachesTheJournalOrTheTranscript` and `TestTheCueDoesNotBecomeWhatThePersonAskedFor` hold that. Bounded at four sections, a 120-byte clause each and a 1000-byte block; absent inside a task node, on a woken turn, on the session's own notes, and on a marked standing draft.

**The gate is `manual.Chat().Cued`** — the corpus's own derived vocabulary, already the resident's self-question trigger — so there is one answer in this repository to "is this about the product itself" rather than two that drift. **It is loose and the number is written down:** it fires on 23 of the 25 plain questions, 21 of the 22 held out, and **8 of 14 ordinary work sentences that are not about aforge at all**, because this manual describes a general-purpose program and its headings are the words work is asked for in. `TestOrdinaryWorkAlsoCuesAndThatIsTheCost` holds that measurement. A false one costs under a kilobyte of prompt, once.

## On the wire

`TestManualOnTheWire`, `deepseek/deepseek-v4-flash` pinned on every role and verified off the usage ledger, one fresh conversation per question, 2026-09-02. Three complete passes with the cue, one complete pass without it **on the same night, the same build and the same code path** — `AFORGE_WIRE_NO_CUE=1` sets `Config.ManualCueOff`, which exists for exactly this.

| pass | 25-set opened | first | within four | held-out opened | first | within four | wall | ran to completion |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| **with** the cue, 1 | **16/25** | 14 | 16 | 12/22 | 7 | 9 | 9m58s | yes |
| **with** the cue, 2 | **25/25** | 23 | 25 | 18/22 | 8 | 16 | 18m54s | yes |
| **with** the cue, 3 | **17/25** | 16 | 17 | 19/22 | 8 | 16 | 15m0s | yes |
| **without**, same night | **21/25** | 14 | 21 | 14/22 | 6 | 13 | 11m5s | yes |
| — an earlier with-cue pass | — | — | — | — | — | — | 26min silent | **NO — casualty, pooled with nothing** |
| #313's without-cue baseline | 19/25 · 17/25 | 11 · 14 | 18 · 16 | 18/22 · 17/22 | 5 · 5 | 12 · 16 | | (quoted, not re-run) |

**Read it plainly.** The acceptance wanted 23/25 or better on three consecutive passes with reachability holding at 95% of opened turns. It got **16, 25, 17** — and the pass measured without the cue on the same night, **21/25**, sits above two of the three. Sixteen and twenty-five on identical code in consecutive passes is #321's own finding, unchanged: whether the model reaches for the tool at all is still the dominant term, and this did not take it off the table.

Reachability normalised to the turns that did look something up is the half of the acceptance that passes, on a measurement that no longer discriminates: **58 of 58** on the 25-set with the cue (16/16, 25/25, 17/17) against **21 of 21** without — both 100%. Held out, 41 of 49 with against 13 of 14 without.

### Why it can hurt, in the model's own words

On turns that never opened the manual, the model reads the block as a lookup that **already happened**:

- `what is aforge` → "aforge is the program running this session … That's the manual you found", no tool call.
- `what does it remember` → "Yes. The manual covers memory in three sections:", no tool call.

Four such replies in pass 1, two in each of passes 2 and 3, **none in the pass without the cue**. Titles and an opening clause are enough to answer from and not enough to be right from — which is the exact failure `tools_manual.go` exists to prevent, arrived at from the other side. **The obvious next experiment is titles with no clause at all, and possibly a line saying the text has not been retrieved. It is not run here**, because the ruling was cue-prepend first and nothing else until it is measured, and it is now measured.

### Spend

Not recoverable exactly for these passes: no question breached the lane's `$0.05` cap in any of them, and the ledger `ledgerSince` reads lives under the throwaway home the test deletes. At 10–19 minutes a pass against #313's $0.15–0.20, each is in the low tens of cents and the five passes together are a few dollars. **The lane now sums it and logs `THIS PASS SPENT $…`**, so the next pass reports it instead of estimating it.

### The casualty

One earlier with-cue pass went silent at question 22 of 25 and stayed silent for 26 minutes on 27s of CPU. `SIGQUIT` named an **ABBA deadlock between `hedgeRace.mu` and `streamWatch.mu`** in `internal/provider` — `streamWatch.facts` takes `w.mu` and calls `race.askedLane` inside it, while `hedgeRace.spend` holds `race.mu` and walks every arm's `watch.lane()`. Nothing to do with this change; filed as **#330** with the stacks and a deterministic replication. Its partial numbers are pooled with nothing and are not in the table.

## Tests

Deterministic, in `internal/session/manual_cue_test.go`: cued vs not cued through a real turn, the block never reaching the transcript or the journal, the person's ask unchanged, the cue dying with its turn, the section/clause/byte caps over all 47 measured questions, an empty corpus answer, the gate agreeing with `manual.Chat().Cued`, a question asked over a picture, the ordinary-work cost, and absence inside a task and under the switch. **Nothing new is asserted about the wire** — the floors are #313's and the numbers above are a report.

`what-i-can-do.md` says what a person now gets, including that they never see the block.

```
go vet ./internal/session/ ./internal/manual/... && go test ./internal/session/ ./internal/manual/... -count=1
ok  github.com/Agent-Field/aforge-v2/internal/session  107.117s
ok  github.com/Agent-Field/aforge-v2/internal/manual   0.684s
ok  github.com/Agent-Field/aforge-v2/internal/e2e      0.219s      (untagged, the tuiwords gate)
```

Full suite red, all of it on `.github/known-red.txt`: `TestTheGuidesRender`, `TestReviewerCarriesTheWholeLaw`, `TestTheReviewerBriefCarriesBothDuties`, `TestBothBriefsSayTheSyntaxIsASCII`, `TestEveryLockInTheGuardedTreeUnlocksFromADefer`, `TestEveryGoroutineInTheGuardedTreeIsGuarded`, `TestRenderTerrainStaysUnderTheCap`, `TestEveryMessageWriteUsesThreadPost`, `TestSettingsSheetIsOneCalmColumnAtEveryWidth`.

## Codex review

Four findings, three taken. The block now joins the last **text** part, so a question asked over a screenshot still gets it (with a test). The gate's looseness is measured and held rather than left unsized. A sentence spliced into a running turn deliberately gets no cue and the file says why. **Declined:** a hand-written question/assignment-verb gate beside `Cued` — that is a second reading of "is this about aforge" drifting away from the corpus's own, and it is tuning invented in this lane rather than measured. **Declined:** asserting an opened-count floor in the committed wire test — the brief is that the wire numbers are a report and nothing is asserted beyond what #313 already asserts, and on these numbers such a floor would be red on a model's mood.

## The verdict

**It did not move the open rate.** Reachability when consulted is perfect and was already perfect. The mechanism is sound, cheap, bounded and invisible to the person, and it has one measured pathology: the block reads as a completed lookup. Do not merge on these numbers. The cheapest next step is titles alone, measured the same way against the same-night baseline this branch now makes possible.

https://claude.ai/code/session_01DLjRso4MaUPdvU8b2jP11t


santoshkumarradha:
Closed as superseded, not merged. Its own body says not to merge on its numbers (cue-prepend measured a null on the open rate and a fabricated-lookup harm; the finding sits at the top of #321); the manual change it carried was a paragraph with no heading and no probe; and #458 changed the scorer under it, so its before/after no longer measures the current retriever. Any future "titles beside the question" cue returns as a fresh PR measured against 199620c5's before/after with the fabricated-consultation rate reported alongside the open rate. #321 stays parked for the owner's model-versus-mechanism ruling; the branch and its logs remain the evidence.

PR #333, opened 2026-09-02
tasks: a task on an attached repository seals, and work aforge starts is named before it is announced

santoshkumarradha:
## What broke, from one run

A words-only turn was handed over at its ceiling, task 1 started, and 7.5 seconds later it failed with:

```
could not prepare a working copy: this task's world could not be sealed: The following paths are ignored by one of your .gitignore files:
```

It was announced and reported under the raw ask (`https://github.com/…/252 Can you look at this and …`), never under a name.

## Three faults, three fixes

**The seal.** `sealGroundWork` staged with `git add -A -- . ':(exclude).aforge-v3' ':(exclude).furrow'`. Git treats an exclude pathspec that literally names an ignored path as an attempt to add it and exits 1 after doing the work. `hideFurrowMarker` writes `.furrow/` into `.git/info/exclude` the moment furrow attaches a folder, so **every task grounded on an attached repository failed in its first second.** Reproduced with the same command on the repo. The corners are now taken back out with `git reset -q -- <paths>` on the private index, the way `stageTaskWork` already does. Same fix on the fork rung (`sealForkWorld`), which had the same latent shape for anyone whose `.gitignore` lists the corner.

**The namer's empty answer.** The `taskname` call went out at `max_tokens: 160` (32 for the answer ÷ the "high" share the model's profile runs at). `deepseek-v4-flash` ignores the reasoning disable, thought through all 160, and returned nothing (`empty_at_ceiling: true` in `calls.jsonl`). The quirks memo already knew the model, so `learnFromAnswer` did not ask again. A pass the caller never asked for now gets at least `unaskedThinkingFloor` (2048) tokens of room, whatever the answer's size; a word the caller chose keeps the router's share, and large answers are unchanged.

**The name arriving after the announcement.** The namer started at admission, after the 23-second brief-writing stage. Both roads that start work on their own (`route_judge.go` after the judge says work, `checkpoint.go` at the top of the handover) now ask for the name at the moment they decide (`nameAhead`), in parallel with the brief. A name in hand at admission is the title on the told-after line and the row, written as `named` so nothing renames it; one still in flight rides `taskSpec.ahead` and `nameNode` waits on it instead of making a second call; a road that declines cancels the call.

## Verification

- `go test ./internal/provider/` green; `./internal/session/` green except `TestAFolderLandingRefusesToWriteOverThePersonsOwnEdit`, which fails identically on a clean `origin/dev` (macOS `/var` vs `/private/var`), so it is not this change's.
- New tests: `TestASealCarriesAWorldWhoseFurrowMarkerIsHidden`, `TestAForkSealsAWorldWhosePrivateCornerIsIgnored`, `TestAPassNobodyAskedForGetsAPageToThinkIn`, `TestANameAskedAheadIsOnTheTaskWhenItIsAnnounced`, `TestANameStillInFlightAtAdmissionIsWaitedFor`, `TestARoadThatDeclinesLetsTheNameGo`.
- `routeCompleter` in the tests now answers the namer with nothing, because its `(errand)` placeholder would otherwise land as a name on the runs where the goroutine won the race.
- Manual: `tasks.md` naming section says work aforge starts on its own is usually named before it is announced. `go test ./internal/manual/` and the tui3 manual gate green.

https://claude.ai/code/session_01C7uqTHMSxgcdwcKchUyi6z


santoshkumarradha:
Note for the author from a flake investigation (#392, no action needed): on a dropped ceiling the early namer's low-tier call is still spent and only best-effort cancelled for a task that never starts. Small waste, not a defect; the flake itself was the test fixture, and #333 stays.

PR #337, opened 2026-09-02
lens L2: the task room adopts the conversation's reducer

santoshkumarradha:
Refs #252 — lane **L2** of the lens design (`docs/design/lens/DESIGN.md`): Decision 1
applied to the task page, Decision 4's event list, and ruling 3.

## What it does

`taskRoom` embeds the same `feed` the app does. The thirteen hand-copies in `room.go`
are gone; `roomEvent` keeps the collapse rule and the page's own after-effects
(`prefetchWritten` on a tool end, `closeLive` on a turn done, the error note) and calls
`feed.ingest` for everything else. `room.entries`, `room.live`, `room.think` and
`room.turn` still mean exactly what they meant — the fields moved house, not name.

Three facts a task's page did not have, and now does:

- **what a call took.** `EventToolFinished` was not in `room.go`'s switch at all.
- **a retry.** The dead attempt's text and forming rows go, and the reason is written
  down, on both surfaces.
- **a notice, a nudge, a guardian line.** All three were dropped on the floor in a room.

**Ruling 3** lands with it: `feed.closeTool` pairs args-match-then-oldest — the rule the
room already had — on both surfaces. The user-visible edge is two same-name tools running
at once: the result now resolves onto the call it is actually about.

Also from the review of #259: `newFeed` replaces four hand-spelled `-1` sentinels
(finding 5); announce and begin keep the `CallID` session offers them (finding 3);
`closeTool` re-resolves its row after the view's hook and `feedHooks.closed` states the
no-append invariant (finding 4); `settleCompaction` repaints (finding 6); `apply`
dispatches `ingest` unconditionally so one list is authoritative (finding 1);
`EventRetrying`'s four-call reduction is one `feed.retry` with the spawn-card half on a
hook (finding 2); `appendText` and `appendThought` are `feed.say` and `feed.reason`
(finding 8).

Findings 7 (`resolveUnfinished`'s dead forming arm) and 9 (the O(n) claim walks) are
**not** in this change — 7 is `app.dropForming`'s ordering and 9 is a settled-floor index,
both of which are their own change.

## The salience contract

`internal/tui3/salience_test.go`. A table walks **every** `session.EventKind` — enumerated
by reading `internal/session/session.go`'s own const block, so a kind added there and
forgotten here fails the build — through the chat's pump and a node's, and asserts the
room has a block for whatever the conversation drew.

> **THE LENS MAY LOWER SALIENCE; IT MAY NOT DROP A FACT.**

Twenty-eight exclusions carry their reason in the table, and each reason is itself tested:
an event a node cannot receive must leave a node's page untouched. Lane L3 extends the
same shape with the journal-role half.

## e2e

**No wire path.** The whole change is surface work over events the engine already sends,
so the gate is a *scripted* endpoint standing in for every model call
(`internal/e2e/roomfeed_e2e_test.go`, build tag `e2e`) — the real `bin/aforge`, a real
terminal, tmux, and a stub speaking the same OpenAI-compatible dialect
`internal/provider` writes and reads. The **approval** below is against the shipped
default model.

### 1. Scripted — a batch of two calls, and the figures are each call's own

```
go test -tags e2e -count=1 -run TestATaskPageShowsWhatACallTook ./internal/e2e/
```

Captured pane, on this branch:

```
─ ⠴ main ▸ run both probes · checking what it left · 4s · stub/scripted ─────────── esc/← main · ✕ ─
  ├─▶ bash sleep 1; echo PROBE-A                                          1.0s│ tasks
  ╰─▶ bash sleep 4; echo PROBE-B                                          4.0s▌ ⠴ ◇ run both probes  #1
                                                                              │   checking what it left
Both probes printed.                                                          │   4 · scripted
```

**The same run at `origin/dev` fails**, which is what makes it a gate rather than a
screenshot — a batch's results land together, so a page that timed a row from its own
begin to its own end writes the slowest call's span on every row in it:

```
  ├─▶ bash sleep 1; echo PROBE-A                                          4.0s│ tasks
  ╰─▶ bash sleep 4; echo PROBE-B                                          4.0s▌
  roomfeed_e2e_test.go:97: the row for PROBE-A says "…  4.0s", want "1.0s"
```

(Run from a throwaway detached worktree at `origin/dev` @ `69029c4e` with the same test
file copied in; the worktree is removed.)

### 2. Scripted — a retry inside a task

```
go test -tags e2e -count=1 -run TestATaskPageSaysWhenAnAttemptWasCut ./internal/e2e/
```

The node's second turn is answered by a stream that speaks once and then goes quiet until
the stream guard's mid-stream bound cuts it. Captured pane:

```
─ ? main ▸ run both probes · needs your look · 49s · stub/scripted ──────────────── esc/← main ─
  ├─▶ bash sleep 1; echo PROBE-A                                          1.0s│ tasks
  ╰─▶ bash sleep 4; echo PROBE-B                                          4.0s▌ ? ◇ run both probes  #1
                                                                              │   finished — look it over
  · the model went quiet mid-reply — asking again                             │ + /task

Both probes printed.                                                          │ + /standing
```

The dead attempt's words (`I will start by`) are asserted absent. At `dev` this room shows
neither the line nor the removal.

### 3. Approval — the real binary against the shipped default model

```
go test -tags e2e -count=1 -run TestATaskPageShowsWhatACallTookAgainstARealModel -v ./internal/e2e/
```

Model `~deepseek/deepseek-v4-flash-latest`, brief typed as `/task solo with the shell,
sleep for two seconds and list the files here; then read README.md; …`. Spend for the run:
**$0.0009**. Captured pane:

```
─ ⠏ main ▸ with the shell · working · 46s · ~deepseek/deepseek-v4-flash-latest ──── esc/← main · ✕ ─
  ⠿ thought for 0s · 27 tok · ctrl+e                                          │ tasks
                                                                              ▌ ⠏ ◇ with the shell   #1
  ╰─▶ bash sleep 2 && ls                                                  2.0s│   deepseek-v4-flash-latest
```

`--- PASS: TestATaskPageShowsWhatACallTookAgainstARealModel (74.80s)`

## Validation

- `go build ./... && go vet ./...` — clean. `gofmt -l internal/ cmd/` names only
  `internal/cachedir/cachedir_test.go` and `internal/tui3/ground_test.go`, both untouched
  by this branch.
- `go test ./internal/tui3/ ./internal/session/ ./internal/manual/ ./internal/e2e/` — all
  four `ok` (tui3 497.8s).
- The light gate of `docs/rules/ci.md` — build, vet, `make test-packed-manual`,
  `git diff --exit-code`, `make changelog-check` (114 entries well formed),
  `go test -run Manual ./internal/tui3/ ./internal/session/` — all green.
- `make check` — red only from nine tests, every one of them in `.github/known-red.txt`:
  `TestTheGuidesRender`, `TestReviewerCarriesTheWholeLaw`, `TestTheReviewerBriefCarriesBothDuties`,
  `TestBothBriefsSayTheSyntaxIsASCII`, `TestEveryGoroutineInTheGuardedTreeIsGuarded`,
  `TestEveryLockInTheGuardedTreeUnlocksFromADefer`, `TestRenderTerrainStaysUnderTheCap`,
  `TestEveryMessageWriteUsesThreadPost`, `TestSettingsSheetIsOneCalmColumnAtEveryWidth`.
- `make size` — `bin/aforge` 49,152,265 bytes, under the 54,600,000 budget.

## Manual

One new section at the **end** of `internal/manual/chat/reading-a-task-page.md` (appended
so L3's and L4's hunks stay apart): what a call took on a task's page, a retry inside a
task, and the three dim `· ` lines that now reach a room. `make embed` run; the generated
archives are ignored and not committed.

## Left for other lanes

A room's steer is still an ordinary line rather than an elbow (L3). `feedHooks.closed` is
left nil on a room with a one-line note that L4 installs the spend walk there. The journal
shaper, `steer()`, the header family, `roomRows`/`roomToolTail` and `render.go`'s deck are
untouched.

https://claude.ai/code/session_01Kb2QfDZAaN9cdcwCyczUay

## Note for whoever re-lands #437 (reveal pacing)

#437 added `e.catchReveal(len(text), a.linear)` inside `appendText`, `appendThought`, `roomSay` and `roomThink`. All four are gone after this change: the first two are `feed.say` and `feed.reason`, the last two are deleted mirrors. The re-land must call the reveal from the feed, once, for both surfaces: one hook on `feedHooks`, `snap func() bool` (the app installs `func() bool { return a.linear }`; the room installs the same, because the linear tier is process-wide and both pages owe it; a nil hook means pace), called from `feed.say` and `feed.reason` as `e.catchReveal(len(text), f.snap())`. Never re-add the call to a view function: the room would fall behind again, which is the defect this change exists to end.


santoshkumarradha:
Review (santosh-0a, review-only, at 70153b6d). **No blocker.** Two rebase notes for the PRs behind it, one nit.

**Fact-loss trace, per deleted function** (dev room.go:1183–1500 vs feed.go on this branch): roomSay→feed.say, roomThink→feed.reason, roomSettleThought→settleThought, roomCollapseThought→collapseThought, roomCloseLive→closeLive, roomFormTool→formTool, roomResolveUnfinished→resolveUnfinished, roomAnnounceTool→announceTool, roomBeginTool→beginTool, roomClaimAnnounced→claimAnnounced, roomClaimRunning→claimRunning, roomCloseTool→closeTool, roomSettleCompaction→settleCompaction. Every field the old body wrote, the feed body writes. Three deliberate deltas, all salience not fact: `collapseThought` now also shuts an unlatched open thought (the chat's rule); `resolveUnfinished` marks the row stale; `say`'s late-turn arm is inert in a room (`settledTurn` is 0). Six kinds gained (ToolFinished, Retrying, Nudge, GuardianAllowed, Notice, TaskReplyTags). The salience table is what keeps it that way. ✅

**Ruling 3 / seam contract.** `claimRunning` (END) pairs args→oldest and ignores `callID`; `finishTool` pairs callID→args→oldest. I checked the wire before calling that a defect: `EventToolEnd`/`EventToolFailed` carry no CallID (loop.go:1874, 1891) and results arrive in call order (session.go:357), so args→oldest is deterministic for END; only `EventToolFinished` carries the id (loop.go:1968) and finishes arrive out of order, so it needs it. The asymmetry is the wire's, not the PR's — worth one line in `claimRunning`'s comment saying so, since the comment currently argues from clipped args alone. The L3 journal row (callID + unparsed args, `replay.go:627`) is consumed correctly on both halves. ✅

**Changelog** states the user-visible edge of the pairing change (two same-name calls running at once). ✅

**Rebase notes for #351 and #355** (trial-merged locally, not pushed):
- #351 conflicts in room.go at four spots, all mechanical: `newRoom` + `replayBlocks(session.ReadTranscript…)`; take #351's deletion of the old reader; keep this PR's `roomNote`; take #351's elbow body with `room.collapseThought()` (the `a.roomCollapseThought` it calls no longer exists here).
- #355 conflicts in app.go at `feedHooks`: this PR adds `forming`/`closing`/`retrying` hooks and `newFeed(...)`; #355 rewrites the signature to `feedHooks(l lens)` and drops them. Keep all three hooks *and* the lens parameter.

**Nits (no action needed to merge):** `internal/e2e/roomfeed_e2e_test.go` hardcodes glyphs (`"╰─▶"`, `"├─▶"`) instead of the tuiwords table; `salience_test.go:199` scrapes the const block by regex (a `EventFoo // reason` form would be missed) where go/ast would be exact.

PR #351, opened 2026-09-02
lens: one shaper for both transcripts, and a room steer is an elbow

santoshkumarradha:
Lane L3 of #252 — the shaper and the elbows (owner rulings 2 and 4, Decision 2 and the
"shaper" third of Decision 1).

## What changed

**One record, one reading.** `room.go` held a second reader of the session file — its own
line struct, its own scanner, its own list of which roles exist, and hand-copied duplicates
of two display caps `internal/session` does not export. It was the copy that fell behind:
it knew `user` and `assistant` and nothing else, so a marked line, a session's own line and
a divider all reopened as ordinary questions. The reading is the engine's now
(`session.ReadTranscript` / `ReadTranscriptBytes`), and `replayBlocks` is the only
entry-shaping path on the surface, parameterised by two named literals — `chatReplay` and
`roomReplay`.

**A room steer is an elbow.** `└ ` where it was said, no turn opened, and on replay a
correction that reads as a correction. Every task steer confirms delivery with a short
fading clause — `session.SteerDelivered`, the engine's own sentence, drawn verbatim.

**The record keeps it.** A line delivered into a node is journaled with a steer mark: the
instant the person sent it, delivered, and that landing sentence.

Deleted from `room.go` (281 lines): `readRoomJournal`, `readRoomJournalTail`,
`readRoomJournalBytes`, `shapeRoomJournal`, `readJournalLines`, `scanJournalLines`,
`journalLine`, `journalCall`, `journalPart`, `journalPartImage`, `journalUserText`,
`journalArgs`, `journalOutput`, `journalArgsLimit`, `journalOutputLimit`; `clipBytes` and
`runeStart` moved to `export.go`, their one remaining caller.

## e2e

Two scenarios, both driving `bin/aforge` in tmux through `/task solo` → `ctrl+t` → the
node's page → a correction typed into a step still in flight.

`go test -tags e2e -count=1 -run TestARoomSteerIsAnElbowAndTheRecordKeepsIt -v ./internal/e2e/`
— scripted endpoint standing in for every model call, PASS 33.9s. Live pane:

```
─ ⠋ main ▸ PROBE-STEER write a · working · 2s · stub/scripted ──────────── esc/← main · ✕ ─
› WHAT THE PERSON ASKED FOR, IN THEIR OWN WORDS
  ▸ …6 more lines · ctrl+o
  ╰─▶ bash sleep 25 # step 2
└ the config lives under etc/ · delivered
```

the node's own record, read back through the new door:
`at=2026-09-02T14:30:58Z consumed=true landing="delivered"`, and the page reopened:

```
─ ? main ▸ PROBE-STEER write a · needs your look · 25s · stub/scripted ──── esc/← main ─
› WHAT THE PERSON ASKED FOR, IN THEIR OWN WORDS
  ╰─▶ bash sleep 25 # step 2
└ the config lives under etc/
Understood — reading etc/ instead. Nothing else to do here.
```

**The approval run — the shipped default.**
`go test -tags e2e -count=1 -run TestARoomSteerIsAnElbowAgainstTheShippedModel -timeout 20m -v ./internal/e2e/`
— every seat pinned to `config.DefaultModel`, model `~deepseek/deepseek-v4-flash-latest`,
$0.0004, **PASS 83.4s**. Live pane:

```
─ ⠼ main ▸ run `sleep 120` · working · 21s · ~deepseek/deepseek-v4-flash-latest ── esc/← main · ✕ ─
› WHAT THE PERSON ASKED FOR, IN THEIR OWN WORDS
  This is the message this work came out of. Where anything below reads differently from
  it, their words are what was asked for.
  ▸ …7 more lines · ctrl+o

  `bash` blocking for 120s now to let the service settle, then writing the file.

  ╰─▶ bash sleep 120

└ the config lives under etc/ · delivered
```

the node's own record, read back through the new door:
`at=2026-09-02T14:37:04Z consumed=true landing="delivered"`, and the page reopened:

```
─ ⠋ main ▸ run `sleep 120` · working · 51s · ~deepseek/deepseek-v4-flash-latest ── esc/← main · ✕ ─
› WHAT THE PERSON ASKED FOR, IN THEIR OWN WORDS
  ▸ …7 more lines · ctrl+o
  `bash` blocking for 120s now to let the service settle, then writing the file.
  ╰─▶ bash sleep 120
└ the config lives under etc/
```

The turn-glyph count is asserted unchanged across the steer in the live run: one `›` on the
page — the instruction — before and after.

**Second data point.** The same scenario ran once on `deepseek/deepseek-v4-pro` ($0.05,
PASS 67.0s) before the seats were pinned: `newHome` copies the person's own profile and
this machine's crew pins a bigger row for work, which is the defect the pin closes rather
than a property of the product. Same three assertions, same three shapes —
`└ the config lives under etc/ · delivered` live,
`at=2026-09-02T14:32:31Z consumed=true landing="delivered"` in the record, and
`└ the config lives under etc/` with no clause on the reopened page.

## Tests

`go test ./internal/tui3/` ok 487.6s · `go test ./internal/session/` ok 114.8s ·
`go test ./internal/manual/` ok. `make check` red is exactly the nine names already in
`.github/known-red.txt`.

Refs #252


santoshkumarradha:
Review (santosh-0a, review-only, at 15696a05). **Two in-PR defects I'd fix before merge, one manual nit, one comment/behaviour mismatch.** The elbow, the steer mark and the seam are right.

**1. The reading door rebuilds every picture on every read.** `readJournal` calls `journalPart.contentPart()` for each image part (sessionfile.go:1509): `os.Stat` + `os.ReadFile` + sha256 + base64 per image, per call. `ReadTranscript` now runs on every room open (room.go:493), every hosted record load (room.go:596), and the run page's 250 ms poll (roomorch.go:1034 → `orchReadTranscript`). The old reader built chips from the path and never touched the file (dev room.go:864 `picturesHere`). On a **hosted** page the paths are the other machine's, so `placeholderBecause` (sessionfile.go:714) emits a text part and `messageContentText` (title.go:181) puts `[image /their/path — file changed or gone]` into the person's own line, which is false. Fix: a reading path that keeps the refs (`journal.imageRefs` already gives shapeEntries the chips) and never rebuilds content — e.g. `readJournal(reader, path, rebuild bool)` with `false` from the two `ReadTranscript*` doors, `true` from `replaySessionFile`.

**2. Compaction history is dropped, and the design's own law forbids it.** `transcriptFrom` shapes `replayed.messages` only (transcriptread.go:89); `replayed.earlier` — every real result above the marker — is discarded. The conversation gets it back through `EarlierHistory` (replay.go:105); a room has no equivalent, so a page reopened after a node compacted can no longer expand the pre-marker calls onto anything but stubs. The old reader ignored markers and drew every line (both copies, which was wrong the other way). "THE LENS MAY LOWER SALIENCE; IT MAY NOT DROP A FACT" — this drops one. Either return the earlier region from the door and let `replayBlocks` draw it ahead of the floor, or file it as a known gap in the PR body before merge. Your call which; it should not be silent.

**3. Manual is stale.** `internal/manual/chat/tasks.md:2136–2148` still shows `· it was waiting on its pieces — your line wakes it` as its own dim row. This PR moves that sentence onto the elbow as a fading clause (`└ … · it was waiting…`).

**4. Comment says partial, code says empty.** transcriptread.go's header promises "the scan stops at the first unreadable line and hands back everything above it", but `readJournal`'s `scanner.Err()` arm (sessionfile.go:1464–1465) returns a `replayedSession` with no `messages`. A journal with one line over the scanner buffer reads as a blank page. Pre-existing shape in the resume path; the new comment makes a promise it doesn't keep.

**Checked and fine:** the deleted `steer_test.go` assertion is replaced by `TestALineDeliveredToANodeIsMarkedInItsRecord` (reads the reopened file); `TestAnUnansweredCallReadsBackAsUnansweredWithItsIdentity` pins the seam (CallID, Args, Answered) and `replay.go:627` carries it onto the row as `toolRunning`; elbow clause draws last so the context mark stays with the words; receipt fades on `hudWarm` and is empty on replay (`landed` from the mark's `At`).

**Nits:** `userMessage.steered` is now `crossed != nil`; `internal/e2e/roomsteer_e2e_test.go` hardcodes `"└ "`/`"› "`; `keys.md`'s new topic sits inside an existing `##` section.

Rebase note: four room.go conflicts against #337, all mechanical (listed on #337).

Issue #286, opened 2026-09-01
call log: workspace unit tests write fake model rows into the launching profile's ledger — test traffic never lands in a person's ledger


santoshkumarradha:
## Problem

Same run: the workspace under test was the aforge repo itself, and running its unit tests (`go test ./...` inside the leaf) wrote **356 rows of fake model traffic into the real profile's call log** — `vendor/vision-model` priced at $4.25, `work/model`, `sim/model` — interleaved with the run's genuine rows in `AFORGE_HOME/logs/calls.jsonl`. Any later reader summing that file gets fiction; the fake $4.25 vision row alone dwarfs the run's real spend.

## Root cause shape

Tests construct clients that inherit the ambient `AFORGE_HOME` (this was seen before on 2026-09-01: tests writing `test/model` / `cheap/one` rows into the live calls.jsonl). The call-log writer trusts the environment; the tests trust the writer.

## The law

Test traffic never lands in a person's ledger. A process's call log belongs to the home that launched it; a test either gets its own temp home by construction or writes nowhere.

## The fix shape

- The test helpers that build clients set an isolated `AFORGE_HOME`/log sink by default (one seam — whatever constructor the tests share), so no test can forget.
- Defense: the call-log writer refuses obviously-synthetic vendor ids only if that is cheap and honest — the primary fix is isolation, not a blocklist.

## Acceptance (functional, required)

- A structural test: no test in the tree can produce a row in a call log outside its own TempDir (run a sentinel: set AFORGE_HOME to a canary dir, run the affected packages' tests, assert the canary log gains zero rows).
- The run-side repro: `go test ./internal/...` executed inside a leaf workspace adds zero rows to the launching profile's calls.jsonl.

PR #357, opened 2026-09-02
call log drops any end row carrying +Inf or NaN, silently

### LINKED ISSUE #334: call log drops any end row carrying +Inf or NaN, silently — every long lane-watched call ends in flight forever
## Problem (measured 2026-09-02, headless `aforge do`, dev @ 779ed018)

The model-call log drops any end row that carries a non-finite number, and it drops it silently. In practice that is **every lane-watched call that lived long enough for the wait controller to reach a verdict**: the controller's cost of acting is `+Inf` whenever no alternative lane is in hand or λ is zero (`hazard.cost`, internal/lane/control/hazard.go), the verdict copies that into `Act.Cost`, `Client.record` copies it into `CostS`, `encoding/json` refuses `+Inf`, and `log.write` returns on the marshal error with no trace (internal/calllog/calllog.go, the first `return` in `write`).

Measured on one run: four `compile` calls on `deepseek/deepseek-v4-pro` (two in the failing run, two in the retry) each have a `phase:start` row and **no end row**, although the lane ledger recorded their sightings (a 137 s generation, 5,393 tokens) and the run was billed for them. Across the whole 19 MB log there is not one end row for that model on a long call; the same model's 3–6 s leaf calls, which end under the controller's floor before any verdict, all have end rows. Calls with no lane watch (the planning model) are unaffected.

## Consequences

- The log understates spend and hides exactly the calls it exists to explain: the stalls. A person reading it sees the call "in flight forever", which is the wording the code itself uses for the case it believes it prevents.
- `AFORGE_CALL_LOG_BODIES` cannot capture the reply of a long call at all, so a compile that came back malformed cannot be autopsied (this is what blocked the replication of the empty-goal defect filed alongside this one).
- `calllog.Last` is updated before the write, so the headless waiting line keeps saying "last call … 1s ago" while the file says nothing ended.

## Replication (runnable, no machine paths)

Drop this test into `internal/calllog` and run it; it passes today, which is the defect:

```go
func TestAnInfiniteFieldDropsTheRow(t *testing.T) {
	path := filepath.Join(t.TempDir(), "calls.jsonl")
	t.Setenv(EnvVar, path)
	Append(Record{ID: "start", Phase: PhaseStart, Model: "m"})
	Append(Record{ID: "end-inf", Model: "m", Status: 200, CostS: math.Inf(1)})
	Append(Record{ID: "end-ok", Model: "m", Status: 200, CostS: 1})
	data, _ := os.ReadFile(path)
	if !strings.Contains(string(data), "end-inf") {
		t.Fatal("the +Inf row was dropped silently") // fails today
	}
}
```

Live: run any headless task whose first call thinks for longer than the controller's floor on a lane-watched model, then grep the log for that call's id. One row, `phase:start`, nothing else.

## Acceptance

- A record with a non-finite float is written, with the non-finite fields omitted or clamped and the fact noted on the row; the test above passes.
- `write` never returns silently on a marshal failure: the failure is visible once per process (stderr or a synthetic row), because a log that silences itself is the one failure mode a log may not have.
- After a headless run whose compile ran past the floor, the compile's end row exists with status, tokens, finish and cost.
- The in-flight-forever wording in `client.go`'s stream epilogues becomes true.

## Not in scope

Why the compile on this crew thinks for ~100 s at effort low (noted in the empty-goal issue), and the test traffic in the same log (#286).

https://claude.ai/code/session_01QM1syvGCs7y5wRiXDD9xzy

santoshkumarradha:
Closes #334

## What was true

`(*log).write` marshalled the record and `return`ed on the error with no trace. `encoding/json` has no spelling for `+Inf`, `-Inf` or `NaN`, and the record it refused was almost always an END row: the wait controller prices acting at `+Inf` whenever it holds no alternative lane (`hazard.cost`), the verdict carries that into `Act.Cost`, the watch hands it to `Client.record` as `CostS`. So every lane-watched call the controller acted on had a start row and nothing beside it — read by a person, and by the manual's own words, as a call in flight forever. `calllog.Last` had already been updated, so the headless waiting line said "last call … 1s ago" over a file where nothing had ended.

## What is true now

`write` marshals through `marshalRecord` (new `internal/calllog/finite.go`, one call site). On a marshal error it takes every non-finite float off the record — `waste_usd`, `wait_s`, `cost_s`, `cost`, from one table — writes a sentence on the row saying which number it was and what it said (`cost_s was +Inf and is not on this row.`), and marshals again. The rest of the row is written whole.

**Taken off rather than written as a string or a null**, because the field's type is what every reader of this log decodes with, today and for every row already written: a `cost_s` that is sometimes a number and sometimes `"+Inf"` breaks the decode of a whole file for the sake of one line of it. A missing field already has a meaning here (the emptiness law — nothing was measured), and the note is what tells "infinite" apart from "unknown", in words, on the same row. The call's own note, when it has one, stays first.

A record that still cannot be written now reaches a once-per-process line on the package's existing `stderr` seam and **does not** silence the log — the rows that follow belong to other calls.

Per the ruling: the fix is in the writer, and `internal/lane/control` and `internal/provider/calllog.go` are untouched. The `+Inf` sentinel is legitimate where it lives, another non-finite value would find the same hole tomorrow, and a source-side zeroing would have erased the fact that the controller held no alternative lane — which the row now states.

## Evidence

Real-model e2e, `aforge do -model deepseek/deepseek-v4-pro` on a two-line file, `$0.0303`, `AFORGE_CALL_LOG` pointed at the run's own file. 16 rows, **8 starts / 8 ends, none unpaired**, and one of them is the issue's exact condition — a 39.6 s `compile` the controller hedged at the ceiling with cost `+Inf`:

```
{"ts":"2026-09-02T09:56:35.077-04:00","id":"b0699ba3","tag":"compile","model":"deepseek/deepseek-v4-pro",
 "served":"Baidu","status":200,"ms":39568,"finish":"stop","ttft_ms":1263,"lane":"DigitalOcean","hedged":true,
 "silence_ms":30000,"action":"hedge","arms":2,"wait_s":35.65458573121897,
 "note":"cost_s was +Inf and is not on this row.","prompt_tokens":4138,"completion_tokens":2718,
 "reasoning_tokens":2422,"cost":0.00993455684}
```

That is the row the issue measured as missing: status, tokens, finish and cost all present. A second run on `deepseek/deepseek-v4-flash` ($0.0153) was 7/7 paired but never ran past the ceiling, so it acted on nothing — the pro run is the one that reproduces.

Acceptance 3 needed no edit: both `client.go` epilogues say a row is written "without one the log shows the call in flight forever", which is now true rather than aspirational, and the manual's `**A call still running shows as ⋯ in flight.**` becomes true again. `aforge logs` does not render `note`, so no person-facing string moved and no manual page changed.

## Tests

`internal/calllog/finite_test.go` — the issue's own replication (verified red before the fix: *the +Inf row was dropped silently*), `-Inf` and `NaN` and `waste_usd`, the wording of the note and the facts around it surviving, several non-finite numbers named after the call's own sentence, and the once-per-process complaint that does not silence the log. That last one uses a new `marshalJSON` seam, because no `Record` can fail to marshal once its non-finite numbers are off it and a law a log may not break should not rest on unexercised code.

```
go vet ./internal/calllog/ ./internal/provider/ && go test ./internal/calllog/ ./internal/provider/ -count=1
ok  github.com/Agent-Field/aforge-v2/internal/calllog  0.065s
ok  github.com/Agent-Field/aforge-v2/internal/provider 53.031s
```

Full suite: red only on names already in `.github/known-red.txt` (harness-design ×4, guard ×2, thread, tui settings, plan terrain).

## Conflict note

#286 edits the `!l.resolved` branch of the same `write`. This change touches the first six lines of that function and nothing else, plus one now-unused import; the rebase either way is trivial.

https://claude.ai/code/session_01DLjRso4MaUPdvU8b2jP11t


santoshkumarradha:
Codex review: one finding (low) — the four-field table in `finite.go` duplicates `Record`'s float schema, so a float added later would silently reintroduce the dropped row. Addressed in d3d90bab with a structural test that reflects over `Record` and demands every `float64` on it appear in the table by its JSON name (verified red against a scratch fifth float field). No other findings.

Issue #266, opened 2026-09-01
routing: a 'provider.only permits only: X' refusal strikes nothing, survives every relax rung, and is drawn as 'slow' — one refusal object for strike, ladder and screen


santoshkumarradha:
## Problem

Measured chat run, 2026-09-01, dev 05b99537, deepseek-v4-flash pinned. Six HTTP 404s of the shape `…Providers serving <dated slug>: digitalocean, streamlake, baidu, deepinfra, but your request's provider.only preference permits only: coreweave` (also `nextbit`). CoreWeave was picked three separate times across the run. The refused arm's failure was drawn as `· slow · trying nextbit…` and the line only left the screen by aging out at the 10-minute window. The refusal class that means "this lane cannot serve this model" is the one class that can never act.

## Where it lives — three mirror-image defects and one mute surface

1. **The strike cannot fire.** `internal/provider/velocity.go` `refuseUpstream` (~:656) requires `refusal.FromUpstream()`; `internal/provider/endpoints.go` `routingRefusal` (~:248) classifies this 404 only when `upstream == ""`. Opposite preconditions: the router's own refusal is structurally exempt from striking the lane. So the lane is re-picked.
2. **The ladder cannot drop the pin.** `endpoints.go` `relaxationPlan` (~:336) offers rung 1 only for `RequireParameters`/`Ignore`/`MaxPrice`; `velocity.go` `relaxedPreferences` (~:385) clears those three and not `Only`, not `AllowFallbacks`. A pinned arm climbs six rungs still pinned to the lane that said no. (In the run the race walked instead — correctly — but the walk neither struck the lane nor updated any belief.)
3. **The frontier and the router disagree about who serves the model.** The frontier comes from the endpoints sheet (`internal/lane/sheet.go` Refresh → `frontier.capable`) fetched under the bare id; the completion resolves a dated slug; three of the five tool-capable lanes on the sheet were not in the router's serving set. Nothing writes refusals back. (The alias/identity fold itself is #215.)
[...]

Issue #264, opened 2026-09-01
lane ledger: a belief write on the send path holds a process-wide mutex across a blocking flock — one stuck lock silences the wire for the whole process


santoshkumarradha:
## Problem

During a measured chat run (2026-09-01, dev 05b99537, every role pinned to deepseek/deepseek-v4-flash), after two provider-refusal 404s at 17:09:20Z and 17:09:22Z the process issued **no model call at all for 29m49s** — zero `start` rows in the call log — while the surface said `waiting for deepseek-v4-flash · nothing has come back yet · slow`. Escape left the turn in `stopping` for minutes; only two ctrl+c ended the process. The tell: `v3/lanes.json` (the lane ledger) froze at 17:09:02Z — the second of the failure — while `v3/lanes/<model>.json` (the sheet cache, written *before* the lock is taken) kept advancing to 17:39:11Z. The network was fine; a lock was not.

## Where it lives

`internal/lane/belief.go` `ledger.Note` (~:448) takes the **process-wide ledger mutex** and, still holding it, calls `save()` → `store.saveMerging` → `store.locked` (`internal/lane/store.go` ~:194), which calls `filelock.Lock(gate, true, false)` — `flock(LOCK_EX)`, **blocking, no LOCK_NB, no deadline, no context** (`internal/filelock/filelock_unix.go` ~:14). The chooser reads the same mutex on every encode, so one stuck save freezes every model call in the process — chat and headless `do` alike. `Note` is called from the hedge's `settle` (`internal/provider/hedge.go` ~:1115), i.e. **on the send path**. Additionally `hedgeRace.run`'s select (`hedge.go` ~:701) has no `<-ctx.Done()` case, and `flock(2)` ignores context, which is why Escape could not reach it.

**Evidence gap, stated:** the lock *holder* in that run is not identified — plausibly a lock-order deadlock between the sheet beat's `primeFrom` and `settle`'s save inside one process. The fix below is right under either cause; a timeout alone would not be (a deadlock would churn instead of freeze).

## The law

The send path takes no lock that can wait on the world. A belief is advisory; the wire is not. A wait that cannot be cancelled may not be entered on the turn path.

## The fix

- Belief **writes leave the send path entirely**: `Note` enqueues; one background writer goroutine owns `save()` and the flock. The chooser reads a mutex-free snapshot (copy-on-write or RWLock with no I/O under it).
[...]

PR #358, opened 2026-09-02
stop: stopping is bounded at ten seconds and there is a second stage behind it

### LINKED ISSUE #265: stop: 'stopping' is unbounded by design and there is no second stage — a turn parked on an uncancellable wait cannot be ended from the keyboard (ruling wanted)
## Problem — ruling wanted

`stopping` is unbounded by design. `internal/tui3/app.go` `windingDown` (~:6786) documents the window as "a real window and it is not short", and the section below it ("WHY THERE IS NO SECOND STAGE") deliberately declines a hard stop. In the 2026-09-01 measured run, a turn parked on an uncancellable wait (the flock in #264) left Escape inert: `stopping` for 4 minutes, then only ctrl+c twice — killing the whole process, every conversation with it.

#264 removes that concrete instance. This issue is the standing question: **what does a person's stop mean when the turn cannot hear it?**

## The law proposed

Stop is stop. Cancellation reaches every wait on the turn path; a wait that cannot be cancelled may not be entered there (#264 enforces this for the lane ledger; the law generalizes it). Past a stated bound, the surface **detaches**: it stops representing the turn as its own, says so, and the turn is hard-ended — in-flight HTTP aborted, the goroutine abandoned, one journal line marking the turn `abandoned` with its last known spend — never silently orphaned while it keeps spending.

## The two designs to rule between

1. **Bounded second stage** (proposed): `stopping` has a deadline (e.g. 10s). At the deadline: abort the transport, mark the turn abandoned in the journal, free the surface. The tool batch's `wg.Wait()` is given a context.
2. **Status quo, stated louder**: keep the unbounded window but the surface must say what it is waiting on (which tool, which wait) so the person can judge — the current `stopping` says nothing.

## Acceptance (functional, required), if design 1 is taken

- e2e (tmux): a turn whose provider stream never ends (lane stub that writes forever) — Escape ends the surface's wait within the bound; the journal carries one `abandoned` line with spend; the next prompt is usable. 
- e2e: a turn parked on an external flock (as in #264's test) — same result even before #264's fix, proving the stop no longer depends on every wait being cancellable.
- The abandoned turn's cost still reaches the ledger (see the per-call-flush issue).

## Dependencies

After #264 (removes the known instance; this is the law). Related: the per-call spend flush issue — an abandoned turn is exactly a turn that never seals.

## Correction to the acceptance, 2026-09-02 (PR #358)

**The first e2e example above was wrong, and the test must not be "fixed" back to it.**

It asked the stream-that-never-ends case to end within the bound AND to write one
`abandoned` journal line. Those two cannot both be right. A provider stream is
cancellable — the request is built with `http.NewRequestWithContext` on the turn's
context — so Escape ends it in well under a second, the engine lets go properly, and
nothing is abandoned. The journal line means *this turn was left running past the
bound*; writing it for a turn that stopped cleanly would make the ledger record
something that did not happen.

So the acceptance is:

- **Stream that never ends** — the wait ends within the bound and **no** `abandoned`
  line is written. This is also the only proof that the deadline does *not* fire on
  turns that end tidily, which is the regression a bounded stop is most likely to cause.
- **A wait that genuinely cannot be cancelled** — the bound fires, the surface detaches,
  and exactly one `abandoned` line carries the turn's last known spend. Asserted on two
  such waits, one of them driven by a real model.

The ruling's substance is unchanged: past the bound, waits end, in-flight HTTP is
aborted, and one journal line records it.

santoshkumarradha:
Closes #265.

`stopping` was unbounded **by design**, and `internal/tui3/app.go` argued the case at length under its own heading `WHY THERE IS NO SECOND STAGE`. A turn parked on a wait no cancellation reaches left `esc` inert — four minutes in the measured run — and the only key left was `ctrl+c` twice, which takes the process and every other conversation in it. The ruling is design 1: **stop is stop**, past a stated bound the surface detaches.

### 1. `stopping` has a stated deadline — 10s, unchanged from the issue's proposal

`internal/tui3/app.go`'s `stopGrace`, armed in `interruptTurn` and **measured from the keypress**, not from the last event. I did not move the bound and did not need to: the waits that make an ordinary winding-down long are bash's three-second `WaitDelay` on a leaked output pipe and a `jobs` kill's two-second SIGTERM grace plus two more for the SIGKILL. Ten sits comfortably *above* the worst ordinary settle, so the deadline can only fire on a turn that was genuinely not going to let go — never on one that was three seconds from handing back what it did.

### 2. The bound is visible before it fires

```
stoppipe · bounded · running read · 2s   crew custom · $0.01 · 5.3k/200k · 3% · YOLO · stopping · detaching in 10s
```

`app.stoppingSegment` composes it and it counts down on the frame clock, beside the rewind arm and the quit arm, which run down there for the same reason. Where the agent has **no** door behind the bound the line says the bare `stopping` and nothing more — a capability that cannot work is absent, not broken, and a countdown drawn over nothing would be the surface promising an act it has no way to perform (`TestAStopWithNoDoorBehindItAnnouncesNoBound`).

### 3. At the deadline: the waits end, the HTTP aborts, the surface is freed, and ONE journal line

Built **bottom up**, in the order the old comment block itself named, because a deadline whose only act is to look away is exactly the failure the block warned about:

| rung | where | what changed |
| --- | --- | --- |
| a | `internal/session/jobs.go` | `waitDoneUnder` — the SIGTERM and SIGKILL graces take the call's context; `jobRegistry.kill` threads it from the `jobs` tool |
| b | `internal/exec/bare/tools.go` | a `ctx.Done()` arm on bash's wait that hands back what the command had accumulated and lets the reaper finish behind it |
| c | `internal/session/loop.go` + `abandon.go` | the tool batch's `wg.Wait()` gets `waitBatch`, cut by an **abandon signal** rather than by the turn's context |
| d | `internal/session/abandon.go` | `Agent.Abandon` — the door onto all of them, plus the second cancel, the disown, and the journal line |
| e | `internal/tui3/app.go` | the deadline, whose whole body is opening that door |

**(c) is keyed on the abandon signal and deliberately not on the turn's context.** An ordinary stop must still give every tool the seconds it needs to hand back what it actually did; cutting the batch at the keypress would throw that away on every stop. The signal is closed only once the person's stop has been given its whole settling window, and it rides down on the turn's context (`withAbandon`) because the waits that need it are several frames below anything holding a reference to the agent.

**The transport abort needed no new machinery, and that is deliberate.** Every provider request is already built with `http.NewRequestWithContext` on the turn's context (`internal/provider/client.go:1727`), so `Abandon`'s second cancel aborts the socket through the seam that already exists — the transport closes the connection and the reader parked inside `Read` is released. **This PR adds a test that proves the existing seam carries it rather than code that duplicates it** (`TestAbandonAbortsTheRequestThatIsStillInFlight`). Please do not go looking for abort code; it was not written on purpose. **No file in `internal/provider` is touched by this PR at all.**

**The journal line is the acceptance, not decoration.** `journalAbandoned` in `sessionfile.go`, written by `appendAbandoned`, carrying the reason and the turn's last known spend (`Agent.turnSpend`, moved by `bank` on the same call that moves the session's meter). A turn nobody waits for never writes the seal that carries its cost, so without this a detached turn is money spent with no record that anything was let go of. It is written **exactly once** — `Abandon` reports `false` for a session with no turn in flight, so a duplicated deadline writes no second line — and the replay drops it, for the reason it drops call lines: summing it would bill the turn twice.

### 4. The generalisation is verified with a wait that is NOT the flock

#264 removed the one concrete flock. A test that re-created it would prove only that #264 works. So both new uncancellable waits are of shapes nothing in this build has any special knowledge of:

* **unit** (`internal/session/abandon_test.go`): a tool spliced onto the belt that blocks on a channel nobody ever closes and ignores its context completely. The test asserts `Interrupt` alone changes nothing about it — if it ever does, the test fails loudly rather than passing vacuously.
* **e2e** (`internal/e2e/stopbound_e2e_test.go`): a **named pipe with no writer**, entered through `read`'s own `os.ReadFile`. No context exists anywhere on that path and none could help if it did. Run twice — once scripted, once with a real model choosing to make the call itself.

Nothing in the change is specific to the wait it was tested against: the abandon signal ends the batch wait whatever the tool happens to be doing, and the door does not know what it is interrupting. The wedge tool, the named pipe and the leaked bash pipe are three pieces of evidence, not three cases handled.

## e2e evidence (tmux, real binary, `bin/aforge`)

```
go test -tags e2e -run TestBoundedStopE2E -count=1 -timeout 12m -v ./internal/e2e/

--- PASS: TestBoundedStopE2E (19.97s)
    --- PASS: TestBoundedStopE2E/a_stream_that_never_ends (4.17s)
    --- PASS: TestBoundedStopE2E/a_wait_that_cannot_be_cancelled (15.80s)
```

**Pane after `esc`, the bound stated:**

```
  · esc interrupts · ctrl+c twice quits
› read the pipe please
  ▸ stopped by you at 2.2s · 1 tool call · ctrl+e
  · stopped
─ main ───────────────────────────────────────────────── esc again to rewind ─
 ›
stoppipe · bounded · running read · 2s  … · YOLO · stopping · detaching in 10s
```

**Pane 10s later, after the detach:**

```
  · esc interrupts · ctrl+c twice quits
› read the pipe please
  ▸ stopped by you at 12s · 1 tool call · ctrl+e
  · stopped
  · detached — the turn was let go of and nothing is waiting for it — it spent $0.01
                                          · 10:10 · 12s · 1 tool · $0.01 ·
─ main ─────────────────────────────────────────── / shows every command ─
 ›
stoppipe · bounded            crew custom · $0.01 · 5.4k/200k · 3% · YOLO · interrupted
```

**The journal line the run read back:**

```
{Reason:stop timeout Input:1200 Output:40 CacheRead:0 CacheWrite:0 CostUSD:0.0123 Calls:1}
```

The next prompt after the detach runs an ordinary turn; both scenarios assert it.

### The real-model replication — the issue's own report, no stub anywhere

```
go test -tags e2e -run 'TestBoundedStopE2E/a_real_model' -count=1 -timeout 12m -v ./internal/e2e/

--- PASS: TestBoundedStopE2E/a_real_model_on_an_uncancellable_wait (33.09s)
```

`deepseek/deepseek-v4-flash` on the wire, nothing scripted but the named pipe — which is the environment, not the harness. A person types `use the read tool on …/wedge.fifo and tell me what is in it`, the model decides for itself to call `read`, the call parks inside `os.ReadFile` on a pipe with no writer, and `esc` is pressed only once the status line has read `running read` for six unbroken seconds — proof the turn is genuinely inside the call and not between steps.

**After `esc`:**

```
stopreal · deepseek-v4-flash · running read · 6s   $0.0042 · 5.3k/1M · 1% · YOLO · stopping · detaching in 10s
```

**Ten seconds later:**

```
  ▸ stopped by you at 29s · thought 2.0s · 2 tool calls · ctrl+e
  · stopped
  · detached — the turn was let go of and nothing is waiting for it — it spent under a cent
                                          · 10:17 · 29s · 2 tools · $0.0042 ·
```

**The journal line:**

```
{Reason:stop timeout Input:19299 Output:107 CacheRead:0 CacheWrite:0 CostUSD:0.00415142 Calls:2}
```

Then `say the word ready and nothing else` runs an ordinary turn on the same session.

**The subtest skips on `config.APIKeyAt(config.ProfileDir())`, not on an environment variable** — deliberately, because on a machine whose key lives in `~/.aforge/config.json` an env-var guard would skip silently and clear this bar by never running, which is exactly how a project ends up believing it has real-model coverage it does not have.

### One correction to the issue's literal acceptance, agreed and recorded

See **"Correction to the acceptance, 2026-09-02 (PR #358)"** on #265, which records this and why. The issue's original text asks for an `abandoned` line in the **stream that never ends** case too. It does not get one, and that is the *right* answer rather than a weaker one: a provider stream **is** cancellable, because the request is built on the turn's context — so `esc` ends it in well under a second, the engine lets go properly, and a line claiming the turn had been abandoned would be a record of something that did not happen. That subtest therefore asserts the surface's wait ends **inside** the bound *and* that **no** `abandoned` line was written — which is also the only way to demonstrate the deadline does not fire on turns that end tidily. The bound firing, and the line, are asserted on the wait that genuinely cannot be cancelled.

## The new lock, and the race evidence for it

`internal/exec/bare`'s `outputAccumulator.finish`, `snapshot` and `closeTempFile` were unlocked, and that was safe **only** while nothing could reach them before `cmd.Wait` returned. The cancellation arm in (b) is precisely a new way to reach them beside a live pipe writer, so the three are now one door, `sealed()`, under one hold of the accumulator's own mutex, unlocked from a `defer`. The reason is written in the file, not only here.

```
go test -race ./internal/exec/bare/ -count=1
ok  	github.com/Agent-Field/aforge-v2/internal/exec/bare	7.098s

go test -race ./internal/exec/bare/ -run TestACancelledBashCall -count=1 -v
--- PASS: TestACancelledBashCallReturnsBeforeTheLeakedPipeIsClosed (0.70s)
```

`TestACancelledBashCallReturnsBeforeTheLeakedPipeIsClosed` builds the exact grandchild the old comment described — a `setsid` child holding the shell's output pipe after the group is killed — and asserts the cancelled call comes back in under a second with the output it had, rather than after `WaitDelay`'s three seconds. Under `-race` it is also the proof that returning while the writer is live is safe.

## A defect this change introduced, found and closed before review

Giving the tool batch's `wg.Wait()` an escape has a cost that has to be paid or the escape is worse than the wait it replaces. The loop under an abandoned turn now reaches its step boundary while its tools are still running — and **that boundary records**, one tool message per call, into `a.messages`, which by then belongs to whatever turn the person started after they stopped this one. Under the old unbounded code the goroutine simply stayed blocked forever and never got that far, so the escape was **strictly worse** than the unbounded wait until this was closed: one conversation growing another's tool rows.

Closed at the seam that produced it — a non-blocking read of the abandon signal immediately after `runToolsWarm`, and the turn returns having written nothing. `TestATurnThatWasLeftBehindNeverWritesIntoTheConversationAgain` pins it, and **the test has been watched failing**: with the guard disabled it reports the abandoned turn's tool result sitting in the live conversation's transcript.

Two more found the same way, by reading the diff rather than by the suite:

* **The sweep's return value was being dropped**, so a turn detached at the bound settled without the repository probe and the fade ticks every other finished turn is owed. Nothing in the suite would have said — every other test here asks whether the surface was *freed*, and a discarded return frees it perfectly. Now batched into the frame, and pinned by `TestADetachedTurnIsOwedTheSameSettleAsAnyOtherFinishedTurn`, which has also been watched failing with the bug put back while `TestAtTheBound…` stayed green beside it.
* **A deadline landing on a turn the engine had already let go of** would have dropped a stream that was closing cleanly and written a note claiming a detach that did not happen. It now stands down. `TestADeadlineOnATurnThatAlreadyEndedDetachesNothing`.

## An abandoned turn is not a silently orphaned one

`Abandon` **disowns** the turn under the lock: `turnSeq` is bumped, `running` is cleared and the turn's channels are let go of, so the next keystroke opens an ordinary turn. The abandoned goroutine — which may genuinely never return, that being the whole premise — captured the old number, and its own cleanup does nothing at all when the two no longer match. Without that guard a turn returning later would clear a live turn's `cancel`, close a live turn's hub, and close an already-closed `done` channel, which is a panic. A tool still running in the left-behind batch is recorded as unfinished rather than keeping its seeded panic sentence: nothing panicked, the person stopped the turn.

## The old comment block

Deleted honestly and **replaced with the true account**, not merely removed: the first half of its argument (there is no key left) still stands in full and is why the second stage is a **clock**; the second half (there would be nothing behind it) is why the door was built bottom up before the clock was wired to anything. `render.go`'s `stoppingWord` doc block is rewritten for the same reason; the constant itself is unchanged.

## Manual

`internal/manual/chat/keys.md` — "Why is the turn still finishing after esc" now states the bound, what happens at ten seconds, the exact note, and the `abandoned` journal line. The old paragraph `No key makes it stop harder, and there is no second stage` was false and is replaced.

## Tests

* `internal/session/abandon_test.go` — the uncancellable wait, the session free for the next prompt, exactly one journal line with spend (plus idempotence), the cost still on the meter, the in-flight request aborted, the left-behind batch's slots
* `internal/tui3/stopbound_test.go` — the bound announced and counting down, no bound where there is no door, nothing detached inside the window, the detach firing once and freeing the surface, the money named (and not named when there is none), the next prompt, the frame clock's lifetime, the settle a detached turn is owed, and a deadline landing on a turn that had already ended detaching nothing
* `internal/exec/bare/cancelwait_test.go` — the cancel arm and the race
* `internal/e2e/stopbound_e2e_test.go` — all three tmux cases, one of them a real model; needles added to `tuiwords_test.go` so the untagged gate keeps them honest

Suites, rebased on current `origin/dev` (`8d590ccd`):

```
ok  	internal/tui3       497s
ok  	internal/provider    50s
ok  	internal/exec        22s
ok  	internal/exec/bare    7s
ok  	internal/manual
ok  	internal/e2e              (untagged: the needle gate)
internal/session — one failure, and it is NOT this branch's:
  --- FAIL: TestNoRoadInTheTaskEngineHasMoreEndingsThanItsLedgerRow
      Agent.askTask (task.go) holds 18 decisions and the ceiling is 15.
```

`internal/guard`'s two structural tests are red as usual and are in `known-red.txt`; the goroutine they name in `internal/exec/bare/tools.go` is the pre-existing `cmd.Wait` spawn, not one of mine (this branch adds no goroutine to that file), and the new `sealed()` lock does not appear in the lock-defer list — it unlocks from a `defer`.

That complexity gate is **red on a clean checkout of `origin/dev` itself** and has been reported separately — reproduce with `git worktree add --detach /tmp/af-devcheck origin/dev && cd /tmp/af-devcheck && go test ./internal/session/ -run TestNoRoadInTheTaskEngineHasMoreEndingsThanItsLedgerRow -count=1`. This branch touches neither `internal/session/task.go` nor `complexity_test.go`, and the same suite was green at its original base `779ed018`. All seven new `internal/session` tests here pass.

## Coordination

* Rebased onto current `origin/dev` (`35c1a79e`, the #279 revert as merged, tagged `checkpoint/2026-09-02/dev-green-after-revert`) — **clean, 10/10 commits, no conflicts anywhere**, and no behaviour resolved silently.
  * A note for whoever checks that tag next: it is **annotated**, so a bare `git rev-parse checkpoint/2026-09-02/dev-green-after-revert` prints the tag object `87e77fa4` and looks like it disagrees with `origin/dev`. It does not — `git rev-parse <tag>^{commit}` is `35c1a79e`, which is `origin/dev` exactly. The mismatch is a reading error, not a bad tag.
* Touches **no** file in `internal/provider`, so it cannot meet #266's `refuseUpstream` call-line change, nor the fresh `streamguard.go` / `thinking.go` work on dev, nor the open off-machine PR **#308** against `client.go`. Noted here because that is where a reader would expect the abort to live.
* `render.go` is touched in two places only, both cleared beforehand: the `windingDown` branch in `stateWord`, and the `stoppingWord` doc block. Nothing near the deck struct or `conversation()`.
* No arm added to `apply()`'s event switch.

---

## Evidence re-derived on the merge tip (fd5b095c on dev 3967244b)

The rebase moved figures, so the evidence below replaces the earlier run rather than
being edited to match it. Both are shown.

- **Full suite**, verbatim `ci-full.yml` command (`go test -p 1 -timeout 15m -skip <known-red> ./...`):
  `EXIT=0`, 85 packages ok, **zero failures** — including `internal/session`, which is
  quiet for the first time since #400 landed the checkpoint-fixture fix.
- **Real-model e2e**: `go test -tags e2e -run TestBoundedStopE2E -count=1 -v ./internal/e2e/`
  → `--- PASS (43.81s)`, all three subtests (`a_stream_that_never_ends`,
  `a_wait_that_cannot_be_cancelled`, `a_real_model_on_an_uncancellable_wait`).
  With deepseek-v4-flash on the wire and nothing scripted, a turn parked inside `read` on a
  named pipe with no writer showed `stopping · detaching in 10s` immediately after esc,
  detached at the bound with `detached — the turn was let go of and nothing is waiting for
  it`, wrote exactly one journal line, and took the next prompt as an ordinary turn.
  - journal line, this run: `{Reason:stop timeout Input:19164 Output:218 CacheRead:256 CostUSD:0.003382468 Calls:2}`
  - journal line, pre-rebase run (superseded): `{Reason:stop timeout Input:19299 Output:107 CostUSD:0.00415142 Calls:2}`
  - the token and cost figures differ because a real model answered twice; the assertions
    — bound honoured, one line, spend carried, next prompt usable — are unchanged.


santoshkumarradha:
Review (santosh-1a, review-only, at the current head; codex verification + my own reads and test runs). **No blocker.** Three in-PR corrections, all in prose, none in behaviour.

**Engine-first order — holds.** `waitDoneUnder` selects on the call's context (jobs.go:763, :1130); bash's wait has the `ctx.Done()` arm and seals the accumulator under one lock (bare/tools.go:455, :499, `sealed()` unlocks from a defer at :990); `waitBatch` selects on the abandon channel and not `ctx.Done()` (abandon.go:183), so an ordinary esc still lets the batch settle (loop.go:1872); `Agent.Abandon` closes the signal, re-cancels the transport, closes the hub, writes one journal line (abandon.go:138, :159), disowns under the lock with the sequence bump and the stale goroutine checks the captured sequence (abandon.go:103, agent.go:1241); the tui3 deadline's only engine act is `door.Abandon` (app.go:6725). Idempotent via lock-guarded `running`.

**No key bound — holds.** esc still walks recall → rewind → interrupt (input.go:604, untouched); the second stage runs off the frame clock (app.go:3729); none of input.go / keys.go / keys_test.go / phase.go is in the diff; keys.md:334 says so.

**Detached-turn settle test — holds for the probe, overstated for the ticks.** `TestADetachedTurnIsOwedTheSameSettleAsAnyOtherFinishedTurn` (stopbound_test.go:241) fails if `stopSweep`'s return is dropped, via the repository probe. Production batches the whole `settle()` (app.go:3735, :4517), so the fade ticks ride along, but the test observes only the probe while its comment says both are pinned. `TestATurnThatWasLeftBehindNeverWritesIntoTheConversationAgain` pins the non-blocking read after `runToolsWarm` (loop.go:851/:865, abandon_test.go:200); `TestADeadlineOnATurnThatAlreadyEndedDetachesNothing` forces Abandon false, keeps the stream, writes no note (stopbound_test.go:281). All three pass here; `go test -race -run TestACancelledBashCall ./internal/exec/bare/` passes.

**The rewritten comment block — the "SECOND STAGE" block is true; two blocks beside it are stale.** The rungs it lists, the clock-not-key argument, the journal line and what a person reads all match the code. But `stopGrace`'s justification (app.go:6583) and `windingDown`'s (app.go:6603) argue the ten seconds from "bash's three-second `WaitDelay` and a `jobs` kill's two-plus-two" as the waits an ordinary winding-down still pays. This PR removed those from the esc path: `Interrupt` cancels the turn context (agent.go:1448), and after that the bash arm returns at once with what it had (your own `TestACancelledBashCallReturnsBeforeTheLeakedPipeIsClosed` asserts under a second) and both `waitDoneUnder` graces select the cancelled context immediately (jobs.go:776, :1133). The bound now guards only waits that ignore the context — the wedge tool, the named pipe — which is a better argument for ten seconds than the stale arithmetic. Reword both blocks to say what the ten seconds is actually above.

**Manual, keys.md:339.** "the background waits are ended" overpromises: the batch's *wait* ends; a wait that ignores its context keeps running behind the detached turn, as loop.go:1881 states. Say aforge stops waiting, cancellable work is aborted, and a wait that cannot be stopped may finish behind it.

**Also fine:** changelog present with `pr: 358` and the visible countdown/detach; no machinery vocabulary in added person-facing strings; `Agent.Abandon` is 67 lines but flat.

Trial merges: clean against #368, #361 and each of #337/#351/#355.

santoshkumarradha:
Caveat on this PR's gating run, recorded after the fact because it turned out to matter.

In the whole-tree run I merged on, `internal/lane` came back from the go test cache rather than being freshly executed:

```
ok  	github.com/Agent-Field/aforge-v2/internal/lane	(cached)
```

(The `internal/lane 0.393s` line in that same output was `make test-laws`, which is only the law tests.)

That package fails on this box on **clean `origin/dev`**, with or without this branch — `internal/lane/sheet.go`'s `cachePathIn` falls back to `home.Join("v3","lanes")`, so the tests read `~/.aforge/v3/lanes/<model>.json`, and this box has one (written 2026-09-02 16:01:51 by a run against deepseek-v4-flash). Proven both ways on a clean dev worktree:

```
AFORGE_HOME=/tmp/emptyhome  go test ./internal/lane -run TestAnEmptyLedgerIsAnEmptyChoice -count=1  ->  ok
(real home)                 go test ./internal/lane -run TestAnEmptyLedgerIsAnEmptyChoice -count=1  ->  FAIL
```

So this PR merged on a whole-tree run that did not re-execute `internal/lane`, and that run would have failed if it had. Nothing here touches `internal/lane`, and the `touched packages` CI job passed on a clean box, so I do not believe dev is damaged by this merge — but the green was weaker than it read, and a gate that accepts `(cached)` hides exactly this class of environment-dependent failure. Tracked as #475.

Recorded as a comment rather than by amending `checkpoint/2026-09-02/pr358-bounded-stop`: that tag is already pushed, and rewriting a published ref to add a note would change a checkpoint other sessions may have fetched.

Issue #349, opened 2026-09-02
headless: a failed run keeps its store, and every headless call names its tag and node


santoshkumarradha:
Parent: #343 (epic #338)

Sub-issue of the defaults child.

## What a debugger hits today

`aforge do` runs in a private store deleted on the way out whether the run succeeded or failed; the only way to keep it is to have passed `--keep` before knowing it would fail. The compile, the plan passes, the gate, and every `aforge exec` call leave call-log rows with no node and often no tag, so the record that does survive cannot be read one piece of work at a time.

## The law

A failed headless run keeps its record, and every headless model call names the work it belonged to.

## Fix shape

Keep the private store on any non-success exit and print where it is; delete on success unless kept on purpose; always keep under the debug switch. Set the tag and node on the compile, each plan pass, the gate and the `exec` executor where the node is already known.

## Replication and acceptance

- A headless run that fails leaves its store on disk and its location on the last line; a run that succeeds leaves none unless `--keep` was given.
- After any headless run, every call-log row it wrote carries a tag, and every row for work under a node carries that node.
- A run killed by a signal leaves its store (a signal is a failure the person did not choose), and when the process can still speak on the way out (a termination signal, not a kill) the last line names the folder.
- The private store lives under the state root, never in the operating system's temporary directory, so a kept record survives a temp-file sweep.

Issue #341, opened 2026-09-02
debug record: every tool call leaves its executed arguments, result, duration, status and refusal with reason


santoshkumarradha:
Parent epic: #338

Part of the debug-record epic. Depends on the foundation child.

## What a debugger hits today

A chat journals the model's raw tool arguments and the tool's result as a plain tool message. The duration exists only on a live screen event, the error bit only on another, and a tool the approval gate, the consent step or the guardian refused is written as a tool message that reads exactly like a tool that ran and returned that text. The arguments actually executed after the pre-action rewrite are not the ones journaled. Headless runs keep a trace snipped to a few hundred bytes with no timing and no status.

## The law

Every tool call leaves one event with its name, the arguments as executed, the result, the duration, the status, and, if it did not run, who refused it and why; and the event names the model call that asked for it.

## Fix shape

One recorder call at the single place a tool is executed in the session engine and one in the headless executor: begin (name, executed arguments, call id of the requesting model call), end (result, duration, status). A refusal from the approval gate, the consent step, the guardian, or a budget stop is the same event with a refusal field carrying the refuser's name and the reason in the words the person saw. Under the switch the result is whole; off, the transcript keeps what it keeps today. This also gives the gate, consent and guardian decisions their first line in the durable record.

## Replication

```sh
aforge chat --once "run: sleep 2 && echo done" --yolo
```
Today: the transcript shows a tool message with `done` and nothing says the tool took two seconds or exited zero. Deny the same command through the approval gate: the transcript shows a tool message carrying the refusal sentence and nothing marks it as refused.

## Acceptance (functional)

- After the command above with the switch on, the trace's events hold one tool event with name, the command as executed, the output, a duration of about two seconds, and a status of success.
- Refuse the same tool (approval mode set to ask, answer no): the event carries refused, the refuser's name, and the reason; no tool ran.
- A tool that fails (a command exiting non-zero, a read of a missing file) records the failure status and the error text.
- The event's call id names the model call whose reply requested the tool.
- A headless run's tool calls appear the same way in its trace.

santoshkumarradha:
[...]

Issue #340, opened 2026-09-02
debug record: a streamed call's reply, reasoning and finish are in the trace, and a cut keeps its partial


santoshkumarradha:
Parent epic: #338

Part of the debug-record epic. Depends on the foundation child (the recorder and the trace folder).

## What a debugger hits today

The bodies switch never records the reply of a streamed call, and every call is streamed. Three real runs, 26 attempts: every row carried the request body, none carried a response body. The streamed epilogues hand the record everything but the body; only the whole-body path passes one. Reasoning text is never captured on a streamed call at all. #335 is this gap: a compile came back with a blank goal, the run ended, and a rerun with bodies on still produced no reply text.

## The law

When the record is on, every model call's reply is on disk whole: the text, the tool calls, the reasoning if the model returned one, the finish reason, and, on a failure, the error body.

## Fix shape

The streamed path assembles the reply as it arrives already; at each of its ends (clean finish, cut, decode error, in-band refusal, cancel) it hands the recorder what it has, so a partial reply on a cut is recorded as partial rather than lost. The body file is written once per call under the call id in the trace folder, as one JSON document: request, response as assembled (text, tool calls, reasoning, finish), and the error if any. The reasoning is recorded as text, not as a list of deltas.

## Replication

```sh
AFORGE_DEBUG=1 aforge chat --once "use a tool to read notes.txt, then tell me its second line" --yolo
```
Today (with the old bodies pin): no row for the run carries a response body.

## Acceptance (functional)

- After the command above, the trace folder holds one body file per model call, and each holds the reply text the person saw on screen, the tool call the model made, the reasoning text where the model returned one, and the finish reason.
- Force a cut (a model that streams and a stall guard set low, or a provider answering with a refusal body): the body file holds the partial text received up to the cut and the reason word, not an empty response.
- A whole-body call (a headless plan pass) produces the same document shape as a streamed one.
- With the switch off, no body file is written and the call log is unchanged.

santoshkumarradha:
Consolidated into the epic #338. This issue's law, fix shape, replication and acceptance are reproduced in full in the epic's consolidated plan (comment on #338, 2026-09-10), under the pull request that now carries it — nothing is lost, and this issue stays readable.

[...]

Issue #339, opened 2026-09-02
debug record: one switch, one recorder, a run id and a call id on every record, and a trace folder that is a person's own data


santoshkumarradha:
Parent epic: #338

Part of the debug-record epic (parent linked below). Lands **after #286 and #334**, because it stamps ids onto the call-log row that those two are correcting.

## What a debugger hits today

There is no switch. Two environment pins exist, buried in the settings registry, with no flag, no slash command and no page a person would find first. With the bodies pin on, a 65 KB request is written twice per attempt, on the start row and again on the end row, into a call log that rotates at 32 MB, so one long turn evicts the failure the person came for. And nothing on disk joins anything: the call-log row has no session or run id, the transcript's call line has no call id, the lane journal has no request identity.

## The law

One switch turns on the full record of a run without changing the run; every record it writes carries the run id and, for anything about a model call, the call id; the trace is a person's own data.

## Fix shape

- **The switch, three doors, one meaning**: an environment pin `AFORGE_DEBUG=1`, a `--debug` flag on `chat`, `do` and `exec`, and `/debug` inside a chat (turns the record on for the rest of the session). Behaviour of the run is unchanged. The old bodies pin keeps working for one release and means the same thing.
- **One recorder package**, neutral name, one small interface: record a call body (request, response), a tool event, a decision with a reason. Nil-safe: when the switch is off every method is a no-op behind one nil check, and no feeder site pays more than that. No second logging library, no per-package levels.
- **Two identities**: a run id minted once per chat session and per headless invocation and carried on the context; a call id per model call, which the call log already mints per attempt. The run id is stamped on every call-log row, the transcript header, and lane journal lines; the call id on the call-log row, the transcript's call line, and the trace body file's name.
- **The trace directory**: one folder per run under the state root's `logs/trace/`, holding `events.jsonl` and one file per call body named by the call id. Files are private to the user. No authorization header or key is ever written. Bodies stop being written into the call log on or off; the request body is captured once per attempt.
- **Size law**: a cap per run and a retention count over runs; when the cap is reached the run's trace says so on its last line and stops; when the count is exceeded the oldest run is pruned whole. Never a rotation that evicts a failure mid-run.
- A manual page under the chat corpus that a person finds by asking "how do I see what happened" or "debug log", naming the switch and where the record is.

## Replication

```sh
AFORGE_CALL_LOG_BODIES=1 aforge chat --once "read notes.txt and tell me its second line" --yolo
aforge logs --tail 4
[...]

PR #361, opened 2026-09-02
trace: one switch, one recorder, and a folder per run that is a person's own data (#339)

santoshkumarradha:
Parent #339 (PR A of four), epic #338. Not a closing PR: #339 closes when PR D lands.

## What this PR is judged by

**PR A is judged by what it lays**: the run folder with `run.json` written at the door, the permissions, the exit line naming the folder, no folder at all when the switch is off, and the size law exercised — the cap line and the retention pruning shown. Reconstructing a run from the files on disk is the acceptance of #340 and #341 and will be shown there on real runs; **this PR does not claim it**, and no feeder is folded in.

## What lands: one switch, two scopes, both honest

**The run id on the context is the scope of the switch.** `trace.Enable()` flips a
process-wide atomic while the record folder is per run, so `/debug` typed in one
conversation turned the record on for **every** run the process held — an engine host holds
one per person — and named only its own folder. Somebody else's prompts, files and replies
would land in a folder they never asked for.

- **`/debug` enables ONLY the run whose id is on the context.** It calls
  `trace.EnableRun(ctx)`, which marks that run id in a small mutex-guarded set;
  `For(ctx)` returns a recorder when the process switch is on **or** the context's run is
  in that set. No process-wide flip from `/debug`. The off path costs one extra atomic
  load (`anyRun`), so a process where nobody typed it still pays no mutex.
- **The pin and the flag stay process-wide**, because the person gave them to *this*
  process on purpose: every run it opens is recorded, each into its own folder.
- **The process-run fallback in `RunFrom` no longer carries one run's `/debug` into
  another.** A context with no run id of its own gets a recorder only under the process
  switch — which is what keeps `do`'s inner `Background()` contexts recording under
  `--debug` and `AFORGE_DEBUG`, exactly as they did.
- **`/debug` answers all three ways**: `recording this conversation · it goes to <folder>`,
  `the record is already on · it goes to <folder>`, and under the pin or the flag
  `the record is already on for every conversation this aforge holds · this one goes to <folder>`.

The manual says it in one sentence — *with the pin or the flag, every conversation this
process holds is recorded, each into its own folder; `/debug` records only the conversation
you typed it in* — plus a `/debug` row in the commands table, and the change entry carries
the same line.

## Also in this push — the review's three items and its smaller ones

- **Every door registers the credentials this profile is configured with.**
  `config.Credentials(profileDir)` walks the settings rows marked `Secret` — the model key
  (both environment spellings), the exa/firecrawl/jina keys, the Google app secret — and
  `openDebugRecord` hands each to `trace.Secret`. Scrubbing by shape only ever catches the
  shapes somebody thought of: a Google `AIza…`, a Groq `gsk_…` or a self-hosted endpoint's
  plain token echoed back in an error body would have landed verbatim. The registry's own
  `Secret` flag is the source, so a credential row added later is covered the day it is
  added (`TestCredentialsFindsEverySecretRowInTheRegistry`), and the changelog line now
  states what is true rather than promising shapes.
- **The schema has a node.** `trace.WithNode(ctx, node)` / `NodeFrom(ctx)` — this package's
  own context key, no import of provider — and a `Node string` on `CallBody`, `ToolEvent`
  and `Decision`. The recorder fills `node` from the context when the field is empty, and
  the field wins when a feeder names one. PR B makes `provider.WithCallNode` call it too,
  so one context value is set in one place and both records name the same work.
- **Run ids are eight bytes, not four.** The id names a folder; two runs that minted the
  same id had `MkdirAll` merge them silently. Sixteen hex characters, and the comment now
  says "unlikely to collide". The manual and the change entry moved with it.
- **`cap()` opens the events file if nothing else has**, so a run whose first record is an
  oversized call body no longer caps in silence into an empty folder
  (`TestAFirstOversizedCallBodySaysTheRunWasCapped`).
- **A call with no id is named by the millisecond *and* the run's own count**, so two
  anonymous calls inside one millisecond keep both bodies.
- **The nil-recorder test covers `Header`.**

## Second-eye review — six more, all inside `internal/trace`

- **A call id is not a path.** `writeFile` joined an unsanitised id into the run's folder,
  and a tool call carries the id the *model* wrote: `../../outside` wrote a file wherever
  that resolved to. Body file names are now `[A-Za-z0-9_-]`; anything else is stored under
  its hex form — lossless, collision-free, and it cannot hold a separator or a dot. The
  record inside still says the id the model wrote
  (`TestACallIdThatIsAPathCannotEscapeTheRunsFolder`, `TestSafeNameKeepsOrdinaryIds…`).
- **The fallback no longer guesses between two runs.** It named whichever run began *last*,
  so `Begin(A)`, `Begin(B)`, then a contextless record from A landed in **B's** folder. The
  law now: *the fallback names the first run this process began, and only while it is the
  only one.* A second `Begin` clears it and a contextless record gets no recorder.
  `chat --once`, `do` and `exec` are single-run processes and keep recording exactly as
  before — which is what `do`'s inner `Background()` contexts need until #349's follow-up
  threads the run through them. A record lost is recoverable by threading a context; a
  record misfiled is not. Both halves tested
  (`TestTheFallbackAnswersForOneRunAndGoesQuietForTwo`).
- **Retention cannot delete a live run.** Two conversations open with `KEEP=1` had the
  older one's record pruned out from under it, mid-turn. `prune` now runs holding `runs`
  and skips every run with a recorder open; they still count against the ceiling, they
  simply cannot be the ones removed (`TestPruningNeverDeletesARunThatIsStillGoing`).
- **The capped receipt fits under the ceiling.** Every write leaves `capReserve` (512 B)
  free for it and the receipt is counted, so a folder no longer exceeds `MaxRunBytes` by a
  line — the cap tests now assert the folder's total size against its own ceiling.
- **A capped line that cannot be written silences the run**, through the same door as any
  other write failure, instead of being swallowed
  (`TestACappedLineThatCannotBeWrittenSilencesTheRun`).
- **Modes are repaired on paths the record did not create.** `MkdirAll` and `O_CREATE` say
  nothing about a folder or file that already existed, so a precreated `0755` folder or
  `0644` file kept its mode while holding a person's prompts and replies. Both are
  `chmod`'d after open (`TestTheRecordRepairsTheModesOfPathsItDidNotCreate`).

Declined, with reasons: removing the fallback (`do`'s inner contexts need it until #349's
follow-up threads the run through them), and `Secret`'s eight-character floor stays — its
comment now says why, that a shorter registered value would redact ordinary text.

## The scope, end to end

**A real run, with the pin.** Isolated `AFORGE_HOME` holding a copy of `~/.aforge/config.json`,
`deepseek/deepseek-v4-flash`, the binary from `make build`:

```
$ AFORGE_HOME=$E AFORGE_DEBUG=1 bin/aforge chat --once "say hi" --model deepseek/deepseek-v4-flash
Hi
debug record: …/e2e-scope2/logs/trace/6b56f4917995a618

$ find $E/logs/trace
…/logs/trace
…/logs/trace/6b56f4917995a618
…/logs/trace/6b56f4917995a618/run.json

$ cat $E/logs/trace/6b56f4917995a618/run.json
{"build":"3a98bfe8 (dirty) built 2026-09-02 15:47","command":"chat","kind":"run","model":"deepseek/deepseek-v4-flash","run":"6b56f4917995a618","started":"2026-09-02T15:48:18.190-04:00","ts":"2026-09-02T15:48:18.190-04:00","workspace":"/home/santosh/af-339"}

$ grep -r "<the configured api_key>" $E/logs/trace   → nothing
$ grep -ri authorization $E/logs/trace               → nothing
```

The same run with the pin off leaves `logs/calls.jsonl` and **no `logs/trace` at all**.

**Two runs in one process, `/debug` on one — one folder.** The unit-level demonstration,
run against the real recorder with a temporary state root: two run ids, `EnableRun` on the
first only (what `/debug` does), then a header opened for **both**.

```
$ go test ./internal/trace/ -run TestScopeDemo -v
    /logs/trace
    /logs/trace/aaaa1111
    /logs/trace/aaaa1111/run.json
--- PASS
```

The second run wrote nothing: `For` returns nil for it while the process switch is off.
That is `TestDebugRecordsOnlyTheRunItWasTypedIn` and, at the surface,
`TestDebugRecordsOnlyTheConversationItWasTypedIn` (two apps in one process, the reply text
of all three answers).

> The transcripts below were captured before the run id widened to eight bytes; folder
> names are sixteen hex characters now, as in the run above. Nothing else in them moved.

## The real end-to-end run — two real model runs with the switch on, and the same two with it off

`AFORGE_HOME` pointed at an isolated directory holding a copy of `~/.aforge/config.json`, model `deepseek/deepseek-v4-flash`, working folder `…/scratchpad/wk`.

### Switch on

```
$ AFORGE_HOME=$ON AFORGE_DEBUG=1 bin/aforge chat --once "say hello in three words" --model deepseek/deepseek-v4-flash
Hello from aforge.
debug record: …/e2eon/logs/trace/ebafb454
exit=0

$ AFORGE_HOME=$ON bin/aforge do --debug --model deepseek/deepseek-v4-flash "write the single word ready into ready.txt in this folder"
files:
  …/scratchpad/wk/ready.txt
29s · 1 node · $0.0109
debug record: …/e2eon/logs/trace/673df9c9
```

Both runs behaved exactly as they do without the switch — the reply, the deliverable, the exit code, the call log.

```
$ ls -la $ON/logs/trace
drwx------ 4 santosh santosh 4096 Sep  2 10:19 .
drwx------ 3 santosh santosh 4096 Sep  2 10:19 ..
drwx------ 2 santosh santosh 4096 Sep  2 10:19 673df9c9
drwx------ 2 santosh santosh 4096 Sep  2 10:19 ebafb454

$ ls -la $ON/logs/trace/ebafb454
drwx------ 2 santosh santosh 4096 Sep  2 10:19 .
drwx------ 4 santosh santosh 4096 Sep  2 10:19 ..
-rw------- 1 santosh santosh  324 Sep  2 10:19 run.json

$ cat $ON/logs/trace/ebafb454/run.json
{"build":"54cda541 (dirty) built 2026-09-02 10:19","command":"chat","kind":"run","model":"deepseek/deepseek-v4-flash","run":"ebafb454","started":"2026-09-02T10:19:21.833-04:00","ts":"2026-09-02T10:19:21.833-04:00","workspace":"…/scratchpad/wk"}

$ ls -la $ON/logs/trace/673df9c9
drwx------ 2 santosh santosh 4096 Sep  2 10:19 .
drwx------ 4 santosh santosh 4096 Sep  2 10:19 ..
-rw------- 1 santosh santosh  322 Sep  2 10:19 run.json

$ cat $ON/logs/trace/673df9c9/run.json
{"build":"54cda541 (dirty) built 2026-09-02 10:19","command":"do","kind":"run","model":"deepseek/deepseek-v4-flash","run":"673df9c9","started":"2026-09-02T10:19:33.344-04:00","ts":"2026-09-02T10:19:33.344-04:00","workspace":"…/scratchpad/wk"}
```

`0700` on the trace root, `0700` on each run's folder, `0600` on `run.json`.

### Switch off — the same two runs, a second isolated home

```
$ AFORGE_HOME=$OFF bin/aforge chat --once "say hello in three words" --model deepseek/deepseek-v4-flash
Hello friend.
exit=0

$ AFORGE_HOME=$OFF bin/aforge do --model deepseek/deepseek-v4-flash "write the single word ready into ready.txt in this folder"
files:
  …/scratchpad/wk/ready.txt
18s · 1 node · $0.0034
```

No exit line, and nothing was created:

```
$ ls -la $OFF/logs/
-rw-r--r-- 1 santosh santosh 5847 Sep  2 10:20 calls.jsonl

$ ls -la $OFF/logs/trace
ls: cannot access '…/e2eoff/logs/trace': No such file or directory
```

### The size law, exercised for real

Three runs against one home with `AFORGE_TRACE_MAX_MB=1` and `AFORGE_TRACE_KEEP=2`, each writing far past the cap:

```
debug record: …/e2ecap/logs/trace/fce9437c
debug record: …/e2ecap/logs/trace/6262de92
debug record: …/e2ecap/logs/trace/2b0a817b

$ ls -la $CAP/logs/trace
drwx------ 4 santosh santosh 4096 Sep  2 10:22 .
drwx------ 3 santosh santosh 4096 Sep  2 10:22 ..
drwx------ 2 santosh santosh 4096 Sep  2 10:22 2b0a817b
drwx------ 2 santosh santosh 4096 Sep  2 10:22 6262de92
```

The third run pruned the oldest folder **whole** — `fce9437c` is gone, the two newer runs are complete. And each capped run ends by saying so and keeps what it had:

```
$ tail -n 1 $CAP/logs/trace/2b0a817b/events.jsonl
{"bytes":1048470,"kind":"capped","max":1048576,"run":"2b0a817b","ts":"2026-09-02T10:22:19.632-04:00"}
```

### Redaction, at the recorder

A call recorded with `"authorization":"Bearer sk-or-v1-notarealkey123"` in its request body and that same key registered with `trace.Secret` — which every door now does for real, for every configured credential (`TestTheDoorRegistersConfiguredCredentialsWithTheRecord` proves a value no regex would match is redacted):

```json
{"call":"9f2a71cd","finish":"stop","kind":"call","model":"deepseek/deepseek-v4-flash","request":{"model":"deepseek/deepseek-v4-flash","messages":[{"role":"user","content":"say hello in three words"}],"credential":"[redacted]"},"response":{"choices":[{"message":{"content":"Hello there!"},"finish_reason":"stop"}]},"run":"52dfbdde","ts":"2026-09-02T10:14:17.073-04:00"}
```

`grep -ri authorization` and `grep -r` for the key's own value over the folder both return nothing — the field's **name** goes with its value, because a redacted value under its own name still answers that grep.

## Acceptance from #339 that this PR meets

- **The switch, three doors, one meaning** — `AFORGE_DEBUG`, `--debug` on chat/do/exec, `/debug` in a conversation, and the behaviour of the run unchanged either way. ✔ (both real runs above)
- **All three doors leave a trace folder for that run; with none of them, a run that succeeds leaves no trace folder and the call log gains its rows as before.** ✔
- **The old bodies pin keeps working and means the same thing.** ✔
- **One recorder package, neutral name, one small interface, nil-safe, no second logging library and no per-package levels.** ✔
- **A run id minted once per chat session and per headless invocation, carried on the context.** ✔
- **The trace directory** under the state root's `logs/trace/`, one folder per run, files private to the user. ✔
- **`grep -i authorization` and the key's own value across the trace folder find nothing.** ✔
- **Trace files and folders are readable by the owner only.** ✔
- **With the per-run cap set low, a run that exceeds it ends its trace with a line saying the cap was reached and keeps what it had; with the retention count set to two, a third run removes the oldest folder whole.** ✔ (shown above, and `TestARunThatReachesTheCapSaysSoAndStops`, `TestTheOldestRunFolderIsPrunedWhole`)
- **The manual answers "how do I turn on the debug log" and "where is the debug record" with the switch and the folder.** ✔

Not claimed here, by the plan above: the run id on the call-log row and the lane journal, the call id on the transcript, the bodies leaving `calls.jsonl`, and reconstructing a run from disk — B, C, D, #340 and #341.

## Tests

Single packages only (a load rule was in force):

```
go vet ./internal/trace/ ./internal/config/ ./cmd/aforge/ ./internal/tui3/   clean
go test ./internal/trace/ -race            ok  1.094s   (33 tests)
go test ./internal/config/                 ok  2.489s
go test ./internal/manual/                 ok  0.777s   (29 tests)
go test ./internal/tui3/                   ok  485.7s   (whole package)
go test ./internal/tui3/ -run Debug|Manual ok  0.282s   (10 tests)
go test ./cmd/aforge/ -run Trace|Debug|Record  ok 1.227s (18 tests)
make build                                 ok           (repacks the manual)
```

The whole `./cmd/aforge/` package fails only `TestTickWalksTheItemsAndWritesAWakeLine` and
`TestTickLeavesQuietlyWhenAWindowIsAlreadyKeepingWatch`, both lines 48-49 of
`.github/known-red.txt`.

New tests — scope: `TestDebugRecordsOnlyTheRunItWasTypedIn`,
`TestTheProcessRunFallbackDoesNotCarryOneRunsDebug`,
`TestTheProcessSwitchRecordsEveryRunInItsOwnFolder`,
`TestARunSwitchedOnByItselfWritesAndAnnouncesItsFolder`,
`TestEnableRunOnAContextWithNoRunSwitchesNothingOn`,
`TestDebugRecordsOnlyTheConversationItWasTypedIn`,
`TestDebugSaysSoWhenThereIsNoRunToRecord`. Review items:
`TestEveryRecordNamesTheNodeOnItsContext`, `TestANamedNodeBeatsTheOneOnTheContext`,
`TestARecordWithNoNodeSaysNothingAboutOne`,
`TestAFirstOversizedCallBodySaysTheRunWasCapped`,
`TestTwoUnnamedCallsInTheSameMillisecondKeepBothBodies`,
`TestCredentialsFindsEverySecretRowInTheRegistry`,
`TestCredentialsReadsBothSpellingsOfTheModelKey`,
`TestTheDoorRegistersConfiguredCredentialsWithTheRecord`. Second eye:
`TestACallIdThatIsAPathCannotEscapeTheRunsFolder`,
`TestSafeNameKeepsOrdinaryIdsAndEncodesEverythingElse`,
`TestTheFallbackAnswersForOneRunAndGoesQuietForTwo`,
`TestPruningNeverDeletesARunThatIsStillGoing`,
`TestTheRecordRepairsTheModesOfPathsItDidNotCreate`,
`TestACappedLineThatCannotBeWrittenSilencesTheRun`.

A real run after all of it, isolated `AFORGE_HOME`, `deepseek/deepseek-v4-flash`:
`AFORGE_DEBUG=1 aforge chat --once "say hi"` → `debug record:
…/logs/trace/0649d90445361f98`, `700` on the folder, `600` on `run.json`.

https://claude.ai/code/session_01PCndzpMTngv6Vrv97ikrBJ


santoshkumarradha:
Review (santosh-1a, review-only; codex verification + my own reads; `go test ./internal/trace/`, manual tests, build and vet all green here). **Hold before merge, for three things that are cheap to fix.** The recorder itself is sound.

**1. The body and the close line describe a change that is not in this diff.** The title says "Closes #349", and the body's first half — "What a debugger hits today", "The law", "What changed" — is #349's work: `keepPrivateStore(asked, debugging, succeeded)`, `errandContext` tagging ten errands, `leafKey()` in `internal/exec/linear.go`, `task-1` for `aforge exec`, node keys in `plan/brief.go` and friends. None of that is here. `cmd/aforge/do.go`'s diff adds `--debug`, `openDebugRecord` and the announce; `defer os.RemoveAll(home)` is untouched (do.go:356) and the store is still minted under the OS temp dir (do.go:665); `exec.go:259` still sets `NodeID: 1` with no key. Merging as-is auto-closes #349 with nothing landed for it. Fix: cut the #349 section, drop "Closes #349", keep the PR-A section ("What this PR is judged by") as the body, and the "Not done, deliberately" list. The second half of the body is accurate about what a foundation lays.

**2. "no key in any shape is ever written" is a promise the scrub cannot keep.** `trace.Secret(value)` exists to register the live key (scrub.go:48) and is called from no product code — only trace_test.go:310/:335. So `Scrub` is the three regexes alone: credential-named JSON fields, `Bearer …`, and `sk-…`. A Google `AIza…`, Groq `gsk_…`, or a custom base URL's plain token echoed in a provider error body lands verbatim once a feeder writes bodies. Nothing writes bodies yet, which is why the body's `grep` acceptance passes trivially. Fix in this PR: `trace.Secret(config.APIKeyAt(...))` for every configured credential inside `openDebugRecord` (cmd/aforge/debugrecord.go) so the guarantee is true before a feeder exists, plus a test with a non-patterned key. Then the changelog line (361-debug-record-switch.md:9) stands; otherwise narrow it.

**3. The schema has no node.** `CallBody`, `ToolEvent` and `Decision` (recorder.go:23–70) carry `run` and an optional `call`; their `ctx` argument is ignored. The body's own law is "every headless model call names the work it belonged to", and the feeders (B, C, D) cannot add that without changing this foundation. Decide now: a `Node string` field, or read `calllog`'s `WithCallNode` value off the context inside the recorder. One line of design, so the feeders do not each invent it.

**Holds:** nil-safe on every exported method (recorder.go:103/:112/:124/:157/:186/:398); default off, one switch with three doors (trace.go:102, chatv3.go:97, do.go:191, exec.go:44, commands.go:1027), the old bodies pin aliasing it on purpose; off path creates no file or directory (trace.go:178/:218) — it does mint a 4-byte id, so "zero cost" is zero *disk* cost; dirs 0700 and files 0600 everywhere, no temp/rename path (recorder.go:278/:286/:322/:340, pinned at trace_test.go:188/:461); no headers, URL or cookies are captured at all — the schema is bodies and metadata only, so the auth-header question is moot until a feeder adds a field, and there is no structural test stopping that; cap checked before every write (recorder.go:249/:282), the `capped` line runs one line past the cap by design; retention prunes whole folders; the folder is under the state root via `home.Join` (trace.go:199), never deleted at exit; no frozen path touched; no machinery vocabulary in new person-facing strings.

**Smaller:** `cap()` writes nothing when `events` is nil (recorder.go:307), so a first oversized standalone file caps silently — open events.jsonl for the receipt, and test the call-body-first path. Run ids are 4 random bytes (trace.go:142) and `MkdirAll` merges a collision silently — `os.Mkdir` with retry, and the comment should say "unlikely to collide", not "will not". Missing call ids fall back to millisecond names and can overwrite (recorder.go:146). The "every method" nil test (trace_test.go:32) skips `Header`.

**Manual** (debug-record.md) documents `/debug`, the flag, the env pins, path, limits, retention and modes, and says no feeder is in yet — accurate. Trial merges: clean against #358, #368 and each of #337/#351/#355.

Issue #289, opened 2026-09-01
ledger identity: one model routed as three — the floating alias, the concrete slug and the dated served slug build separate beliefs


santoshkumarradha:
Split out of #215, which mixes two independent defects. The routing half of #215 is covered by #253's waiting-policy redesign; this half is not, and must not close with it.

## Problem (measured 2026-09-01, headless `aforge do`, shipped defaults, build a850bd89)

The shipped default is the floating alias `~deepseek/deepseek-v4-flash-latest`. On the wire it is served as `deepseek/deepseek-v4-flash-0731`. The lane ledger keys by the **served** slug, so one model built beliefs under three identities in a single run:

| identity | facts held |
|---|---|
| `deepseek/deepseek-v4-flash` | full catalog sheet — Cloudflare/CoreWeave `Tools:true`, prices, context |
| `deepseek/deepseek-v4-flash-0731` | none — `Tools:false, Context:0, PriceIn:0, Status:0` |
| `~deepseek/deepseek-v4-flash-latest` | none, OpenInference only |

## Consequences, both measured in the same run

1. **The shipped default routes with zero priors.** It could not see that Relace serves this model at p50 1.8s while OpenInference takes p50 44s, and chose OpenInference 13 times. One `fanout` call sat there 240s until the transport deadline, on a lane whose own belief was `Rate.X ≈ −0.01` (about 1 tok/s).
2. **The hedge could not be solved.** `verdict()` returns nothing when `w.alt == ""`, and `Choice.Alt` was empty precisely because the `-0731` identity had no sheet prior to solve an alternative from. This is #215's candidate 2, and it is caused here, not in the watch.

Every unknown lane also starts at the 5-minute no-history wall, so the split multiplies the cost of every first call.

## Where it lives

`catalog.Concrete` / `catalog.Identity` — #186 made them level-blind. They should fold the alias, the concrete slug and the dated served slug onto one identity **before the lane ledger sees them**.

## Acceptance

- `~vendor/model-latest`, `vendor/model` and the dated served slug (`vendor/model-0731`) resolve to ONE ledger identity carrying the catalog's facts. A test pins this for the shipped default specifically.
- After a run on the shipped default, `v3/lanes.json` holds one entry for the model, with catalog facts, not three.
- A hedge verdict on the shipped default can solve an alternative — `Choice.Alt` is non-empty where the catalog knows a faster peer.

## Replication

[...]

Issue #138, opened 2026-08-31
reflex: empty-at-ceiling abandons the model for the whole session on one strike, bypassing the provider ladder


santoshkumarradha:
## What happens today

When the reflex model returns empty at its ceiling (finish `length`, no text), `internal/reflex` abandons the model **for the whole session** on effectively **one strike**, jumping straight to the low tier with:

> the reflex model answers nothing at its budget; using \<low-tier\> for this session

Observed in the wild on `nex-agi/nex-n2-mini` (reflex tier) falling to `deepseek/deepseek-v4-pro`, many sessions in a row. The chain, all verified in code and the live call log:

1. The endpoint serving the reflex model accepts `reasoning: {enabled: false}` and thinks anyway. That is learned once and persisted per **model** in `model-quirks.json` (`reasoning_disable_ignored`).
2. Once the quirk is known, `answerBudget` starts every call at `thinkingAnswerTokens` (2000). Most calls fit; occasionally the thinking pass overruns 2000 → empty at ceiling.
3. `reflex.go` `call()` only retries with more room `if budget < thinkingAnswerTokens` — false when the budget already started there. So a single long think goes **straight to `abandon()`**: session-permanent (`Session.unusable` is never cleared), and the notice above.
4. The provider ladder (hedge, endpoint-refusal rungs, affinity) never fires, because HTTP 200 + finish `length` is a *successful* call at the wire. Only reflex knows the answer was useless, and reflex jumps from model to tier fallback with no provider rotation in between. The quirk is also keyed per model, not per endpoint — an endpoint of the same model that honors the disable is never tried.

So the answer to "isn't there an escalation ladder that rotates providers of a model before changing the model?" is: there is, and this failure signature bypasses it entirely.

## Proposed design

Three changes, in ladder order:

**1. Make empty-at-ceiling a rung on the provider ladder, keyed per endpoint.**
Treat "finish `length`, empty text, at the requested ceiling" as an endpoint-failure signature the way a refusal is: the reflex retry goes out with the offending endpoint excluded (the `provider.ignore` / affinity machinery in `internal/provider/affinity.go` + `endpoints.go` already exists), and `reasoning_disable_ignored` is recorded per `(model, endpoint)` instead of per model. It hardens into a model-level fact only when every endpoint serving the model has shown it. If only one endpoint serves the model, this rung falls through instantly and costs nothing.

**2. A breaker with decay, not a session sentence.**
[...]

PR #368, opened 2026-09-02
routing: one refusal object for the strike, the ladder and the screen (#266)

### LINKED ISSUE #266: routing: a 'provider.only permits only: X' refusal strikes nothing, survives every relax rung, and is drawn as 'slow' — one refusal object for strike, ladder and screen

santoshkumarradha:
Closes #266 — parts 1, 2 and 3. Part 4 is deferred on purpose and filed as #374, linked below.

## The defect, and why it could never act

A measured chat run collected six HTTP 404s of one shape:

```
Providers serving <model>: digitalocean, streamlake, baidu, deepinfra,
but your request's provider.only preference permits only: coreweave
```

That sentence means one thing — **the machine this request demanded cannot serve this model** — and it is the one refusal class that is certain about a lane. Nothing acted on it, because the same fact was classified three times by three pieces of code that had each been told a different half:

| where | what it wanted | why the refusal fell through |
| --- | --- | --- |
| the strike, `velocity.go` `refuseUpstream` | `refusal.FromUpstream()` | OpenRouter puts `provider_name` in the metadata **only** when relaying somebody else's refusal, so a refusal the router answered for itself was structurally exempt |
| the ladder, `endpoints.go` `relaxationPlan` | `RequireParameters` / `Ignore` / `MaxPrice` | `provider.only` was on neither the list that offers rung one nor the list that clears it, so a pinned request climbed every rung still pinned to the machine that had said no |
| the frontier, `internal/lane` | the endpoints sheet | nothing wrote a wire refusal back, so the lane was scored and picked again — three separate times in that one run |
| the screen, `tui3/lanes.go` | `LaneNews{Trying}` | a refusal and a slow lane posted identically and were drawn identically, as `· slow · trying nextbit…`, and nothing withdrew the claim when nextbit died |

## The shape: one classifier, three readers

`internal/provider/refusalobject.go` is new and is the whole of the decision. It returns `laneRefusal{Kind, Lane, Terminal}` and nothing interprets a refusal anywhere else.

**The lane comes from our own request, never from their prose.** The refusal above names the machine twice in English; this build reads neither. It reads the composed `provider.only` the encoder actually sent (`Client.wirePreferences`), which this process wrote itself. There is no `strings.Contains(body, "permits only")` anywhere in the change, no provider-name table, and no special case for coreweave or nextbit. `TestTheStruckLaneComesFromOurOwnDemandAndNotFromTheRoutersProse` pins that with a refusal whose prose names a lane the request never asked for: the demanded machine is struck and the innocent one is left alone.

Three readers:

- **(a) the strike** — `refuseUpstream` paces the lane, and a *terminal* refusal also writes it out of the serving set.
- **(b) the ladder and the walk** — rung one now takes the whole provider object off, `only` and `allow_fallbacks` included (`providerPrefs.narrowing` is the one membership rule, so the rung that is offered and the rung that acts cannot drift); the race's walk strikes before it re-picks and never walks onto a machine the wire has since refused.
- **(c) the screen** — `RescueNews{Alt, Reason, Failed}` is narrowed from the same object and travels `internal/provider` → `internal/session` → `internal/tui3`, so the word a person reads is the word the ledger acted on.

**The serving set gains a negative half** (`internal/lane`'s `RefuseServing` / `Serves`, read by `frontier.capable`). Every refusal is filed under `LedgerModel` — the same helper #289 keys the sheet, the cache and the belief store through, so the bare-id/dated-slug disagreement that was half of defect 3 cannot reappear one layer down. It holds for 30 minutes rather than forever: a router's serving set really does change, and a permanent ban earned by one 404 would be a process that is wrong once and wrong until restarted.

## The replication, which did not exist

`internal/lane/lanestub` served its sheet and its completions from one slice, so the sheet-versus-router disagreement was unreachable — which is why a real defect had no replication a stranger could run.

- `lanestub.Lane.SheetOnly` publishes a lane on the endpoints page that `pick` will not serve, refused with the live `permits only:` body copied down to its punctuation.
- `lanestub.Server.RouterURL()` returns a base URL this build recognises as a router. **The product's gate was not changed; the stub was made to satisfy it.** `LaneSheetAvailable(base)` tests the base URL for `openrouter.ai`, so a binary driven through `AFORGE_BASE_URL` at a loopback stub fetched no endpoints page, had an empty frontier, and never demanded a lane — the exact replication #266 asks for was impossible end to end. Loosening that gate would have proved the harness works rather than the product.

## Evidence

**Evidence line, quotable verbatim:**

> `go test -tags e2e -run 'TestTUIRefusedLane' -count=1 ./internal/e2e/` passes against `bin/aforge` driven in a real tmux terminal by `AFORGE_BASE_URL` at a staged router of five tool-capable lanes, two of which it publishes and will not serve: the demanded machine is refused with the live `permits only:` 404, is asked exactly once and never again, the next request leaves 1ms later carrying no `provider.only`, no gap over 5s opens across ten requests in a 20s turn, and the status row of a turn that stops on a refusal reads `ghost refused` rather than `slow · trying …`.

### Why this evidence spends no model tokens, stated rather than left to be noticed

The standing rule is that approval **is** the real-model e2e. This PR does not meet the letter of it and the exception is deliberate, so here it is in plain sight rather than inferred from a green tick.

What is real: the replication is driven end to end through **`bin/aforge` itself**, in a real tmux terminal, typed at like a person and read back with `capture-pane`. The binary, the chat surface, the session, the chooser, the race, the ladder, the status row and the whole HTTP wire path are the shipped ones. What is staged is the **router**, and only the router, because the defect is a refusal **no live provider will produce on request**: `permits only:` means the endpoints page and the serving set disagree, and OpenRouter cannot be asked to disagree with itself on demand. That is exactly why #266 asked for the stub in the first place — it is what makes the staging possible at all, not a shortcut around it. No model tokens were spent and the run costs nothing, so it is cheap to repeat.

The checkable line: what a real model would have added here is coverage of the **completion path** — a genuine token stream behind the routing decision — and the two tmux scenarios already exercise that path against a stub speaking the same protocol, streaming SSE chunks that name their serving lane, through the same decoder, the same stall guard and the same phase clock. Sibling PR #358 in this wave carries a genuine real-model run, so the wave is not short of one.

### The four acceptance criteria

1. **The next request leaves within 2s carrying no `provider.only` for the refused lane.** Router log from the run: `+0s only=—`, `+1.014s only=Ghost`, `+1ms only=—`, and no later request names Ghost. Asserted in `TestTUIRefusedLane`; asserted again at the transport in `TestARefusedLaneIsStruckTheWalkMovesAndTheScreenIsToldRefused` and `TestRungOneDropsTheDemandWhenThereIsNowhereLeftToWalk`.
2. **The refused lane is never asked again — stub request count == 1.** `server.Requests(refused) == 1` in the tmux run and in the wire test, where two machines are refused and each is asked exactly once.
3. **The row renders `refused`, not `slow`, and the `trying X…` claim is retracted.** One tmux frame, at the moment the sentence is actually visible:

   ```
   voidws · refusal-void · ghost refused
   ```

   from `TestTUIRefusedLaneRowAfterTheTurnStops`, which stages a router that publishes five machines and serves none — the reported ending exactly. Pinned deterministically as well in `internal/provider` (the walk posts `refused`, then the retraction, then the next `refused`), `internal/session` (`Failed` arrives with `Trying` false) and `internal/tui3` (`· refused · trying coreweave…`, then `· coreweave refused`, and `slow` still drawn for a lane that was merely late).

   *Labelled a guess:* the sentence is not catchable **during** a walk because `render.go`'s `servedRider` gives that cell to the phase clock whenever `livePhase()` answers, and once a turn lands the answer's own news replaces it. I did not chase a frame the phase clock legitimately owns; `TestTUIRefusedLane` keeps the log line saying so, so the next reader does not think the assertion was dropped.
4. **No call-log gap over 5s while the turn is `working`.** Asserted over consecutive request timestamps in both the tmux run and the wire test. The staged router now stamps every ask (`lanestub.Ask.At`), which is what makes a gap measurable rather than argued.

Artifacts: the router's own request log is printed by both subtests; the tmux frame is in the `TestTUIRefusedLaneRowAfterTheTurnStops` log.

## Notes for review

- **Three strike sites, not four.** `refuseUpstream` is called from `client.go:1069`, `client.go` (the refusal delivered inside a 200) and `retry.go:280`, and the edit at each is the argument list only — no logic moved into either file. `client.go:433` looks like a fourth and is not: it calls `routingRefusal`, a different function.
- **PR #308 is open against `internal/provider/client.go`** and is not rebased on dev. Whoever lands second rebases; this is noted so it is not a surprise.
- This branch is rebased onto green `dev` at `35c1a79e`, and its serving set calls #289's `LedgerModel` rather than a second normaliser that would agree only by luck.
- **`internal/session` carries an intermittent flake that is not this change's — it is #392.** `TestTheCeilingIsDroppedWhenTheReaderAgreesNothingRemains`, `TestAContinuationSayingNothingIsLeftDropsTheCeilingHandover` and `TestADowryOfMachineMarkupIsRefusedAndNeverBecomesTheName` fail roughly one package run in three, the first two with `checkpoint_test.go: the turn ended as a assistant saying "(unscripted)"`. It was reproduced from this branch on an **untouched** detached worktree at `35c1a79e` — three runs of `go test ./internal/session/ -count=1` gave ok / FAIL / ok — and then diagnosed by another lane as a fixture defect (#392, bisected to #333): `nameAhead` starts the namer's model call concurrently with the handover's, and `scriptedCompleter` serves every caller from one FIFO by arrival order, so a namer that wins the race eats the step scripted for the turn. Two results from this branch line up with that mechanism exactly and are recorded because they are what ruled the alternatives out: all three stay green at `taskset -c 0 … -count=20` (one core removes the concurrency the race needs), and `checkpoint_test.go` contains no `t.Parallel` at all (the concurrency is the product's own goroutine, not the harness's). No `provider.Client` is constructed in any of the three, so this change's classifier, strike, ladder and serving set are not merely isolated from that path: they are absent from that test binary.
- **Manual updated in the same change** — `lanes.md` (a new "When a machine refuses to serve the model" section and two new status-row rows), `models-and-cost.md` and `screen.md`. Every person-facing string is quoted as the code spells it.
- **The tmux rigs can leave a binary behind.** A `bin/aforge` from one of this file's own rigs was still running 42 minutes after its subtest finished, on a box where nine sessions run e2e rigs — so that is nine idle binaries, not one. Reaped by pid here; worth somebody's look at whether `rig.kill` should wait for the process rather than only the tmux session, since a leaked one holds a state root and a lane beat.
- **#375** is filed as an observation with a ruling wanted, not as a defect: a person's strict lane pin re-demands a machine the wire has already refused, on every request. Pin semantics are outside this change and are untouched — what #266 fixes is the demand **this process makes on its own account**.

## The deferred part, and why the deferral is honest rather than silent

Issue #266's fourth part — the session emits no request-sent event, so `internal/tui3/app.go` (~:1066) **infers** the wait clock and honestly counted 26m47s of a wait for a request that was never sent — is **not** in this change. `app.go` was claimed by another lane for the duration of this work, and a new on-screen event kind lands as a feed method under the one-reducer law rather than being improvised across that fence.

It is filed as **#374**, with the replication and the acceptance, so the deferral is a linked issue rather than a gap.

---

## Two notes for review

**A 502 and the strike.** A `502 … Model is at capacity` that NAMES its provider strikes
that lane through `refusalUpstream` — the classifier reads `FromUpstream()` before any
status gate, so any status with a named upstream is a strike. One that names nobody is
deliberately left alone: `routingRefusal`'s status gate is 404/400, so a nameless 502
writes nothing to the serving set. Striking a lane we cannot identify would be striking on
a guess. The nameless case belongs to #138's per-endpoint breaker, on repetition rather
than on one sighting, and is filed as #445 with a measured sighting as its evidence.

**Evidence re-derived on the merge tip.** Rebased onto dev `1de89c08`, which carries #438
(the hedge row lock inversion, in this same file). No conflict: #438 hoists
`w.race.askedLane(w.arm)` out of `streamWatch.facts()`'s `w.mu`; this PR's seam is on
`HedgeReport`, whose `tell` copies the callback under `h.mu`, releases it, and only then
calls it — no callback here runs under any lock, so the two changes cannot disagree about
who holds what. Full suite on the tip: `EXIT=0`, 85 packages, zero failures. Both tmux e2e
subtests pass on the tip (`--- PASS: TestTUIRefusedLane`, `--- PASS:
TestTUIRefusedLaneRowAfterTheTurnStops`), with the refused machine asked exactly once and
the next request out 1ms later carrying no `provider.only`.

---

## Frozen paths this PR touches, and why each one is the defect's own seam

The freeze law is satisfied through this entry, not by loosening it. #266 IS the defect in
this seam, so the classifier's reach is the boundary; nothing here goes past it.

| file | why it is unavoidable |
|---|---|
| `internal/provider/refusalobject.go` | new — the one classifier the issue asks for |
| `internal/provider/refusalobject_test.go` | new — pins the laws, including that the struck lane comes from our own demand and not the router's prose |
| `internal/provider/endpoints.go` | `routingRefusal` is defect 1's half of the mirror; `relaxationPlan` is defect 2 |
| `internal/provider/velocity.go` | `refuseUpstream` is defect 1's other half; `relaxedPreferences` is the rung that could not drop a pin |
| `internal/provider/waitreport.go` | the ONE seam a rescue's news reaches the session through; it carried a lane name and nothing else, so `Reason: refused` and the retraction had nowhere to travel |
| `internal/provider/hedge.go` | the walk (~:903) is where a refusal is seen and re-picked without striking; `settle` is untouched |
| `internal/provider/wire.go` | the preference object was composed here and read elsewhere, which is why rung one never saw a rescue's demand; composition is now named once |
| `internal/provider/client.go` | three call lines only — the `refuseUpstream` argument list, so the strike can read what the request demanded |
| `internal/provider/retry.go` | the third of those three call lines |
| `internal/provider/hedge_test.go` | the rig gains the refusal-walk case |
| `internal/lane/sheet.go` | the negative half of the serving set, keyed through `LedgerModel` (agreed with #289/#353), plus `frontier.capable`'s read of it |
| `internal/lane/frontier.go` | that read — a lane the wire refused must stop being capable |
| `internal/lane/serving_test.go` | new — pins the refusal's key and its expiry |
| `internal/lane/lanestub/lanestub.go` | `SheetOnly` and the second mount, without which the issue's own replication cannot be staged at all |

Nothing outside the classifier's reach is touched: no `apply()` arm, no `feed.go`, no
`app.go`, no pin semantics in `lanes.go` or `lanepin.go` (see #375), and the hedge verdict
path is left to #289.

---

## One class this PR does not cover, named rather than left to be found

**A pinned lane that cannot serve the model still kills the turn** — issue #456, measured
on the live router by another session, on dev and on this branch alike.

With `lane.talk` pinned to a machine that never serves the model, the hedged aux call gets
the `permits only:` 404, walks, and lands. The turn's own UNHEDGED call gets the same 404
and ends the turn: `failure{class work, reason "the request itself was refused"}`, no walk,
no rung, no second attempt. And because the walk is what strikes, a refusal absorbed by the
ladder never reaches `refuseUpstream` at all — `client.go:440-459` hands the refusal back
when `canWalk()` and calls `recoverFromRefusal` otherwise, and neither branch strikes — so
`lanes.Serves(model, lane)` stays true and the next turn makes the same demand.

This PR neither causes it nor fixes it: dev behaves identically. It is named here because
the four acceptance criteria CANNOT see it — the staged scenario always has somewhere to
walk, so the stub never reaches the branch that drops the strike. That is the honest limit
of the replication.

It is distinct from #375 (whether a person's pin should retire after a refusal), and the
fix — strike first, then offer rung one and retry once on the unhedged path — lands on top
of this refusal object rather than inside it.

---

## Gate, on the base this merges into

Whole-tree run of the gate's own commands (`make fmt-check`, `make test-laws`, `make test`) on `8b04b6917` over dev `1ecf8e8d7`, in the natural environment, full log kept:

```
FMT=0  LAWS=0
ok    internal/tui3      489.272s
ok    internal/session   111.436s
ok    internal/provider   52.912s
FAIL  internal/lane        6.885s
```

**The one red is not this branch's**, and it is the only one in the whole run:

- `internal/lane` / `TestAnEmptyLedgerIsAnEmptyChoice` — **#475**. `internal/lane/sheet.go`'s `cachePathIn` falls back to `home.Join("v3","lanes")`, so the test reads `~/.aforge/v3/lanes/<model>.json`. This box has one. Proven both directions on a clean `origin/dev` worktree with no branch of mine in it: `AFORGE_HOME=<empty temp dir>` → ok, real home → FAIL. CI runs on a clean box and passes.

`internal/tui3`'s `TestAMissingTypedWindowsPathNamesOneFileAndNeverSubmits` (#483) was the second red on the previous base; #487 reverted #483 and tui3 is green here.

Two notes on how this was checked, since both nearly produced a false green:

1. An earlier run of this gate exported an empty `AFORGE_HOME` to work around #475. That fixes `internal/lane` and **breaks** `internal/rtk` and `internal/resident` — proven both ways on clean dev. No blanket export can satisfy all three; the override was removed and this run uses the natural environment.
2. The gate previously piped `make test` through `tail -60`, which truncated the log at `internal/lane` and hid which lane test failed. It now keeps the full log. The 37 packages sorting before `internal/lane` were additionally re-run on their own to cover the previously truncated range: `make exit=0`, no failures.


santoshkumarradha:
Review (santosh-1a, review-only; codex verification + my own reads and test runs). **No blocker.** Two wording defects to fix in-PR, one edge to state, and a frozen-path list for the orchestrator.

**One classifier — holds.** `laneRefusalFor` (refusalobject.go:142–157) is the only place a refusal becomes `{Kind, Lane, Terminal}`; the lane is `onlyLane` over `wirePreferences` (:204), never the body. No production code matches `permits only` (the hits are comments and the JSON tag). Strike (velocity.go:722) and walk (hedge.go:723) consume the object; the remaining body reads (endpoints.go:200 envelopes, velocity.go:742 429 pacing) are other classes and do not re-derive this lane. *Edge to state in a comment:* `onlyLane` answers "" when `len(Only) != 1`, so a two-lane `only` that is refused is non-terminal and strikes nothing; today a person's pin is scalar (lanes.go:440), so it cannot happen, but the rule should say so where it is decided.

**Rung one — behaviour right, words wrong.** `narrowing()` (velocity.go:403) is the one membership rule and `relaxedPreferences` (:418) clears every field in it, `Only` and `AllowFallbacks` included; a refused strict pin retries unpinned. It deliberately keeps `Sort`/`Order` (:447, pinned by refusalobject_test.go:301, "a preference that still ranks was dropped whole"). That is the better behaviour — ranking is not membership — but the changelog (368-refusal-object.md:15) and the PR body say "drops the WHOLE provider object", which is not what ships. Say "every membership restriction comes off; the ranking stays".

**RescueNews seam — holds.** `{Alt, Reason, Failed}` minted at refusalobject.go:102; session carries them and sets `Trying: !Failed` (lanenews.go:193); tui3 copies both fields (tui3.go:954); the words live in one table, `laneRider` (lanes.go:1079). Your `servedRider` guess is mechanically right: render.go:1941 gives a live phase precedence over lane news, which is why no mid-walk frame is catchable; it stays a guess, not an asserted fact, and the test log says so.

**Staged router — faithful for this defect class; the exception is justified.** The 404 body (lanestub.go:693) matches #266's recorded line including the omitted `provider_name`; `RouterURL()` (:346) satisfies the unchanged `LaneSheetAvailable` gate (provider/lanes.go:643, absent from the diff); the stub speaks plain HTTP SSE (:745) so the real decoder, stall guard, retry path and phase clock run; `Ask.At` (:581) makes the 5 s gap measurable. What it does not cover, and a real-model run would have: auth, the 200-with-refusal envelope, the upstream `metadata.provider_name` relay, and live network timing. None of those is the defect being fixed.

**Manual defect:** models-and-cost.md:2103 says the refused machine "leaves the set aforge chooses from for the rest of the session"; the serving-set hold expires after 30 minutes (sheet.go, and your own body says so). Say "for the next half hour".

**Also fine:** negative serving set keyed through `LedgerModel`, lowercased, expiring; the three strike sites changed argument lists only; changelog has `pr: 368`; no machinery vocabulary in new person-facing literals; targeted provider/lane/session/tui3 tests pass here; build and vet clean. #374 and #375 exist as described.

**For the orchestrator's ruling, not the author's:** this PR touches the frozen paths `internal/lane/{frontier.go,sheet.go,serving_test.go,lanestub/lanestub.go}` and `internal/provider/{client.go,endpoints.go,hedge.go,hedge_test.go,refusalobject.go,refusalobject_test.go,retry.go,velocity.go,waitreport.go,wire.go}`. Nothing under `internal/lane/control`, `bench/lanelab`, `internal/calllog`, or tui3's frozen four.

Trial merges: clean against #358, #361 and each of #337/#351/#355.

Issue #371, opened 2026-09-02
dev: Agent.askTask holds 18 decisions and the ceiling is 15 — #279 crossed the endings ratchet and merged


santoshkumarradha:
## Observable failure

`origin/dev` is red from 2183719a (#279) through 8d590ccd:

```
go test ./internal/session/ -run TestNoRoadInTheTaskEngineHasMoreEndingsThanItsLedgerRow -count=1
complexity_test.go:148: Agent.askTask (task.go) holds 18 decisions and the ceiling is 15.
    There is no road to adding a row to complexityDebt: that ledger is the debt this gate was fitted around and it only shrinks. Split the function along the phases its own comments already draw.
```

Every session running the `internal/session` suite locally tonight sees it.

## The law

`internal/session/complexity_test.go`: no function in `task*.go` may hold more than fifteen decisions unless it was ledgered when the gate was fitted (7720d9b6, 2026-09-01), and the ledger only shrinks. `Agent.askTask` was at 14 before #279.

## Which change

Bisected across the only two commits that touched `internal/session/task.go` since 779ed018:

| commit | askTask | gate |
| --- | --- | --- |
| 779ed018 | 14 | green |
| 2183719a (#279, proposal waits while you type) | 18 | red |
| 69029c4e (#333, task on attached repository seals) | 18 | red |

#279 alone crossed it: the `for`/`select` wait loop gained a `hold` case and the expiry case gained the two `held` checks. #333 did not change the count.

## Resolution

1. #367 reverts #279 from `dev` so `dev` is green again (standing order: `dev` is never red — yank, then fix, then merge).
2. A follow-up re-lands #279 with `askTask` split along the three phases its own comments already draw (admission under the lock, the announcement to the hub, the wait), crediting the original author. No ledger row, no raised ceiling, no known-red line.

Why the PR gate let it through is a separate issue.

PR #401, opened 2026-09-02
ci: the pull-request gate runs every law, and one proof means one thing

### LINKED ISSUE #372: ci: the PR gate runs only the Manual tests of internal/session, so a structural gate red on dev merges green
## Observable failure

Two pull requests (#279, #333) merged to `dev` through a green `check` while `internal/session`'s `TestNoRoadInTheTaskEngineHasMoreEndingsThanItsLedgerRow` was red on the branch and on every merge commit after. The "PR gate" workflow reported SUCCESS on `dev` at 69029c4e and 8d590ccd with that test failing (#371).

## Cause

`.github/workflows/ci.yml`, last step:

```
go test ./internal/manual/
go test -run 'Manual' ./internal/tui3/ ./internal/session/
```

Only tests whose names match `Manual` run. Every structural gate in `internal/session` that is not spelled with that word is invisible to the PR gate: the endings ratchet (`complexity_test.go`), and any sibling written the same way — cheap `go/ast` or table checks that decide in under a second and whose answer is the same on every machine. These are exactly the kind of check the workflow's own header says it exists to run ("does the tree compile, does vet find a real bug, and does the manual still know"), and they were not added when they were fitted.

The full suite (`ci-full.yml`) only runs on the way into `staging` and nightly at 09:00 UTC. It has never been green: every recorded run is `failure`. The 2026-09-02 nightly (run 33635705493, on 779ed018, before either PR) failed on:

- `full tests 1/3`: `internal/tui3` hit the 8m per-package timeout with `TestTheStripCannotOutliveTheRowItWasOpenedOn` still running.
- `cross build (windows/amd64)` and `(windows/arm64)`: `internal/standing/running.go:126: undefined: unix.Kill`, `unix.EPERM`, `store.go:352: undefined: unix.Flock`.

A nightly that is always red is a log line nobody reads, so it names the day nothing.

There is no branch protection on this repository (private, free plan), so the gate can only be the PR workflow itself.

## Fix shape

1. The PR gate runs every structural test that can make `dev` red, within the time a PR can wait for. `internal/session`'s structural tests are sub-second; run them by name alongside the `Manual` ones, or better, mark them with a common prefix or a build tag the gate selects, so the next one written is picked up without editing the workflow. The gate step's own comment must say why the selection is what it is.
2. A full-check failure on `dev` is a page, not a log line: the nightly on failure opens or updates one issue (or notifies), so a red nightly is seen by somebody the same morning.
3. Separate from this issue: the windows cross build and the tui3 timeout above are real reds owed their own fixes.

santoshkumarradha:
Closes #372.

## What was wrong

- `ci.yml`'s only engine test step was `go test -run 'Manual' ./internal/tui3/ ./internal/session/`. No structural test ran on a pull request; #279 crossed the endings ratchet and merged green, and so did #333 after it (#371).
- `ci-full.yml` had never been green. Its `-timeout 8m` cut `internal/tui3` (about 485s on the runner) off at the finish line and reported a test that had been running for two seconds as a hang. Nobody saw it, because a scheduled run reports to nobody.
- `make test` was a bare `go test ./...` that skipped nothing, so `make check` could not pass on a clean tree while CI skipped the ledger. gofmt was not in the gate and two files had drifted.
- Nine sessions running `./...` at once tonight manufactured reds by load.

## What this does

- **The laws hold** step: `make test-laws` runs every test in a file that imports `go/ast` or `go/parser`, selected at run time by `scripts/laws.sh`. 25 files, 13 packages, about 20s after the link. A new law is on the gate the day it lands; there is no list.
- **`touched packages`** job: the full suite of every package the change touched, through `make test`, beside `check` rather than after it. Reports today; promote when green twice, the way `full tests` is meant to be.
- **gofmt** in the gate (`make fmt-check`), and the two drifted files formatted.
- **One reading of known-red.** The Makefile reads the ledger once; `make test`, `touched packages` and `ci-full` all go through it, with `TEST_TIMEOUT := 15m` measured from the slowest package with headroom. An empty or absent ledger skips nothing.
- **The ledger only shrinks.** `internal/ci` ratchets `knownRedEntries` (18 at 35c1a79e) downward and checks every listed name is a declared test. The commit that deletes the file deletes the test.
- **One suite per box.** `make test` on `./...` takes `scripts/one-suite.sh`'s lock: pid with cmdline checked, stale locks taken, named-package runs unaffected.
- **A red nightly is a page**: `ci-full` opens or updates one issue on a scheduled failure.
- `docs/rules/ci.md` and `CLAUDE.md` rewritten to match; CLAUDE.md points at the ledger instead of copying it.

## Evidence

- `make test-laws` green on the branch (all 13 packages ok).
- `make fmt-check`, `go build ./...`, `go vet` on the touched packages green.
- The suite lock proven by hand: refuses a second whole-tree run naming the holder, passes the child's exit status through, clears on exit, takes a stale pid, and forwards a stop to the child.
- Both workflow files parse.

Not verified here: the `page` job and the `touched packages` job only run on GitHub; the first real run of this PR's own gate is the proof of `touched packages`, and the next scheduled nightly is the proof of `page`.

## And it would have caught the next one

While this PR waited, #437 (5995a6e7) landed on `dev` and twelve `internal/tui3` tests went red — reveal pacing moved to the wall clock while the tests settle synchronously — through a gate that runs no tui3 test but the manual ones. The `touched packages` job here runs the full suite of every package a change touches; #437 touched `internal/tui3`, and this job on `dev` would have shown those twelve on the pull request instead of on the trunk. That is the sentence for this PR.


santoshkumarradha:
## The touched-packages job's first run found three runner-only reds

On its first real run (33645143035) `touched packages` ran `internal/cachedir`, `internal/ci` and `internal/tui3`; the first two passed and tui3 failed three tests that pass on a Linux workstation and are in nobody's ledger:

- `TestTheSweepCopiesExactlyWhatItLit` — `ground_test.go:120: copied "The `internal/manual/` directory was deleted, along with the rest."`
- `TestAnAnswerAskedFromHomeIsRenderedMarkdownAndNotItsSource` — `homeaskmd_test.go:70: the answer still carries "`"`
- `TestAHomeAnswerIsRawWhileItFormsAndStyledOnceItSettles` — `homeaskmd_test.go:111`, same

Markdown code spans are drawn raw on the runner. Guessed cause: the renderer chooses a plain style from the runner's terminal environment, so backticks survive. This PR only formats `ground_test.go`, so these predate it; they are exactly what the job exists to surface, and they are reported to the test-hygiene lane rather than fixed here. The job reports and does not block, as the workflow comment says, until it has been green twice.

PR #449, opened 2026-09-02
--one-model: every rung rides the one model, and a role with no model says so (#443)

### LINKED ISSUE #443: --one-model: the mark reader and the brief writer have no model at all, and the person is told "the second model could not be reached"
## Observable failure

Measured on the canary's first chat cell (#407): dev 35c1a79e, `aforge chat -yolo -one-model -model deepseek/deepseek-v4-flash`, a real issue pasted. The turn made the fix, then the write seam moved the rest to a task, and the person read:

```
this is changing more than a quick edit · moving it to a task that is watched and can split · carrying the ask only — the brief could not be written: the second model could not be reached
this looked like work, so task 1 started: paste 1:
```

Nothing had been unreachable. The session journal says what actually happened:

```
{"type":"mark","mark":{"n":1,"rounds":10,"decision":"failed"}}
{"type":"carry","carry":{"rung":"handoff","outcome":"failed","reason":"roles: no model for role \"handoff\""}}
{"type":"carry","carry":{"rung":"draft","outcome":"nothing-left"}}
{"type":"carry","carry":{"rung":"ask","outcome":"written","chars":3756,"used":true}}
```

The worker opened on the bare paste and spent its first nine calls re-reading the file the turn had edited, re-reading the test file the turn had written, and re-running the tests the turn had already run green. No call was made for the brief; no call was made for the mark. The rig's own profile pinned `models.tiers.mastermind` to the same model, so the same run **without** the flag would have had both.

## Cause

Two pieces of code each keep a law the other breaks.

- `applyV3Governance` (cmd/aforge/chatv3.go) under `--one-model` sets `cfg.RolesSource = nil` on the stated ground that "every rung below already ends at the session model when nothing answers".
- `writeHandoff` and `readMark` (internal/session/checkpoint.go) call `callRole` with an **empty** session default, on purpose and with a comment: crew-only, "an install with no mastermind gets no handoff writer at all rather than a fall-through to the conversation's own model".

Under the flag the ladder for those two roles is therefore empty — no pin, no tier, no floor — and `roles.Ladder` returns `ErrNoModel`. Then `carryFault` says "the second model could not be reached" for every error that is not a deadline, so a role with no model reads as a wire failure.

## Law

**One model means one model at every rung**, including the rungs that today refuse to fall back to the conversation's model. The crew-only refusal is a quality judgement about a profile that *has* a crew; a person who passed `--one-model` has said the conversation's model is the crew, and the next crew-only caller must be safe by construction rather than by remembering the flag.

**The surface says what happened.** "Could not be reached" is a wire event. A role with no model, a client that is absent, a provider refusal and a deadline are four different facts, and the line a person reads has to name the one that occurred.

## Fix shape

- Under `--one-model`, every role resolves to the live conversation model. Do it once at the seam where the empty default is honoured (`callRole`, reading a `Config` bit set by the door) or by giving the flag a roles source that answers every tier with the session model — whichever leaves a third crew-only caller correct without knowing the flag exists. The flag's comment in `applyV3Governance` then becomes true.
- `carryFault` tells `roles.ErrNoModel` (and the absent-client case) apart from a wire failure and says so in the person's register, e.g. "no second model is set", keeping the provider's words for refusals and "did not answer in time" for deadlines. The journal keeps the exact error.
- The manual's tasks page gets the new sentence beside the existing two.

## Replication

In an empty directory, with a profile whose only model rows are the four pre-worker tiers (the canary's `bench/canary/lib/home.sh` writes exactly that):

```
aforge chat -yolo -one-model -model deepseek/deepseek-v4-flash -max-cost 1
```

Paste an issue whose fix takes more than a couple of edits and more than ten tool rounds, then:

```
grep -h '"mark"\|"carry"' <state root>/v3/projects/*/*/transcript.jsonl
```

Today: `"decision":"failed"` on the mark and `roles: no model for role "handoff"` on the carry, and the transcript line above. The canary rig (`bench/canary/run.sh --bin <aforge>`, chat door) is the measured form of the same shape.

## Acceptance

A `--one-model` chat run whose turn hands off produces a mark **and** a brief: the journal's mark row carries a decision other than `failed`, the carry rows show the `handoff` rung `written` and `used`, the person's line carries no "carrying the ask only" clause, and the worker's first calls act on the brief rather than re-deriving the turn (nine re-reading calls is the number to beat). A run with no model for a role at all — no flag, no crew, no session model — reads "no second model is set" in the thread and `roles: no model for role` in the journal, never "could not be reached".

santoshkumarradha:
Closes #443.

## What was true

`--one-model` withheld the roles ladder (`cfg.RolesSource = nil`) on the stated ground that "every rung below already ends at the session model when nothing answers". Two callers keep the opposite law on purpose: `readMark` and `writeHandoff` hand `callRole` an **empty** `sessionDefault`, because the mark reader and the brief writer are crew-only — an install with no mastermind should get no second opinion rather than the running model marking its own work.

Under the flag those two rungs therefore had no pin, no tier and no floor, `roles.Ladder` returned `ErrNoModel`, and a measured chat run had **neither a mark reader nor a brief writer**. `carryFault` mapped every non-deadline error to "the second model could not be reached", so a role with no model read as a wire failure, and the worker opened on the bare paste and spent its first nine calls re-deriving the turn.

## What is true now

- `session.Config.OneModel` carries the door's promise into the engine, set by `applyV3Governance`. Its comment no longer claims withholding is the whole of the flag.
- `callRole` answers it once, at the one seam every errand passes through: a set flag replaces the caller's floor with the conversation's live model. The next crew-only caller is correct without knowing the flag exists.
- The crew-only callers still pass their empty floor, and their refusal stands unchanged for every run that did not type the flag.
- `carryFault` tells `roles.ErrNoModel` and `errNoCompleter` apart from a wire failure and says `no second model is set`. The journal keeps `err.Error()` either way.
- The manual's tasks page gains a section for that sentence, and the `--one-model` section says the reader and the writer ride the one model too.

## Tests

- `internal/session`: `TestUnderOneModelACrewOnlyErrandRidesTheConversationsModel`, `TestWithoutOneModelACrewOnlyErrandStillRefusesToFallToTheConversation`, `TestARoleWithNoModelIsToldApartFromAProviderThatCouldNotBeReached`
- `cmd/aforge`: the two `--one-model` governance tests now assert the bit travels, and does not travel without the flag
- `internal/manual`: a probe in the asker's own words

## The propagation into child configs is proven by a unit test, not by a journal

A task node never reaches a crew-only errand today: `checkpoints()` (internal/session/checkpoint.go:2214) returns false for `InTask`, which `newTaskAgentOn` sets, and `TestTheTurnsThatMustNeverBeCheckpointedAreNot/a_node` pins it. So copying `OneModel` into the four child configs (task_run, fork, orchestrate, task_audit) changes no journal row on any run this build can make. It is there so that the next road that opens a checkpoint to a node, or the next crew-only caller, is right without knowing the flag exists, and `TestTheOneModelPromiseTravelsToATaskNodeAndItsCrewOnlyErrands` shows the crew-only path on a child resolving to the inherited model. The second real-model replay (comment below) confirms the conversation side on the rebased head and shows zero mark/carry rows in the three child journals, as the code predicts.


santoshkumarradha:
## Real-model end-to-end, on the built binary

`~/af-443/bin/aforge`, a fresh state root written by the canary's `canary_home` (deepseek/deepseek-v4-flash on every tier, `daily_budget_usd: 1`), the interactive surface driven in tmux exactly as `bench/canary/lib/chat.sh` drives it:

```
env AFORGE_HOME=<home> TERM=xterm-256color aforge chat -yolo -one-model -model deepseek/deepseek-v4-flash -max-cost 1
```

The paste asked for a six-module Python package plus a test file, run the tests, fix what fails. The turn edited files, then the write seam moved it to a task.

### The journal, verbatim

```json
{"type":"mark","mark":{"rounds":1,"model":"deepseek/deepseek-v4-flash","costUsd":0.00018791,"sketch":"write modules & test > run pytest > fix failures > report count","decision":"wrote","durationMs":7938}}
{"type":"carry","carry":{"rung":"handoff","outcome":"written","chars":1758,"used":true}}
{"type":"carry","carry":{"rung":"draft","outcome":"written","chars":2187}}
{"type":"carry","carry":{"rung":"ask","outcome":"written","chars":1143}}
{"type":"mark","mark":{"n":1,"rounds":10,"model":"deepseek/deepseek-v4-flash","costUsd":0.000248594584,"sketch":"(done)","decision":"continue","durationMs":1616}}
```

Before this change those two rows read `"decision":"failed"` and `"outcome":"failed","reason":"roles: no model for role \"handoff\""`.

### The line the person read

```
this is changing more than a quick edit · moving it to a task that is watched and can split
this looked like work, so task 1 started: build ledger package
```

No `carrying the ask only` clause.

### What the worker opened on

A 4,370-character brief carrying the person's words plus 1,758 characters of written handoff, which told it the directories were empty and named every module's contract. Its first calls were `ls`, two directory reads and `pytest` — not the nine re-reading calls of the measured run. The task reached `state: done` with `22 passed`. Whole run: $0.03.

santoshkumarradha:
Second real-model replay on the rebased head d6264ee7 (tmux-driven, $0.06), aimed at the child-errand path.

Conversation side, verbatim call rows of the two crew-only errands under `--one-model`:
```
{"type":"call","call":{"model":"deepseek/deepseek-v4-flash","endpoint":"StreamLake","role":"markreader",...}}
{"type":"call","call":{"model":"deepseek/deepseek-v4-flash","endpoint":"Alibaba","role":"handoff",...}}
{"type":"carry","carry":{"rung":"handoff","outcome":"written","chars":1735,"used":true}}
```
The assistant line carried no "carrying the ask only" clause even though the draft rung came back degenerate this time.

Child side: a task node never reaches a crew-only errand today — `checkpoints()` (checkpoint.go:2214) returns false for `InTask`, which `newTaskAgentOn` sets, and `TestTheTurnsThatMustNeverBeCheckpointedAreNot/a_node` pins it. So the propagation into the four child configs is correct by construction rather than observable: the child errands that did run (`division`, `title`) resolved to the inherited model, and `TestTheOneModelPromiseTravelsToATaskNodeAndItsCrewOnlyErrands` shows the crew-only path on a child resolving to the inherited model the day a road opens it.

PR #470, opened 2026-09-02
provider: isOpenRouter strips the alias marker before the prefix check (#319)

### LINKED ISSUE #319: Client.isOpenRouter reads config.Model raw, so a ~openrouter/… spelling behind a proxy base silently disables the lane path
## What happens

`Client.isOpenRouter` (internal/provider) reads `config.Model` raw, with no normalisation. A model configured with the alias marker, e.g. `~openrouter/…`, behind a base URL that is not `openrouter.ai`, fails the model-prefix half of the check — and the whole lane path (sheets, beliefs, choosing, hedging) silently turns off for that client.

Found while writing #300's e2e: the test had to spell its alias without the `~` purely to get the lane path to engage against the stub's base URL.

## Why this is wrong

Everywhere else the `~` marker is a prefix on a name, not part of one — `catalog.Servable`, `lanes.LedgerModel` and `normalizeModel` all strip it before deciding anything. One raw read makes the lane path's on/off switch disagree with every other reader of the same string. Costs nothing in production today (the real base URL answers the question first), which is exactly why it will bite later, when someone routes an `~openrouter/…` model through a proxy base.

## Replication

Unit-level: construct a `provider.Config` with `Model: "~openrouter/anything"` and a non-openrouter `BaseURL`; observe `isOpenRouter` false and no lane machinery engaged. Same config without the `~`: lane path on.

## Acceptance (red first)

1. `isOpenRouter` (or its caller) normalises the model spelling before the prefix check: `~openrouter/x` behind a non-openrouter base engages the lane path exactly as `openrouter/x` does.
2. A test pins both spellings to the same answer, plus the negative (a non-openrouter model behind a non-openrouter base stays off).

AbirAbbas:
## Fix

`Client.isOpenRouter` read `config.Model` raw, so a model spelled with the alias marker (`~openrouter/…`) failed the prefix check behind a non-openrouter base and the lane path silently turned off.

It now normalizes before the prefix check, reusing the existing `normalizeModel` (which strips the leading `~` and lowercases). The BaseURL half and every other caller are untouched.

## Test

`TestIsOpenRouterReadsTheAliasMarkerOffTheModel` pins that `~openrouter/anything` and `openrouter/anything` answer the same, and that a non-OpenRouter model stays false.

Closes #319.


santoshkumarradha:

PR #483, opened 2026-09-02
chat: preserve arbitrary text across naming and send boundaries

### LINKED ISSUE #441: Preserve arbitrary chat text across slash, paste, and naming boundaries
## What happened

On 2026-09-02, `dev@09f1b3df`, a real v3 chat task started from a folded multi-line paste appeared in the task rail as:

```text
paste 1
```

The `taskname` call completed in 385 ms on `deepseek/deepseek-v4-flash-0731`; it received the full request, but the shared name cleaner accepted the paste transport label as a finished name. The task itself was admitted about 16 seconds later, so this was neither missing context nor a blocking name call.

The same audit found related text-boundary failures:

- every line beginning with `/` is consumed by the command dispatcher, so `/api/v1 returns 500`, `/user`, or `/Users/alice/repo fix this` never reaches chat or task routing;
- folded paste contents are not unfolded before inline `/task`, inline `/standing`, or a picked harness, so those doors receive `[paste 1 · N lines]` instead of the person's text;
- paste expansion uses textual `ReplaceAll`, so a literal lookalike such as `[paste 1 · 4 lines]` elsewhere in the same draft is also replaced and the person's text is mutated;
- the session-title role runs synchronously after `EventTurnDone`, leaving the session marked running while a cheap naming call is still in flight;
- slash command tokenization only splits on an ASCII space, so a tab or newline after a valid command word is misclassified.

No core transport escaping defect was found: once submitted, quotes, slashes, backslashes, Unicode, code fences, `$`, and `@` are carried as ordinary Go strings.

## Replication

**Deterministic (no model).**

From a clean checkout of `dev@09f1b3df`:

```sh
go test ./internal/session -run 'Test.*(TaskName|Title)' -count=1
go test ./internal/tui3 -run 'Test.*(Slash|Paste|Task|Harness|Standing)' -count=1
```

The current pure helpers accept `cleanTaskName("paste 1:") == "paste 1"`. A fake-agent v3 app consumes `/api/v1 returns 500` as `unknown command: /api/v1` instead of submitting it. A folded paste followed by an inline `/task` marker hands the fake task door the visible chip token rather than the fenced paste contents.

**Field (real models).**

With `OPENROUTER_API_KEY` set, run the isolated tagged regression added by the fix:

```sh
go test -tags e2e -run '^TestTaskNamingE2E$' -count=1 -timeout 8m -v ./internal/e2e/
```

It must pin the talk, shaper, task-name role, and tier fallbacks to exact model `deepseek/deepseek-v4-flash-0731`, start one disposable task from a paste-wrapped path-heavy request, observe a meaningful name, and cancel the task immediately. Expected cost is below one cent and expected wall time is under two minutes.

## Where

- `enterLine` and `slash` in `internal/tui3/input.go` and `internal/tui3/app.go`
- `homeEnter` and `homeView.runLabel` in `internal/tui3/home.go`
- `composed` / `unfoldPastes` in `internal/tui3/pastechip.go`, plus the task, standing, and picked-harness send doors
- `cleanTitle`, `maybeTitle`, `taskNameNeeded`, `cleanTaskName`, and `nameAhead` in `internal/session/title.go` and `internal/session/taskname.go`
- model-proposed task title admission in `internal/session/task.go`, and planner-node validation in `internal/orchestrate`
- slash command parsing in `internal/tui3/app.go`

## The fix

Make the syntax boundary explicit and shared:

- only a recognized slash command or alias owns a leading slash; unknown slash-prefixed input is ordinary text and is submitted unchanged on chat and Home;
- real dropped files and real workspace folders retain their existing meanings, and known commands continue to dispatch;
- every route that speaks for the composer consumes the same composed payload exactly once, unfolding paste chips while retaining the compact shown text;
- paste unfolding replaces only the tracked chip occurrence, never identical literal text the person typed;
- naming sees the meaningful payload rather than transport scaffolding, and the shared cleaner rejects generic ordinal labels such as `paste 1`, `task 2`, `part 3`, `step 4`, and `node 5` without rejecting semantic names such as `issue 252 check` or `phase 2 rollout`;
- shaper titles, `propose_task` titles, and orchestrated planner-node titles pass the same validity law before they can suppress asynchronous naming;
- a rejected ahead name falls through asynchronously to one retry from the shaped summary and brief;
- session naming runs after the first turn without keeping the session in its running/steering state;
- command splitting recognizes Unicode whitespace while preserving the argument text apart from documented outer trimming.

## Acceptance

- **e2e:** A real `deepseek/deepseek-v4-flash-0731` run receives a paste-wrapped request containing paths, quotes, backslashes, Unicode, `$`, and `@`; the task is admitted without waiting for naming and its persisted/published title becomes a meaningful two- or three-word label, never a paste/task ordinal.
- **e2e:** Through the fake-agent v3 chat and Home doors, `/api/v1 returns 500`, `/user`, `// literal`, and `/Users/alice/repo fix this` are submitted exactly once and unchanged; `/help`, a command alias, a real dropped file, and a real workspace folder keep their existing behavior.
- **e2e:** Folded paste contents reach inline `/task`, inline `/standing`, and a picked harness in full, while the visible transcript/card retains the compact paste chip and the paste is consumed exactly once.
- **Unit:** A literal string identical to a visible paste token remains literal when a real paste chip is unfolded elsewhere in the same draft.
- **Unit:** Generic ordinal names and paste/fence scaffolding are rejected or removed consistently by session, task, shaper, `propose_task`, planner-node, resume-healing, and ahead-name paths; semantic numbered names remain valid, and a rejected ahead result causes exactly one non-blocking fallback rename persisted to the checkpoint.
- **Unit:** A blocked session-title provider cannot keep the session in running/steering state after the turn is done; the title may land and persist later, exactly once.
- **Unit:** `/task` arguments containing tabs/newlines, absolute paths, spaces, quotes, backslashes, Unicode, `$`, and `@` reach `StartTask` byte-for-byte apart from documented outer trimming.
- The chat manual explains that only recognized leading slash words execute and that unrecognized leading slash text is sent as prose. The change entry invalidates the former all-leading-slashes-are-commands rule.

santoshkumarradha:
Closes #441.

## What changed

- Treat only recognized slash commands and aliases as commands. Unknown slash-prefixed text remains ordinary prose, while real dropped paths and known commands keep their existing behavior.
- Carry exact folded-paste identity through every composer lifecycle and send door, and unfold only the tracked chip occurrence exactly once.
- Validate task, session, shaper, proposal, and planner names consistently; reject transport ordinals, paths, one-word output, and sentence-like labels.
- Move session/task naming fully off the foreground turn, make every naming job session-owned, cancel and bounded-wait it at close, and reject late priced responses.
- Disable unasked reasoning for tiny title/task-name calls, while preserving explicit configured effort suffixes even when the provider catalog is silent.
- Make immediate task cancellation and runner startup share an explicit ownership boundary so only one side finalizes a node.
- Document the slash and paste behavior in the compiled chat manual.

## Why

A live task created from a folded paste was displayed as `paste 1`, and a provider switch appeared stuck. The audit found multiple independent boundary defects: transport scaffolding was accepted as a name, title work remained attached to the foreground lifecycle, hidden reasoning could consume a namer's entire 32-token answer ceiling, task cancellation could race runner handle installation, and several send/lifecycle paths lost or textually over-expanded paste data.

## Validation

- `go test ./internal/session` (2,000+ tests, including deterministic cancellation/name-close races)
- `go test ./internal/provider`
- `go test ./internal/remote ./internal/orchestrate ./internal/manual/...`
- focused changed TUI text-lifecycle matrix, `-count=3`
- focused effort precedence and catalog-silent wire tests, `-count=20`
- `go vet ./...`
- `make build && make size` (50,260,914 bytes; budget 54,600,000)
- real OpenRouter E2E, exact `deepseek/deepseek-v4-flash-0731`: task admitted in 17 ms, immediate cancellation retained async rename and persistence (`narrow path audit`), and a second real worker echoed the slash/quotes/backslash/Unicode/dollar/at-sign literal exactly; $0.003420, 13.56 s

An independent read-only final review found and then verified the catalog-silent effort-precedence edge case; it reports no remaining blockers.


santoshkumarradha:
Reverted in #487 (dev 1ecf8e8d, tag checkpoint/2026-09-02/dev-green-after-revert-483): this merge left internal/tui3 red on TestAMissingTypedWindowsPathNamesOneFileAndNeverSubmits (wslpath_test.go:165, ok on afb0a95c, FAIL on 5e0fa6b9, bisected in clean worktrees). Re-land through the merge queue: rebase onto dev, both CI jobs green (the touched-packages job runs tui3), then message santosh-3c for a slot. Note #448 (aforge-v2-8a) meets this change at setCancel/cancel.go and carries the resolution in its body.

PR #484, opened 2026-09-02
chat: /status --json prints the status note as one JSON object

AbirAbbas:
`/status --json` prints the same ordered status facts as the text note, with string values and absent facts omitted. `/info` and `/context` accept the same exact flag. Output remains a conversation note, so terminal wrapping is not a file export API.

The branch now includes current `dev`. Conflict resolution preserves the separate `/spend` machine ledger, the `/cost` conversation reading, all manual probes, and the new status JSON menu row. The change entry now states the invalidated behavior.

Validation: focused status/command/manual tests passed; full `internal/tui3` (527.516s) and `internal/manual` (1.301s) passed; canonical `make build` passed. Current head CI is still running. This updated branch is held for a secondary review before merge.


santoshkumarradha:
Converted to draft by the coordinating session (santosh-3c): dev is under a WIP freeze with a fixed merge queue. This PR merges through the queue: rebase onto dev, both CI jobs green (the touched-packages job runs the packages it changes), then message santosh-3c for a slot.

santoshkumarradha:
Manual gate (the repo's rule: a person-facing change carries its manual sentence **and** a probe before it goes ready — CLAUDE.md § "THE MANUAL LAW").

The commands.md sentence is good and reaches: on this branch `aforge manual "can I get the status as json"`, `"what does /status --json print"` and `"print the status line facts as json for a script"` all return `commands · /status — everything the status line knows` first. What is missing is the probe that keeps it that way. Please add to the `asked` table in `internal/manual/chat_test.go`, beside the existing `/status` rows:

```go
{"can I get the status as json", "commands"},
{"what does /status --json print", "commands"},
```

and one sentence on the page for the refusal your body describes ("any other spelling falls through to the bare text note"), e.g. under the same `/status` section: "Any other argument is ignored and the text form prints." Verify with `make build && bin/aforge manual "can I get the status as json"` (your page must be in the first four) and `go test ./internal/manual/ -count=1`.

Thank you for building it behind one shared list so the two forms cannot drift — that is exactly the one-source-of-truth law this codebase keeps.

santoshkumarradha:
Held, not merged, and here is exactly why.

`/status --json` emits the object through the note renderer, which re-flows text to the terminal's width. At 78 and at 118 columns the 380-character object wraps inside a string value, and copy mode snapshots rendered rows, so what a person copies is not valid JSON and carries the `· ` lead. A line meant for a script that a script cannot parse is the feature not working.

We looked for an existing lossless path in `internal/tui3` and there is none: the fenced code block and the note block both truncate a long row to one line with `…`; the phone tier re-flows with `↳` marks; there is no horizontally scrolling block, and copy takes the rendered frame. Pretty-printing one key per line into a fence is lossless at 118 columns but still cuts the `crew` row at 78.

So this waits for a lossless render kind (a verbatim row that copy returns whole, whatever the width) rather than gaining a special case. The manual page now states the limit honestly instead of claiming the copy is valid as-is; the probes and the field-name fixes on the page are worth keeping either way. When a verbatim kind exists, route the object through it, add a test that renders at 78 columns and `json.Unmarshal`s the copied text, and this is ready.

santoshkumarradha (changes requested):
Reviewed at `168770371`. The design is right — one `statusItems` behind both forms, hand-assembled so key order is the contract rather than a map's whim, and the emptiness law carried onto the wire deliberately. Green where it can be: `gofmt` clean, `go build ./...`, `go vet`, `go test ./internal/manual/...` ok, `make test-laws` 13/13 ok. Four things before it lands, and the second is a defect the tests pass straight through.

### 1. It no longer applies to `dev`, and the conflict is semantic

`git rebase` onto `dev` stops on the first of the four commits with conflicts in `internal/tui3/commands.go` and `internal/manual/chat/commands.md`. This is not whitespace: dev has since split `/spend` into its own command and **deliberately removed `spend` as a `/cost` alias**, with the comment "The word `spend` used to be an alias here and pointed the one guess a person makes at the wrong reading." This branch still carries the old `alias: []string{"usage", "tokens", "spend"}` and the old `| /cost | /usage, /tokens, /spend | — |` table row.

So please rebase, and resolve **both** in dev's favour — a careless resolution silently reverts that work. Dev also added a `## /search and /spend` section to the same page. (I simulated the post-rebase manual state in a throwaway worktree — your new section, table row and four probes applied onto dev's `commands.md` and `chat_test.go` — and `go test ./internal/manual/...` is `ok 1.229s`, so your heading does not hijack retrieval against dev's newer sections. That part is fine.)

### 2. `--json` emits an empty key `""` whenever two or more conversations are open

```json
{"session":"lab","model":"m","crew":"balanced · …","":"2 open","state":"idle","place":"/t/lab","keys":"/ commands"}
```

The mechanism is entirely pre-existing, but this PR is what puts it on a wire:

- `internal/tui3/statusdeck.go:513` — `deckSegWords` has **no entry for `segOpen`**, so its word is `""`.
- `internal/tui3/statusdeck.go:395` — `add` guards the **value** only (`strings.TrimSpace(value) != ""`), never the label.
- `internal/tui3/render.go:2038` — `add(segOpen, a.openSegment())`, and `openSegment` returns `"2 open"` once `openCount() >= 2`.

In the text form this has always been a dangling unlabelled row, which is cosmetic. As an object key it breaks three claims made in this PR: `commands.md:615` "The labels are the keys", `commands.md:630` "There is no `null`, no `\"\"` and no `0`", and the change entry's "never `\"\"`".

The array's own comment shows the house fix — `segKeeping` had exactly this bug ("the standing side had NO word at all here, so its segment came out of the loop above with an empty label and hung in the value column under nothing") and was cured by giving it the word `watching`. I would do the same: give `segOpen` its word. That fixes the text form's dangling row in the same stroke, and it keeps the fact on the wire, which dropping label-less items would not. If you would rather have `--json` skip label-less segments instead, that is defensible, but then the manual page has to say the object can be missing a fact the text note shows — and I think that is the worse page.

Either way, the guard belongs in `add`, so the next word-less segment cannot do this again. Two of them today would produce duplicate `""` keys, which decoders collapse.

### 3. The contract test's oracle is the thing it is testing

`internal/tui3/statusnote_test.go:334` does `items := a.statusItems()` and compares the object against those items — the same helper `statusJSON` consumes. That proves "the JSON matches `statusItems`", not the PR's claim that it matches what `/status` prints; a header, footer or filter appearing in `statusText` later drifts undetected. (The **order** assertion is real — it token-scans the note produced through `a.slash` — so only the membership half needs strengthening.)

Please compare the key set against the labels parsed out of the rendered text note. That would also have caught item 2: `TestStatusJSONOmitsWhatTheSessionDoesNotHaveRatherThanZeroingIt` checks that values are non-empty strings and never checks a key.

### 4. The PR body names three tests that do not exist

The description promises `TestStatusJSONMatchesTheTextFormInOrder` and `TestStatusWithoutTheFlagStillPrintsText`. Neither string is in the tree. What is actually there — and is good — is `TestStatusJSONCarriesTheSameFactsInTheSameOrder`, `TestStatusJSONOmitsWhatTheSessionDoesNotHaveRatherThanZeroingIt` and `TestStatusAnswersOnlyTheExactJSONFlagAndOtherwisePrintsTheText`. Worth correcting, since the description is what a reader checks the claim against.

### Smaller

- `commands.md:583-586`'s order list, which this PR corrects, is still incomplete: it omits `watching` (`deckSegWords[segKeeping]`) and the `open` row from item 2.
- `statusnote.go:162` writes `"}\n"`, so the JSON note's `entry.text` ends in a newline that no other note carries (`app.go:4995` does no trimming).
- Argument parsing is exactly `strings.Cut` + `TrimSpace` then `rest == "--json"`, which I checked against the manual sentence and it matches: `/status  --json` with two spaces gives JSON; `--JSON`, `--json extra`, `-json`, `--json=true` all fall through to text. The table tests four of those; `--json extra` and `--json=true` are the two worth adding, since a caller who typo'd them silently gets prose.
- An `invalidates` line would be earned here: the command table's `| /status | … | — |` row meant "`/status` takes no argument", and that is a thing a reader believed and no longer should. The rules make the field optional, so this is a judgement call rather than a gate.

### Not yours

`go test ./internal/tui3/` fails two tests on this branch — `TestTheSpendPlaceDrawsTheLedgerItWalkedInOn` and `TestAnEmptySpendWindowKeepsTheControlThatPagesItBack`. Neither is in `.github/known-red.txt` and neither is caused by this change: the PR touches no spendplace file, both pass on a clean dev worktree, and dev has rewritten `spendplace.go`/`spendplace_test.go` by 322/48 lines since this branch's merge-base. They clear on rebase. Prefix budget likewise: this branch reads 47843 where dev reads 47531, and the 312 bytes are base drift — the change itself costs zero.

PR #452, opened 2026-09-02
revert: #440 and #437, the lumped stream that left twelve tui3 tests red

santoshkumarradha:
Dev-never-red: reverts #440 (1de89c08) and #437 (5995a6e7) together, since #440 edits #437's `reveal.go`. #438 (`waitreport.go`) is independent and stays. A separate session re-lands the two with the fix.

## Bisect

Same four-test subset (`TestARoomDrawsAndCollapsesTheNodesThinking|TestAnInterruptedTurnPromotesNothing|TestMarkdownArrivesOnSettle|TestTheThinkingBlockStreamsCollapsesAndExpands`), each in a clean detached worktree, 2–4 s per run:

| commit | PR | internal/tui3 |
|---|---|---|
| 3967244b | #336 | ok |
| babbf033 | #352 | ok |
| **5995a6e7** | **#437** | **FAIL** |
| 09f1b3df | #438 | FAIL |
| 1de89c08 | #440 | FAIL |
| 858cac58 | #390 | FAIL |

santosh-3c reproduced `TestMarkdownArrivesOnSettle` cut at `som` on a clean 858cac58.

## The twelve

TestARoomDrawsAndCollapsesTheNodesThinking, TestAnInterruptedTurnPromotesNothing, TestTheLiveStripKeepsTheNewestTwoThingsAndScrolls, TestAResizeBetweenTheTwoShapesKeepsTheExchange, TestAPersonsLineNeverCutsAStreamedAnswerInTwo, TestATaskRoomWalksItsWholeLifeWithoutAStaleWord, TestARunningRoomShowsTheTurnsBehindTheOneInFlight, TestTheStreamingThoughtShowsOnlyItsLastThreeLines, TestTheWindowFadesInTruecolorAndNotUnderNoColor, TestMarkdownArrivesOnSettle, TestAWatcherDrawsTheTurnAnotherWindowStarted, TestTheThinkingBlockStreamsCollapsesAndExpands.

Failure shapes: `the reply never arrived` (hierarchy_test.go:214) and `a streaming reply is not plain` (tui3_test.go:1120) with the streamed text cut at `som`.

## Verification

On the tip d67fad58, locally, `-count=1 -timeout 15m`:

```
ok  	github.com/Agent-Field/aforge-v2/internal/tui3	490.719s
ok  	github.com/Agent-Field/aforge-v2/internal/manual	0.827s
ok  	github.com/Agent-Field/aforge-v2/internal/session	114.273s
```

[...]

PR #486, opened 2026-09-02
tui3: double space goes home from every place, not only from a conversation

### LINKED ISSUE #457: tui3: double space answers only from the conversation box — from any other tab of the home screen it never goes home
## Observable failure

Double space is sold as the universal door home, and in the conversation it is: an empty box and two spaces and you are on the home tab (`home_test.go`: TestDoubleSpaceInAnEmptyBoxGoesHome, TestDoubleSpaceGoesHomeFromABoxThatShowsNothingButHoldsANewline, TestEveryBoxTheFootCallsEmptyAnswersTheDoubleSpace).

Tab to any other tab of the home screen — tasks, standing, memory, spend, search, settings — and press space twice: nothing goes home. On the typing places the box eats both spaces; on settings space is the activate verb; on standing the box never types a space at all. The gesture the foot advertises stops existing the moment you leave the conversation, which is the place you need it least.

The person's sentence for it: anywhere in home or anywhere, pressing double space should land you on the home tab of the home screen — including from another tab in the home screen itself.

## Where the wire ends

- `home.go:3538` `homeGesture` — fires only on `" "` with `homeDoorOpen()` and an empty box whose previous character is a space.
- `home.go:3571` `homeDoorOpen()` = `canOpen() && !a.at(pageHome)` — on a place page the door condition itself is already true; what is missing is that no key path on a place ever calls the gesture. The conversation router (`input.go`) reaches it only on the conversation surface; the place pages route through their own `placeKey` handling and never arrive.
- `room.go`'s goHome doc already claims the ambition for ←←: "the one gesture that does not care where you are". Double space is advertised in the conversation foot as the same door, and from a tab it is not.

## What the fix must preserve

- Typing is sacred: a single space then a letter must type normally into a place's box; only the double tap on an empty box is a door.
- settings keeps space as its activate verb; the door must not disarm the verb that already owns the key there.
- home itself stays shut: on the home tab double space is nothing — you are already there.

## Acceptance (functional, required)

From every place of the home screen and from the conversation, double space on an empty box lands on the home tab; single space plus a letter still types into a place's box; settings' space still activates; the three existing double-space tests keep passing.

santoshkumarradha:
Double space now opens the home tab from every place, not only from a conversation. Closes #457.

**What was true, and what is true now**

`space space` in an empty box opened home only when a conversation owned the frame. The gesture lived at the bottom of the conversation router, past the rung where a standing place takes the whole keyboard — so a person on the tasks, memory, spend or search place, or on settings, had the two spaces die under their hands while the tab bar sat one walk away. The person's word for what they wanted was `universal`.

The door is now read at the bottom of the place router as well, against the box the standing place types into, with the same guard and the same law as the conversation's:

- `homeGesture` and the new `placeHomeGesture` share one box-law, [app.homeDoorArmed] — one law about emptiness and the caret, two doors in.
- The door loses to every other meaning a space can have where a person is standing: the whole-keyboard layers above it (`composerLayerKey`, `pl.owns`), the router's six classes, and every key a place claims for its own rows.
- On home itself it is a no-op, and two spaces type into home's own filter.
- Typing is sacred: a single space then a letter types normally into any place's box; the door takes only the second of two spaces into a box that still shows nothing.

**The review caught one real bug, and it is fixed here**

The parallel review of this change found the memory card editor: a whole-keyboard surface living *below* the door in `memoryKey`, on the very box the door reads — so a card cleared down to its last space armed the door, and the second space wiped the wording and stood the person on home mid-edit. `placeMemory.owns` now claims the keyboard while the editor is open, the same claim the settings panel already makes for its value editor and model picker, and `TestTheDoorStaysShutOverTheMemoryCardEditor` holds it — the two spaces type into the card and the person stays where they are.

**Probes (all passing on this branch)**

- `TestDoubleSpaceFromEveryTypingPlaceGoesHome` — tasks, memory, spend, search
- `TestDoubleSpaceFromStandingGoesHome` — the shared composer's armed box
- `TestSpaceStaysTheVerbOnSettings` — space keeps its `activate` meaning on the row
- `TestASingleSpaceThenALetterTypesNormallyOnAPlace` — typing is never eaten
- `TestThePlaceDoorIsShutOnHome` — home is a no-op
- `TestTheDoorStaysShutOverTheMemoryCardEditor` — the nested editor keeps its keyboard
- The three pre-existing double-space tests keep passing.

**Manual and changelog**

`keys.md` and `home.md` name the widened reach and the honest exceptions — settings' verb, the memory card editor, standing's shared-composer nuance, home's no-op. The change entry is `docs/changes/unreleased/457-double-space-universal-home.md`.

**Verification**

`go build ./...`, `go vet ./internal/tui3/`, `aforge-changes check`, and the full `internal/tui3` + `internal/manual` suites are green on this branch (rebased onto `dev` at `880a90d`, after #452's revert landed).

## Real-model evidence

`TestTUIE2E/space_in_the_task_room_pages_the_card` — the tmux suite, the real
`bin/aforge` in a real terminal, against a real model. **PASS in 50.6s** (51.3s
wall), for **$0.03** of task spend as the record's own line reports it. The
conversation ran on the suite's pinned `deepseek/deepseek-v4-flash`; the task's
worker ran on `deepseek/deepseek-v4-pro`, which is the crew's low row in the
profile the lane copies from the person's own home. Run at `dfab7a88`, the
commit that carries the fix; `4ebd62ca` adds only this subtest and its needles.

It starts a real `/task solo` from the chat, waits for the project's record to
be written, and reads it back from a second window that has never held the node
— then presses one space into the roster's filter, `enter` into the record, and
the space that used to walk out of it. Both halves were measured:

- the person stayed in the room — `esc back · ↑↓ scroll · m puts it in your message` still on the frame, home's foot absent
- and the record **paged**: `space paged the record exactly as pgdown does: "done · landed moments ago · ran 26s" is off both frames`

The frame is fourteen rows on purpose: at twenty the whole record fit, and the
first passing run could only report that there had been nothing to page.

Three needles joined `internal/e2e/tuiwords_test.go` — the roster's two doors
(`enter open its room`, `enter go inside it`) and the record's own foot — so the
untagged gate keeps them honest.


santoshkumarradha:
Converted to draft by the coordinating session (santosh-3c): dev is under a WIP freeze with a fixed merge queue (48 open PRs draining in order), and #483 from the same road left dev red an hour ago and is being reverted in #487. This PR merges through the queue: rebase onto the revert sha, both CI jobs green, then it takes a slot after the in-flight tui3 PRs (#473 reveal re-land, #337 lens). Message santosh-3c for the slot.

Issue #133, opened 2026-08-31
a shaper that spends its whole window thinking starts the task unshaped, silently


santoshkumarradha:
Found by #125's e2e (see its PR body): on one real run the shaper (`kimi-k3`, `model:high`) burned the entire `TaskShapeWindow` (25s) in reasoning, the stream ended `StreamFailed` at exactly +25.0s, and the task started **unshaped** — the person's raw words became the brief with no sign anything was lost. Pre-existing behavior of the window cap meeting a thinking shaper; #125's new tail makes it visible for the first time (the italic thinking just vanishes and the task starts).

Wanted, in order of preference: (a) a shaper cut off mid-think falls back gracefully — one honest dim line on the started row (`brief kept as you wrote it`), never silence; (b) consider whether the window should scale with the role's effort, which is a cap change and carries its PERF.md line in the same commit; (c) a test pinning whichever is chosen.

Refs #112 / PR #125.

santoshkumarradha:
**Reduced in scope by PR #253 — do not build the mechanism half from this issue.**

This issue's mechanism ask (should the shape window scale, does a thinking shaper get cut) is being answered by #253's waiting-policy redesign. Its ruling separates two kinds of silence: *wire silence* — any delta at all, hidden reasoning included, never heartbeats — feeds the stall clocks; *answer silence* — visible output only — feeds the think-duration model. A shaper that spends its window reasoning is producing hidden deltas the whole time, so under the new rule it is a healthy thinker rather than a silent one, and is no longer cut for it.

What remains here after #253 lands is option (a) alone, a surface fix: **a shaper cut off mid-think must leave one honest dim line on the started row (`brief kept as you wrote it`), never silence.** That is a small tui3 change with no clock semantics in it.

Recommended: hold until #253 merges, then re-scope this to the dim line only.

santoshkumarradha:
Closed in the 2026-09-10 triage sweep — fixed by PR #493 (`chat: a shaper cut mid-think shows a fallback line (#133)`), option (a) of the three this issue asked for. Verified at dev@95a0d0a9b: the line ships as `brief kept as you wrote it` (`internal/tui3/taskcommand.go:50`), is pinned by `internal/session/task_shape_test.go:533`, and has a manual retrieval probe (`internal/manual/chat_test.go:1311`). Options (b) — scaling the shape window — and (c) were not taken; file them fresh if they are still wanted.

PR #493, opened 2026-09-02
chat: a shaper cut mid-think shows a fallback line (#133)

AbirAbbas:
A shaper that spends its whole shape window thinking used to start the task with
the brief as written and nothing on the record — no shape, no line saying why.

session.StartTask now returns a fourth value, the fallback note; remote
forwards it on the task-start wire, and the chat draws it as its own dim line
under the started line, reusing the note slot the stopped-turn line already
uses. Empty is every ordinary start, and the absence law draws nothing for it.

Gates (from a fresh origin/dev): session Shape tests, the full remote suite,
tui3 Task tests, and the families + task-naming e2e lanes (the e2e pair skips
without OPENROUTER_API_KEY).

## Since the first review

- The three committed `.orig` files are gone, in a commit of their own.
- THE MANUAL LAW: the person-facing line `brief kept as you wrote it` was in no
  page, so `internal/manual/chat/tasks.md` now carries it under its own heading
  — what it means, what the task still has, and which of the other shaping
  failures stay silent — with two probes in a person's own words.
- The wire half had no test: `fakeAgent` returned an empty note and
  `TestTaskDoorsRunOnTheEngineAgent` threw the value away, so deleting `Note`
  from the server's receipt broke nothing. The loopback receipt is now checked
  for it.
- `remote.TaskStarted.Note` is additive and `omitempty`; the protocol `Version`
  stays at 10, because an engine that never sets it is read as an ordinary
  start, which is what it is. The change entry says so.

## Real-model evidence

`TestFamilies` (`go test -tags e2e -run '^TestFamilies$' -count=1 -timeout 40m -v ./internal/e2e/`) on **deepseek/deepseek-v4-flash** at tip `669bde09`: **PASS**, 4/4 scenarios, wall **5m48s** (348.16s), cost **$0.0752** — the lane drives the changed `runFamily`/`StartTask` seam against a real model, no skip. Log: `/tmp/claude-1001/-home-santosh-src-aforge-v2/1d906f17-723b-4168-adb6-eec78d0dc175/scratchpad/e2e-493.log`


santoshkumarradha:

Issue #399, opened 2026-09-02
tui3: the test harness sleeps 150ms for every command that never returns, which is 76% of the package's wall clock and past CI's ceiling


santoshkumarradha:
## Problem

The nightly full check at 779ed018 failed with `internal/tui3` timing out:

```
panic: test timed out after 8m0s
	running tests:
		TestTheStripCannotOutliveTheRowItWasOpenedOn (2s)
```

The named test is an innocent bystander. It runs in 3.62s every time, alone or starved,
and its subtests are exactly 0.45s each. The whole tui3 test binary crossed CI's
eight-minute per-binary ceiling: 397s two nights ago, 409s last night, past 480s after
a day of merges. Whichever test is in flight at the cliff gets blamed, so this must
never enter known-red under any test's name.

## Mechanism

The test harness's `runCmd` in `tui3_test.go` runs every command with a 150ms budget and
abandons the losers. Test fakes never feed the surface's wait channels, so every
`waitEvent`, `waitPilot`, `waitRoom`, `waitTask` and every tick longer than 150ms costs
exactly 150ms and leaks one goroutine. Measured over the full package with a counter in
`runCmd`:

| | |
| --- | --- |
| drops | 2415 |
| time spent sleeping | 362s of 478s |
| goroutines leaked by the cliff | 806 of 886 |

Three quarters of tui3's wall clock is the harness sleeping. The budget was documented
as the way to handle commands that never return on their own. It was right at three
hundred tests and is the dominant cost at 2696.

## The law

A harness pays nothing for a command it knows will never return. The set of such
commands is closed and checked, so a new waiter cannot quietly reintroduce a sleep per
call.

## Fix shape

Test files only. The surface code is correct and untouched.

- `runCmd` recognises the known never-returning commands by function name and drops
  them at once, with no goroutine: the app's `wait*` methods and bubbletea's `Tick`.
- The 150ms budget stays as the fallback for anything unrecognised, so nothing real is
  dropped faster than today.
- A structural test in the same file enumerates every `wait*` method on the app, and the
  tick uses, and fails when one is missing from the recognised set. That is what stops
  the list rotting.

The cleaner seam, fakes that own and close their channels, exists and is not done here:
the package has three other lanes in flight tonight and that refactor crosses all of its
test doubles. With the structural test in place it becomes an option rather than a debt.

## Replication

```sh
go test ./internal/tui3/ -count=1 -timeout 8m
```

[...]

PR #499, opened 2026-09-02
ci: the touched packages run on dev, not on pull requests

santoshkumarradha:
## What this changes

The `touched packages` job no longer runs on pull requests. It still runs on every push to `dev` and on manual dispatch — `if: github.event_name != 'pull_request'`. The `check` job (the only required gate) is untouched.

## Why

The job never blocked a merge — it was `continue-on-error`, and only `check` is required — but it still burned up to ~15 minutes per PR, because most changes touch `internal/tui3` (~500s on a free runner) plus `session` (~130s) and `provider` (~50s), all run serially with `-p 1`. The nightly ledger shows its per-PR red signal was not being acted on (the nightly has been red 8 runs in a row), so the cost bought nothing.

The result is still visible where it is actually read: on the merge to `dev`. A red there still means what it always meant.

## What does not change

- Every push to `dev` still runs `touched packages` in full — same job, same `-p 1`, same neutral debt rules.
- The nightly `ci-full` is untouched.
- The required PR gate stays `check` alone, and now answers in ~2.5 minutes instead of sitting at "checks running" for ~15.

Do not merge until CI passes.


santoshkumarradha:

santoshkumarradha:
This does not merge as written. The `if: github.event_name != 'pull_request'` on `touched` removes the job from pull requests entirely; it does not add a dev-push run (the job already ran on every push to `dev`, and still does).

The pull request is the one place a red can be read before it is on the trunk. #437, #439 and #483 each landed a red on `dev` that a full run of the package they touched would have shown on the pull request. Moving that run to the push moves it back to exactly the moment #401's workflow comment says nobody is watching.

The two facts in the body argue the other way. "It never blocked a merge" is because it is `continue-on-error` until it has been green twice, and it has now been green on #401's own tip; the next step is to make it block, not to remove it. "Its red signal was not being acted on" is a page problem, and the job already writes a warning annotation and a step-summary line on failure.

If the cost is the objection, the cheaper shapes that keep the signal on the PR are: run the touched packages with the tui3 suite's fixed pacing once #399 lands (it takes the package to about two minutes), or skip `internal/tui3` alone on PRs by name with the reason written beside it. Either keeps the hole closed.

Issue #96, opened 2026-08-31
internal/standing breaks the Windows build — the shipped .exe has not compiled since 2026-08-20, and the next release tag will fail


santoshkumarradha:
## The problem

`cmd/aforge` does not compile for Windows. It has not since **2026-08-20**.

```
$ GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build ./cmd/aforge
# github.com/Agent-Field/aforge-v2/internal/standing
internal/standing/running.go:126:14: undefined: unix.Kill
internal/standing/running.go:127:43: undefined: unix.EPERM
internal/standing/store.go:352:17: undefined: unix.Flock
internal/standing/store.go:352:44: undefined: unix.LOCK_EX
internal/standing/store.go:355:26: undefined: unix.Flock
internal/standing/store.go:355:53: undefined: unix.LOCK_UN
internal/standing/store.go:373:17: undefined: unix.Flock
internal/standing/store.go:373:44: undefined: unix.LOCK_EX
internal/standing/store.go:373:57: undefined: unix.LOCK_NB
internal/standing/store.go:375:26: undefined: unix.EWOULDBLOCK
```

`internal/standing` imports `golang.org/x/sys/unix` from `store.go`, `running.go` and `tick_test.go` with no build tags. `cmd/aforge` depends on it, so the whole shipped binary is unbuildable on Windows — not just the tests.

## Why it is urgent

`.github/workflows/release-binaries.yml` still publishes `windows/amd64` and `windows/arm64`, with `if-no-files-found: error`. **The next `v*` tag fails at the build step.** `v0.1.0` did ship `aforge-windows-amd64.exe` and `aforge-windows-arm64.exe`, so this is a regression against a platform that has real released artifacts.

## Why nobody noticed

`internal/standing` landed in `1349a458` (2026-08-20), three days after `v0.1.0` (`fde1f1fc`, 2026-08-17). Until now the only cross-platform build ran on `master`, which never received those commits — so no Windows compile has happened since. The new `cross build` job in `.github/workflows/ci-full.yml` (added in #82) caught it on its first run: https://github.com/Agent-Field/aforge-v2/actions/runs/33414121900

## The fix — both replacements already exist in this repo

No new shim is needed; two internal packages already have `_unix.go`/`_windows.go` halves with matching semantics.

### 1. Locking → `internal/filelock`

[...]

PR #501, opened 2026-09-02
session: make cmd/aforge build on windows (#96)

AbirAbbas:
internal/session still reached for unix-only syscalls after #378 moved internal/standing through processgroup and filelock, so the .exe has not compiled since 2026-08-20. Session now spells the same acts through those packages (detached spawn and group signal via processgroup, session-file locking via filelock), and the sid-split session test moved behind its unix build tag.

## Proof boundary, stated plainly

The windows process-group (`CREATE_NEW_PROCESS_GROUP`, `taskkill /T`) and `LockFileEx` behaviour in this PR is **compile-proven only**: both windows cross-builds exit 0 on the tip, and no windows box ran the binary. Nothing in CI runs a windows binary. One consequence is already known and filed as #508: resume reads the session file through a second handle under its own mandatory lock, so the first windows binary will likely fail to resume a non-empty session. That is #378's filelock semantics made reachable by this build fix, not a unix change; unix behaviour is byte-for-byte unchanged here.


santoshkumarradha:

PR #558, opened 2026-09-03
canary: the whole suite runs past a module that cannot import, and an anchor can be remeasured

santoshkumarradha:
Bench-only change (`bench/canary/`), outside the core queue like #505, #509, #517, #519 and #545.

Three of the nine frozen anchors had a whole-suite baseline the judge could not use. `tox-dev/tox` (`--group dev`) and `Human-Agent-Society/reef` (`.[dev]`) both die at collection — one test module imports `argcomplete` (only in the `completion` extra), six import `slime` (not installable) — so pytest stopped in half a second with `passed 0` and the suite column said nothing on either; live doors noticed the red module and pip-installed the missing package themselves, so live suites ran while a regrade (which rebuilds the venv from the recipe) could not. `pypa/virtualenv`'s recorded rung `--group dev` stopped resolving on 2026-09-03 (pip `resolution-too-deep`), so both its cells read `venv: pip install --group dev failed` at 0 s on the 517b91c2 table.

Three changes:

- **The whole suite continues past collection errors.** Both the judge's whole-suite run and `pick.py`'s base measurement pass `--continue-on-collection-errors`, spelled once in each file as `SUITE_FLAGS` with the law THE BASE IS MEASURED EXACTLY THE WAY A CELL IS GRADED. The fix pull request's own tests are not run this way; a collection error there is the grade.
- **`pick.py --remeasure REPO ...`** re-clones a pool entry's base, walks the install ladder from the recorded rung downwards, and rewrites `install`, `base_suite` (or the cap note) and a `measured` date in place, leaving everything else in the entry as it was. Run here for tox, reef and virtualenv; the pool diff is the result.
- **`base ⊘` in the table** where a base collected nothing (`passed 0`, `errors > 0`), read from the cell's own `entry.json`, because failures counted against a base that never ran are not a regression the run caused.

Verified: `py_compile` on the three Python files, `pick.py --help`, the offline `--remeasure` guards, and the live remeasure of the three anchors (output in the PR's first comment).


santoshkumarradha:
remeasure tox-dev/tox: --group dev -> --group dev · 0 passed, 0 failed, 1 errors -> not measured
remeasure Human-Agent-Society/reef: .[dev] -> .[dev] · 0 passed, 0 failed, 6 errors -> 1370 passed, 10 failed, 6 errors
remeasure pypa/virtualenv: --group dev -> --group dev · 452 passed, 9 failed, 0 errors -> 452 passed, 9 failed, 0 errors
rewrote /home/santosh/af-canary/bench/canary/pool.json

PR #589, opened 2026-09-03
task: repository work is always on a branch, and a landing never writes a protected checkout

AbirAbbas:
## What this changes

Two roads put a task's work into the person's live checkout, and both are closed.

**A model's `where` is evidence, not authority, inside a repository.** The chat model set `where: "in place"` on four code tasks in one session; `resolveTaskGround` treated any `where` as the person's own placement and skipped the ground ladder, so the workers edited the shared checkout with no branch. Now a `where` that points into a repository with a commit — `in place`, or a path inside one — is redirected: the work goes on a `task/` branch in the session's own tree, and the door that accepted the placement says so once (`in place was asked for, and <root> is a repository — the work goes on a branch cut from it instead`). Plain folders keep the in-place road, and the one channel that still puts repository work in place is the person's own word on a referred place (`PlaceRef.Mode`). Both placement roads (`resolveTaskGround` and the older `prepareTaskTreeAt`) read one function, `whereInsideRepository`.

**A landing never writes a protected, moved or detached checkout.** A landed branch was merged into whatever branch the root checkout was on; a person sitting on `dev` had their local `dev` fast-forwarded with an `aforge`-authored commit carrying a scratch file. Every worktree cut now records the branch the root was on (`home`, additive on the task record and the standing tree), and `comeHome` keeps the branch — outcome `kept`, worktree unregistered, state as the audit settled it — when the root is on a protected branch (`main`, `master`, `dev`, `develop`, `development`, `staging`, `stage`, `trunk`, `production`, `prod`, `release`, or any remote's default branch), has moved since the cut, or is detached. The person reads one exact sentence naming the branch and the reason. Protection applies only where the root is the person's own repository: family parts, mirrored family trees, legacy task folders and the conversation's owned workspace are exempt, so nothing about families changes.

**A task that depends on kept work starts from it.** On a protected branch, work no longer accumulates in the checkout, so a root task whose `depends_on` names kept work is cut from that branch's tip; several kept dependencies are merged before the worker starts, and a conflict fails the task to start naming every branch.

## Surface and manual

`kept` is a fifth merge word. The settled card, the rail and the room draw `branch kept · task/x`; the structural test that walks the engine's const block covers it. The task manual gains a section answering "why was my task's branch kept", the protected list is held against the engine's constant by a test, and the prompts (`shape.md`, `task.md`, `divide.md`) stop promising that work always lands on your branch.

## Not in this PR

Design nodes and saved subharness runs still run in the session workspace with the full belt; fork hands share their parent's directory; the chat's own inline-edit allowance and the cross-session write lease are untouched; kept `task/*` branches are not swept. Each is a follow-up with its own design.

## Verification

- `go build`, `go vet`, `make fmt-check`, packed corpora, `make test-laws`, and `make test` over `internal/session`, `internal/tui3`, `internal/manual`, `internal/e2e` (untagged) — all green locally.
- Real binary (`make build` on this branch) against a real model (deepseek-v4-flash via OpenRouter), driven through tmux: a scratch repository on `main`, one commit; the chat was asked to hand off "create hello.txt" with `where: in place`. The task record after landing: `groundMode: worktree`, `home: main`, `merge: kept`, `state: done`, and its report carries `its branch task/create-hello-txt-… was kept: your checkout is on main, which aforge never writes to — merge it when you are ready`. `git rev-parse main` is identical before and after, `git status` is clean, no worktree is left registered, and `git show task/…:hello.txt` prints `hi`. The worker also divided: one part failed and its retry merged into the parent's tree (`home: task/…`), which is the exemption for harness-owned trees doing its job.
- A first real-model run, with an acceptance the chat wrote wrongly (`wc -c` = 5 for `hi\n`), still left `main` untouched and four task branches holding the file; that road is the existing incomplete-work keep, not this change.


AbirAbbas:
The red on `touched packages` is dev's own, not this branch's. The same three tests fail on dev's gate for 4985121d (#563) and 4aaac923 (#534): `TestASteerAdoptsAnOldBashAndItsExitArrivesLater`, `TestAStopSteerKillsAnOldBash`, `TestTheBackgroundAfterClockKeepsTheCommandAndTheConversationMovesOn` (see https://github.com/Agent-Field/aforge-v2/actions/runs/33793181921 and https://github.com/Agent-Field/aforge-v2/actions/runs/33794160876). The pull-request checkout is the merge with current dev, which carries those commits; this branch is based on 713945e3, where dev's gate was green, and the three tests pass here with `-count=3`. Everything else on the gate is green: light gate, `internal/tui3`, `internal/manual`, `internal/e2e`.

santoshkumarradha (dismissed):
Reviewed at `c378d2c48` as it stands, since the rebase does not apply. Green on that head: `gofmt` clean on all 21 Go files, `go build ./...`, `go vet`, `ok internal/session 113.402s`, `ok internal/tui3 459.350s`, `ok internal/manual 1.074s`, `ok internal/e2e 0.199s` (the untagged tmux-words gate), `make test-laws` green at 44 files / 14 packages. The prefix budget reads 47995 on this head, but this PR contributes none of it — `widestPage()` weighs only `system.md` + `discipline.md`, and `shape.md`, `task.md` and `divide.md` are separate embeds that are not on that budget at all. The 464-byte gap is dev drift at the merge-base and returns to 47531 after the rebase. Nothing to act on there.

**The road this exists to close is closed, and I went looking for a way through rather than confirming the happy path.** `mergeIntoGround` has exactly one caller — `comeHome` at `task_run.go:6183`, directly after the new guard — and `mergeTaskBranch` is reachable only through it, so the root merge is genuinely single-choke-pointed. `landMirror` runs before the guard but `groundMode` only returns mirror when the root is not a repository, so it cannot lay over one.

So the e2e went at the road you did **not** verify: the `/land` standing-tree road, which is the second consumer of `comeHome` and the only user of the new `StandingTree.Home`. A scratch repo on `dev`, attached as a place, written into by the chat, then `/land now`:

```
BEFORE  git rev-parse dev = 3eac00716e7e03fa00ff1080c5b0db1f753d853c   status clean
AFTER   git rev-parse dev = 3eac00716e7e03fa00ff1080c5b0db1f753d853c   status clean
        git worktree list → only the root; git show chat/proj-b22fb8:hello.txt → hi
```

Identical sha, clean tree, no worktree left registered, work safe on its branch. **The claim holds on that road too.** ~$0.007.

I also probed one predicate that looked weak — `landsInThePersonsRepository` returns false whenever `root != ground` (`task_branch_protection.go:71-74`), and `Agent.Land` builds its tree with `ground: tree.Folder`, `root: tree.Root` — by replicating `Land`'s construction in a scratch test:

```
subdir ground: landsInPersons=false  merge="merged"  dev 90af7304 -> cc6d4d05   ← protected branch written
root ground:   landsInPersons=true   merge="kept"    dev unchanged
```

**It is unreachable today** — `ReferPlace` snaps to the repository root (`places.go:141-144`), as does `groundRoot`, and I confirmed live that `/attach …/proj/notes` prints `folder · /t/a/5/proj`. So this is latent, not a hole. But the guard's safety rests on an invariant enforced in a different file with nothing pinning it. **Please add a test asserting a repo-subdirectory ground is still protected**, so the invariant lives where it is relied on.

Now the things I would fix before this lands. None of them is the safety road; all of them are what ships beside it.

### 1. "Moved since the cut" is not what the code does

`task_branch_protection.go:114` is `t.home != "" && current != t.home`, and `home` is a branch **name** (`currentBranch(root)`). No sha is recorded anywhere — `taskRecord.Home` is a `string`.

So a person who commits, rebases or resets on the *same* branch has not "moved" by this test, and the landing merges into them. That is the long-standing behaviour and this PR does not regress it. What is wrong is that four texts promise otherwise: the PR body, `docs/changes/unreleased/589-…md:8`, and `how-tasks-run.md` all say "has moved since the cut", which every reader takes as "the checkout advanced". Only "you are on a different branch now" is implemented.

Either record the sha in `home` and compare that, or make all four texts say what the code checks. I lean to fixing the texts — whether a moved sha *should* block a landing is a product decision and not one to smuggle in here — but the two must agree, and on a safety claim the text is the thing people will rely on.

### 2. A deliberate keep is reported as a failure

`internal/tui3/landcmd.go:194`:

```go
if landing.Kept() {
    return "could not put it all into " + landing.Name + " · " + landing.Note
}
```

`FolderLanding.Kept()` (`standingtree.go:168`) now returns true for `mergeKept`, so my e2e's successful, deliberate, policy-driven keep printed:

> `could not put it all into proj · its branch chat/proj-b22fb8 was kept: your checkout is on dev, which aforge never writes to — merge it when you are ready`

Nothing "could not" be done. Everything worked exactly as designed. That phrasing was written for a conflict and this is not one, and the `internal/tui3` half of the diff only updates task cards — no surface was updated for the new word on the `/land` road. The sentence after the dot is excellent; the clause before it undoes it.

Related: `Land` runs `dropStandingTree` on the kept path (`standingtree.go:578`) and meta `trees` went to `null`, so the chip and the record are dropped. The person's only route back is the `git merge` the sentence names, which is fine — but the `/land` manual page does not say so.

### 3. Five manual sentences the change makes false

The gates only check that a *name* is mentioned, so all five pass while contradicting the new section 150 lines away in the same file.

1. **`choosing-a-folder.md:136`** — "For a repository the work is committed on a branch of its own and **merged into your checkout**". This is the page the chat answers `/land` questions from, and my e2e proves it false. This is the worst of the five.
2. `how-tasks-run.md:250` — "its work comes home as a merge into your branch."
3. `how-tasks-run.md:699` — "merged home onto your branch when the task lands — that is the road, and it is the only one."
4. `how-tasks-run.md:748` — "What lands on your branch is…"
5. `reading-a-task-page.md:139` — the terminal-word enumeration has no `branch kept`, though `roomStateWord` now returns exactly that on the page that section describes.

### 4. Two prompts promise what the code refuses

`shape.md:51` "A repository always gets a branch" and `task.md:3` "A named path inside a repository still gives you a branch of that repository" are both false in two documented cases: a repository with no commit (`whereInsideRepository` ends `ok && hasCommit(root)`, so an unborn repo takes the in-place road and the person's directory is written directly) and a bare repository. Both are deliberate and commented in the code, and neither is on any manual page. `divide.md` is correct — parts are excluded by `n.parent != 0` and by the `Trees()` exemption.

Also narrow but real: `task.md`'s sentence is false whenever `PlaceRef.Mode == "in place"`, since `taskstands.go:203-212` returns the person's own mode untouched. That is the exemption working as designed; the prompt just states it too absolutely.

### 5. Smaller, none blocking

- The exemption predicate is lexical: `strings.Contains(root, "/.aforge-v3/")` (`task_branch_protection.go:79-95`) exempts a person's own repository at `/work/.aforge-v3/client`. Practically unreachable, and the second marker check is redundant with the first — but it is a safety predicate with no test on its negative side. Worth one.
- Remote-default protection fails open when `refs/remotes/<remote>/HEAD` is absent or stale. Codex called this major; it is not, because the fixed-name loop runs first, so `main`/`dev` and the rest stay protected offline. Only an oddly-named remote default is affected.
- `hideFurrowMarker` (`groundladder.go:567`) appends to the person's `.git/info/exclude` during grounding. Pre-existing and metadata only, but it means "never writes the person's checkout" is not literally true, and it is worth a clause somewhere.
- The depends-on merges are all done with cwd `tree.dir` — the session's worktree — and only a read-only `rev-parse` touches the root. On conflict it aborts and discards. **That half is right.** The cleanup errors are all `_ =`, so a failed abort or worktree removal is silent.
- No test covers same-branch sha movement, an absent remote default, an exemption false positive, or a universe-rung landing on a protected root (the two `home:` lines at `groundladder.go:445/462` have no assertion).

### The rebase

It conflicts against dev `4c08cd3b` in `internal/manual/chat_test.go` only, in two hunks, and both are trivial: the import block (dev wants `regexp`, you want `go/ast`/`go/parser`/`go/token`/`strconv` — union them) and the end of file, where dev's `TestNoChatPageSpeaksTheHarnessesOwnVocabulary` and your `TestC14TheChatManualNamesEveryProtectedBranch` are appended at the same anchor and git cannot tell they are disjoint. I checked dev's new banned-vocabulary gate against every line this PR adds to `internal/manual/chat/`: zero hits. **#608 created no new conflict** — your `task_store.go` changes are disjoint from its `interruptedAt` work and auto-merge.

One consequence to name rather than discover: adding `go/ast` and `go/parser` to `chat_test.go` puts that file on the laws gate, so all seven of its `Test` functions now run there, the big probe table included. That is correct — `TestC14…` reads the engine's const block and is exactly what the gate is for — and it is cheap at ~1.2s. It is just a gate change nobody asked for.

### Where I am unsure, stated rather than smoothed

Whether "moved" ought to mean a sha or a branch name is a product decision I am not making for you. What is unambiguous is that the code and four texts disagree about which one it is.

And I could not construct a live path to the `root != ground` bypass, but I could not prove none exists either — I checked `ReferPlace`, `groundRoot`, `placeStand`, `cutTaskWorktree` and `universeBranch`, not every future caller. Hence the test in the ask above.

santoshkumarradha:
Pushed `be03d30` — all five asks, rebased onto dev `4c08cd3b`. The rebase is the part worth reading first, because it was not the trivial one I told you it would be.

### The rebase merged clean and did not build

The `chat_test.go` conflict was exactly as expected: union the imports, keep both appended tests (dev's `TestNoChatPageSpeaksTheHarnessesOwnVocabulary` and your `TestC14…`, which git could not tell were disjoint). Everything else auto-merged — **and then failed to compile.** Dev has since changed `comeHome` to return a third value:

```
internal/session/task_run.go:6424:22: not enough return values
	have (string, string)
	want (string, string, landingRefusal)
```

Your protected-checkout block returns two. Git had no way to see that; the text did not overlap.

The fix needed a decision rather than a filled-in blank, so here is mine and please check it: the keep path returns **`refusedNothing`**. A policy keep is a landing *honoured*, not refused, and routing it through a refusal would put it on the unsaved-landing road (`task_land_unsaved.go`) and offer to retry it — which is the one thing that must never happen to a checkout aforge will not write. It is commented in place. Your `task_landing_protected_test.go` and `task_depends_kept_test.go` also still called the two-value form and now take the third.

### The five asks

**1. The texts say what the guard checks.** `home` is `currentBranch(root)` and the test is `current != t.home`, so committing or rebasing on the branch you started on is not "moved" and still merges home — as it always did. The PR body, the change entry and two manual pages said "moved since the cut", which reads as "the checkout advanced". They now say which branch you are on. Whether a moved *commit* should keep the branch is a product decision and is filed separately rather than decided here. Your `keptLandingSentence` was already correct, incidentally — it names both branches.

**2. A deliberate keep is not a failure.** `FolderLanding.Kept()` answers two different questions, and `landed()` gave both the conflict phrasing — so my own `/land` e2e, which worked exactly as designed, reported `could not put it all into proj ·`. `KeptByPolicy()` separates them and the policy keep says its own sentence, which already names the branch, the reason and the command that takes the work.

**3. Five manual sentences, `choosing-a-folder.md:136` first** — the page the chat answers `/land` questions from, which still said a repository's work is "merged into your checkout" full stop. It now carries the protected case in the words somebody meets it in, and says plainly that it is not a failure. The others: `how-tasks-run.md`'s three "onto your branch" promises, and `reading-a-task-page.md`'s terminal-word list, which had no `branch kept` while `roomStateWord` returns exactly that.

**4. The prompts state the exceptions.** `whereInsideRepository` ends `ok && hasCommit(root)`, so an unborn repository — and a bare one — takes the in-place road and is written directly. `shape.md` said "A repository always gets a branch"; it now says which repositories, and `task.md` names the `PlaceRef.Mode` case too.

**5. The subdirectory-ground test — and I closed the hole rather than pinning it open.** You asked for a test; I want to flag that I did slightly more, so you can push back.

`landsInThePersonsRepository` answered false the moment `root` and `ground` differed, and `Agent.Land` builds its tree with `ground: tree.Folder`, `root: tree.Root`. Nothing reaches that today — `ReferPlace` and `groundRoot` both snap to the repository root, and I confirmed live that `/attach …/proj/notes` prints `folder · /t/a/5/proj`. But a test that merely asserts the current answer would pin the *bypass*, and this is the guard whose failure costs somebody their working tree. So a ground **within** the root now reads as the person's repository, and `TestC15ARepositorySubdirectoryGroundIsStillProtected` fails on the old comparison:

```
--- FAIL: TestC15ARepositorySubdirectoryGroundIsStillProtected
    a ground inside the person's repository (…/001/notes under …/001) is not read as theirs,
    so a landing there would merge onto their branch with none of this file's protection
```

Every exemption below it is untouched and still decides the borrowed cases; the test also asserts a ground *outside* the root is still not the person's, which is what the old comparison was for.

The change entry gained a line for that and one for `hideFurrowMarker`, which appends to the repository's `.git/info/exclude` while grounding — the one write "never writes your checkout" does not cover.

### Green on the rebased tip

`gofmt` clean, `go build ./...`, `go vet`, `ok internal/session 126.960s`, `ok internal/tui3 460.025s`, `ok internal/manual 1.408s`, `ok internal/e2e 0.340s`, `scripts/laws.sh` 44 files / 14 packages green, `make changelog-check` well formed. **Prefix budget is back to 47531** — the 47995 I reported was merge-base drift in `system.md`, not this change, and the rebase confirms it.

### What I did not decide

Whether "moved" ought to mean a sha or a branch name is yours and the coordinator's; the texts now match the code either way, and changing the code later changes four sentences with it.

And I could not prove no live path to the `root != ground` bypass exists — I checked `ReferPlace`, `groundRoot`, `placeStand`, `cutTaskWorktree` and `universeBranch`, not every future caller. That is why the guard now stands on its own rather than on the snap.

For the record on what I verified rather than took: the root merge really is single-choke-pointed — `mergeIntoGround` has exactly one caller, `comeHome`, directly after your guard, and `mergeTaskBranch` is reachable only through it. And the adversarial run went at the road your verification did not cover, the `/land` standing-tree road: scratch repo on `dev`, attached, written into, `/land now` — `git rev-parse dev` identical at `3eac0071` before and after, tree clean, no worktree left registered, work safe on `chat/proj-b22fb8`. It holds.

PR #599, opened 2026-09-03
checks: a step out of a pasted reproduction is never run as a check when it is written as a command line

### LINKED ISSUE #552: checks: a step out of a pasted reproduction is never run as a check
## Observable failure

An unattended chat session was given an issue pasted verbatim as its ask. The issue's reproduction section read:

```console
$ chmod 000 tox.ini
$ tox -c tox.ini list
```

The check harvest admits any line that begins with `$ ` as a check, so both lines were run against the deliverable tree at the session's start as baseline checks and recorded as failing. They failed only because that repository carries `tox.toml` and no `tox.ini`. On a repository that does carry the file, the first line would have succeeded, made the project's own configuration unreadable for the rest of the run, and passed the tree-moved guard, since a mode change alters neither size nor modification time.

## Law

A check is a command the WORK names, never a step out of the reproduction the person pasted. A line that begins with `$ ` in a pasted issue is a story about how to see the bug, not a promise that running it proves the work.

## Fix shape

Task package, the check harvest (`declaredChecks`, the `$ ` branch): a `$ ` line is admitted only from the work's own account (a node's door, the acceptance the session composed) and never from the ask's prose; equivalently the harvest takes a source kind and refuses the `$ ` branch for the ask. #534 already stops the session-level harvest from reading the ask; this issue is the harvest itself, which the node doors also use. Named tests: a `$ chmod 000 tox.ini` line in an ask yields no check; the same line in a node's own door still does.

## Replication

aforge-v2-14's rig, cell `scratchpad/pr534/20260903T124442Z-d7e968dd9-chat3-513b/cells/tox-dev-tox-4031-chat/` under `/tmp/claude-1001/-home-santosh-src-aforge-v2/e6a3213b-328f-4073-9618-2ade0b09f5a7/`: the head journal's `baseline` row at 12:45:14Z lists both lines as failed checks. Row: https://github.com/Agent-Field/aforge-v2/issues/407#issuecomment-5526166757.

Owner per the coordinator: AbirAbbas. Filed from the #534 read.

AbirAbbas:
## What was wrong

An unattended session was handed a GitHub issue pasted verbatim as its ask. The issue's reproduction section read `$ chmod 000 tox.ini` and `$ tox -c tox.ini list`, and the check harvest admits any line beginning with `$ ` as a command the work has named — so both lines were run against the deliverable tree as baseline checks and written down as failing. They only failed because that repository carries `tox.toml` and no `tox.ini`. On a repository that does carry the file, the first line would have **succeeded**: it would have made the project's own configuration unreadable for the rest of the run, and it would have passed the tree-moved guard, because a mode change alters neither size nor modification time.

A check is a command the work names, and a line that opens with a shell prompt in a pasted issue is a story about how to see the bug, not a promise that running it proves the work.

**The scope, stated plainly: this closes the shell-prompt spelling and only that one.** A backticked span is still read as a check out of either account, deliberately — backticking the command is how a person usually names one, and taking that away would break the ordinary case to fix the dangerous one. So the #552 command itself survives one Markdown step away: `` `chmod 000 tox.ini` `` in the same pasted issue is still harvested and still passes the gate. That asymmetry is worth its own issue rather than a silent widening here.

## What is true now

The harvest takes the account the text came from and reads a prompt line only out of the work's own. A `$ ` line in a brief, or in a done-condition somebody actually wrote, still names a check exactly as it did; the same line in the person's pasted words names none. A backticked span is untouched in both accounts, because backticking the command is how a person usually names a check, and taking that away would break the ordinary case to fix the dangerous one.

Two roads carry the ask into the harvest and both are closed. A node's brief carries the person's request verbatim at the top, so that half is now read as the ask. And where no judge could write a done-condition, the fallback frames the person's own words as one — so a session's acceptance *and* an auto-started node's acceptance can literally be the pasted issue. That frame is now recognised at both doors and the words behind it are read as the ask they are. #534 closed the session's ask and left this second road open.

- **A `$ ` line in the person's own words is not a node's check** — not on the door, not in its offer, refused by the gate. `TestAStepOutOfThePastedReproductionIsNeverANodeCheck`
- **The same `$ ` line in the work's own account is still a check.** Same test, second node.
- **A backticked command in the person's own words is still a check.** Same test, third node.
- **A done-condition that is only the person's words carries no `$ ` step onto a node's door, while a backticked check inside those same words still opens it.** `TestANodesDoneWhenThatIsThePastedAskCarriesNoStepOutOfIt`
- **A session whose done-condition is the pasted ask runs no `$ ` step out of it, and still runs the backticked check in it.** `TestASessionWhoseDoneWhenIsThePastedAskRunsNoStepOutOfIt`
- **A settled unit of work contributes no pasted step to the session's own checks, while its own check still arrives.** `TestASettledNodeNeverContributesAStepOutOfThePastedAsk`
- **A done-condition somebody wrote keeps its `$ ` lines.** `TestAStepOutOfThePastedReproductionIsNeverASessionCheck`, which came with #534 and passes unchanged.
- **Nothing else about the harvest moves** — the command-shape and runnable-here readings, the receipts, the reading commands, the critical floor and the per-source cap are all as they were. The existing `internal/session` suite carries that.

Both manual pages that said otherwise are corrected in the same change: the task page listed a shell-prompt line as a check wherever it appeared, and the unattended page still said the session runs what "your request" backticks, which stopped being true when #534 took the ask out of that harvest. The task page also now carries the question somebody actually asks — why a command out of a pasted issue was run.

## Gates

| Gate | Result |
| --- | --- |
| `go build ./...`, `go vet ./...`, `make fmt-check` | pass |
| `go generate ./internal/manual`, packed-corpus test, clean `git diff` | pass |
| `go test ./internal/manual/` and `-run Manual` over `internal/tui3` + `internal/session` | pass |
| `make test-laws` | pass, 43 files in 14 packages |
| touched packages — `make test` over `./internal/session` | pass |

Re-measured during review on a box that has `ffmpeg` and `ffprobe`: clean `dev` and this branch both read `the fixed prefix is 47531 bytes (~11882 tokens): prompt 20712 + tools 26819` and both PASS, 469 under the 48000 budget. An earlier draft of this section reported that test as a pre-existing red at 48129 bytes on any machine with ffmpeg; that is no longer true of current `dev`, and this change contributes nothing to the prefix either way.

## What was not verified

No real-model run. The change is in the harvest that decides which commands exist, which the tests above exercise directly; nothing about it needs a live model, and the tmux e2e suite does not cover the check harvest.

Closes #552


santoshkumarradha (dismissed):
Reviewed at `69197a573`, rebased onto dev (clean, no conflicts). Green on the rebased tip: `gofmt` clean on all six Go files, `go build ./...`, `go vet ./internal/session/...`, `go test -count=1 -timeout 15m ./internal/session/` → `ok 126.225s`, `go test ./internal/manual/...` → `ok 1.272s`, `-run Manual` over `internal/tui3` and `internal/session` both ok, `make test-laws` 44 files / 14 packages all ok, `make changelog-check` well formed, prefix budget unchanged. No failing tests.

**The fix works, and I have it end to end against a real model with a control.** `aforge do` turns out not to reach this code at all — it runs the plan engine over `internal/store`, and `cmd/aforge/do.go` never sets `TaskAudit`, so `task_run.go:4133` short-circuits before the harvest. The door is `session.Agent.StartTask`, which is what `/task` calls. Driving that with `cfg.TaskAudit = true`, against a seeded repo, with an ask that is a pasted GitHub issue containing `$ touch …599-PASTED-STEP-RAN`, `$ ./greet.sh` and a backticked `` `./check.sh` ``, the checker's own door listing reads:

on this branch —
```
THE CHECKS THIS WORK NAMES OR RAN, which are the only verification commands your bash will run:
  the check <TREE>/check.sh — run it as `<TREE>/check.sh` or `sh <TREE>/check.sh` or `./check.sh`
```

on clean dev, same ask, same model —
```
  the check <TREE>/check.sh — run it as ... or `./check.sh`
  touch /tmp/af-abir/599-PASTED-STEP-RAN
  the check <TREE>/greet.sh — run it as ... or `./greet.sh`
```

and on dev the checker went on to **run** one of them: `audit · bash ./greet.sh`. Both pasted prompt lines are gone from the fixed door, the backticked one survives in both. That is exactly the claim, and it is worth noting the marker file was a weak proxy — dev happened to stop before reaching `touch` — while the door listing separates cleanly and deterministically. Three things I would fix before this lands. None is a design objection.

### 1. The gate fails open on its zero value

`checkSource`'s zero is `checksFromWork` (`task_checks.go:493-496`), and the enforcement is `if from == checksFromWork` (`:545`). Both production callers set the field explicitly today — `task_checks.go:276` and `principal_audit.go:112` — and I checked there is no third; `task_divide_scope.go:203` does its own reading and never reaches the harvest. So this is latent, not live.

But it is the wrong way round for a safety gate. A future caller that forgets the field harvests prompt lines out of whatever text it was handed, which is precisely the failure this change exists to prevent, and it fails silently and permissively. Please make the restrictive value the zero — `checksFromAsk` first in the `iota`, or an explicit unset zero that neither branch accepts — so that forgetting it costs a missing check rather than a run `chmod`. It is a two-line change and both call sites already pass explicitly, so nothing else moves. Happy to push it myself if you would rather not respin.

### 2. "never" is not true, and the #552 command itself survives the other spelling

The title says a pasted step is **never** run as a check, and the body says "A check is a command the work names, never a step out of the reproduction the person pasted." Backticks are admitted regardless of source (`task_checks.go:539`), and `runnableHere` admits anything whose first word is on `$PATH`. So a pasted issue that writes `` `chmod 000 tox.ini` `` instead of `$ chmod 000 tox.ini` still gets harvested and still passes the gate. I reproduced it:

```
checks=[chmod 000 tox.ini rm -rf build]
allowed=[chmod 000 tox.ini rm -rf build git diff git log ...]
GATE ALLOWS backticked chmod out of the pasted ask
```

The change entry is honest about this — "A backticked span is unchanged in both accounts" — and your reasoning for keeping it is sound: backticking is how a person names a check, and taking it away breaks the ordinary case. I am not asking you to change the behaviour. I am asking the title and the body to stop saying "never", and to say plainly that the #552 command is one Markdown spelling away from surviving. Somebody will read this page in six months as "that hole is closed"; half of it is.

Worth a follow-up issue in its own right — the asymmetry is real: #534 established that the person's ask is not the work's promise, and that argument does not stop at the `$ `.

### 3. The `ffmpeg` red in the Gates table does not reproduce

The body says `TestTheFixedPrefixStaysUnderItsBudget` fails at 48129 bytes on a machine with ffmpeg, on a clean checkout of dev. On this box, with `/usr/bin/ffmpeg` and `/usr/bin/ffprobe` both installed:

```
clean dev:   the fixed prefix is 47531 bytes (~11882 tokens): prompt 20712 + tools 26819   --- PASS
this branch: the fixed prefix is 47531 bytes (~11882 tokens): prompt 20712 + tools 26819   --- PASS
```

Byte-identical, 469 under the budget, green both sides. Whatever was true when you measured it is not true of current dev, so that row and its paragraph should come out — they currently tell a reader that a full local run trips over something it does not.

### Smaller, not blocking

The ask-detection is an exact prefix test against the 170-character frame `routeAcceptance` writes (`principal_audit.go:147`, `route_judge.go:1104`). Codex called both a false positive and a false negative on it; I traced every writer and neither is reachable — `Steward.setAcceptance` only `TrimSpace`s, the node path is raw, restore is verbatim, and `checkTexts` correctly tests the raw `n.spec.acceptance` rather than the composed one — so no live path can prepend to the frame. Fine as it stands; worth a comment saying the prefix test is load-bearing, since a future composer that prepends anything would break it silently.

One new `## ` heading — `## Why did it run chmod or reproduction steps from the issue I pasted` in `how-tasks-run.md` — and it displaces nothing: `go test ./internal/manual/...` is green and so are the tui3/session name gates. `starting-aforge.md` is a body edit only. Both pages now say which account a prompt line is read from, which was the other half of the job.

Also: the four new tests are load-bearing. I reverted the gate semantically in a scratch copy and all four fail; reverting `acceptanceIsAsk` alone independently fails the two ask-fallback tests, so the second road is separately pinned rather than riding on the first.

santoshkumarradha:
Pushed `f64e9d1` with ask 1 and the prefix-test comment, and edited the title and description for asks 2 and 3, so this is ready to land without another round-trip from you. What changed and why:

**The gate's zero value.** `checkSource`'s zero was `checksFromWork`, and the enforcement is `if from == checksFromWork` — so a caller that forgot to say whose account a text came from got the reading that harvests prompt lines out of it. Nothing is live today: both callers set it explicitly, and I checked there is no third (`task_divide_scope.go:203` does its own reading and never reaches the harvest). But it is backwards for this particular gate, whose entire reason for existing is that the free answer used to be "the work's" — which is how a pasted `$ chmod 000 tox.ini` became a promise. `checksFromAsk` is the zero now. Forgetting costs a run one check it could have made, which is recoverable; the other order costs the gate.

`TestAForgottenAccountReadsAsThePersonsAndNotTheWorks` pins it, and I checked it fails on the old ordering rather than passing on nothing:

```
--- FAIL: TestAForgottenAccountReadsAsThePersonsAndNotTheWorks
    task_checks_test.go:835: the zero checkSource is not checksFromAsk, so a caller
    that forgets whose account this is harvests the person's pasted commands
```

It also asserts the backticked span still survives the zero account, so the deliberate exception `declaredChecks` states stays pinned rather than becoming collateral.

**`acceptanceIsAsk` gets the comment it was owed.** I traced every writer of an acceptance before deciding this was fine as written — `Steward.setAcceptance` only trims, the node path is verbatim, restore is verbatim, and `checkTexts` correctly tests the raw acceptance rather than the composed one — so no live path can prepend to the frame and the prefix test holds. But it holds *because* of that, and nothing said so. A future composer that puts anything in front of the frame turns it silently false and puts the pasted reproduction back on the door, so the comment now says that in as many words.

**The title and the description.** The title said a pasted step is never run as a check; it now says never when it is written as a command line. The body carries a short scope paragraph in its own right: this closes the shell-prompt spelling and only that one, a backticked span is still read from either account deliberately, and the #552 command itself survives one Markdown step away as `` `chmod 000 tox.ini` ``. Your change entry was already honest about this — the title and the opening were not, and the title is what somebody reads in six months as "that hole is closed".

I reproduced the residual before writing it down, for the record:

```
checks=[chmod 000 tox.ini rm -rf build]
GATE ALLOWS backticked chmod out of the pasted ask
```

It is being filed as its own issue rather than widened here. Your reasoning for keeping backticks is right: backticking is how a person names a check, and taking it away breaks the ordinary case to fix the dangerous one.

**The `ffmpeg` row.** Re-measured on a box with `ffmpeg` and `ffprobe` installed: clean `dev` and this branch both read `the fixed prefix is 47531 bytes (~11882 tokens): prompt 20712 + tools 26819` and both PASS, 469 under budget. The "pre-existing red at 48129" is no longer true of current `dev`, so that row and its paragraph are replaced with the re-measurement rather than deleted — the fact that it was once true and now is not is worth keeping.

**What I did not touch.** The fix itself, the four tests, and both manual pages. They are right. Worth saying explicitly that the tests are load-bearing, since that is the thing reviews usually cannot tell: reverting the gate semantically fails all four, and reverting `acceptanceIsAsk` alone independently fails the two ask-fallback tests, so the second road is pinned separately rather than riding on the first.

Green on the branch: `go build ./...`, `gofmt` clean, `go test -count=1 -timeout 15m ./internal/session/` → `ok 123.298s`, `scripts/laws.sh` 14/14 ok, prefix budget unchanged.

One thing found on the way that is not yours and is being filed separately: `aforge do` never sets `TaskAudit`, so an unattended errand runs with no check harvest at all — `chatv3.go:1284` sets it from `config.TaskAuditEnabledAt`, `do.go` has no equivalent, and `task_run.go:4133` short-circuits. That is also why the first attempt to verify this change through `aforge do` proved nothing; the door is `session.Agent.StartTask`.

PR #604, opened 2026-09-03
docs: the headless contract names the ending #593 added, and the manual holds its retrieval

### LINKED ISSUE #514: do: a delivery the gate could not judge is never ok; it is its own outcome, said in words
## Observable failure

A headless run's repair leaf finished its work; both calls the delivery gate made to judge it were refused by the provider. The run printed "the delivery gate did not judge task-2-x1 — the gate could not be reached; delivering unjudged", then "✓" on the node, and the door ended ok (exit 0) at 554 s with one gate round. The rig read that row as a clean pass and it was compared against runs that had been judged. Read from the store, the run had one gate row, a failing one from the round before; nothing ever judged what was delivered.

## Law

An unjudged delivery is never "ok". It is its own outcome, said in words on the door, in the record, and in the rig's column: "delivered without a check: the gate could not be reached". The yolo posture decides what to do when the gate cannot be reached, retry the gate within the wall or deliver as unjudged, and says which it did; a person reading the door's last line and its exit status can tell an unjudged delivery from a judged one without opening the store.

## Fix shape

At the seam where the gate's unreachability is caught and the leaf is let through (the "delivering unjudged" note), the delivery gets a gate row of its own kind (unreachable, with the reason) rather than none, the settlement reads that row as neither pass nor fail, the door's last line and exit status say so (a code distinct from ok and from partial, or partial with the reason named), and `--json` carries it as a field. Under the yolo posture, one retry of the gate before the wall, then the unjudged outcome. Named tests: a scripted gate that cannot be reached yields the unjudged outcome, not ok; the retry happens once; the rig's parse of the door line maps it to its own column.

## Replication

aforge-v2-14's rig, anchor 1 (reef-145, do door), stack tip 48e920cc: cell `scratchpad/pr461/20260902T233928Z-48e920ccd-a1do-stack/cells/Human-Agent-Society-reef-145-do/` under `/tmp/claude-1001/-home-santosh-src-aforge-v2/e6a3213b-328f-4073-9618-2ade0b09f5a7/`; the two gate 404s at 19:48:52 in home/logs/calls.jsonl, the "delivering unjudged" note in do.err, one delivery_gate row (seq 143, fail) in store/graph.db and none for the repair leaf. Row: https://github.com/Agent-Field/aforge-v2/issues/407#issuecomment-5518115148 (recorded as ok at the time; corrected in #464's body).

Second replication, same cell, stack tip 0dae4d73 sample 4: `scratchpad/pr461/20260903T013517Z-0dae4d733-a1do-stack2d/cells/Human-Agent-Society-reef-145-do/`; the leaf delivered at 4m41s, the single gate call got the 404 in 58 ms, do.err line 17 is the "delivering unjudged" note, the store holds zero delivery_gate rows, and the door printed ok at 281 s. Row: https://github.com/Agent-Field/aforge-v2/issues/407#issuecomment-5519093184.

Second on the post-drain list after #513, beside #456.

AbirAbbas:
## What was wrong

`aforge do` reports `stop: "unchecked"` and an `unjudged` field on a run whose final
delivery check could not be reached — the ending #593 added when it closed #514.
`docs/HEADLESS.md` is the document a harness is written against, and it did not know.
Its exit table gave `2` as `incomplete` alone, its `stop` row listed eight words where
the binary now speaks nine, and its field table had no row for `unjudged`. Somebody
writing a wrapper from that page would branch on a vocabulary the binary had already
outgrown, and would meet a word the page says does not exist.

The chat manual's own page does carry the ending, inside the section about what `$?`
means. What it did not have was a probe: nothing in `internal/manual/chat_test.go`
held a question to it, and a page whose retrieval nothing holds is one the running
chat talks over the top of — the same shape as the lanes page in #453.

## What is true now

- `docs/HEADLESS.md`'s exit table names `unchecked` beside `incomplete` on rung 2 and
  says what the difference is: the work was delivered and nothing has vouched for it,
  so read it — it may be perfectly good.
- Its `stop` row lists `unchecked`, and its field table has a row for `unjudged`
  saying what it holds and that it appears on exactly the runs nothing checked, so its
  presence is itself the answer to "was this checked?".
- Three probes hold the retrieval, in the words somebody meets the ending in: the
  stderr line they have just read, the word in `--json`, and the exit code they are
  staring at with a good-looking answer above it. All three already land on
  `running-from-the-terminal`; no page was moved and no existing probe was touched.

## The contract, and what proves each line

- **The documented headless contract names every `stop` word the binary emits.** —
  `TestHeadlessDocumentsTheLadderAndTheEnvelopeItActuallyHas` (`cmd/aforge`) reads
  `docs/HEADLESS.md`; the tables it reads now carry `unchecked` and `unjudged`.
- **A question about the unchecked ending reaches the page that answers it.** —
  three new rows in `TestTheChatManualAnswersTheQuestionsPeopleAsk`
  (`internal/manual`), each asserted against the retrieved page.
- **No page changed to make a probe land, and no existing probe was weakened.** —
  the diff touches no Markdown under `internal/manual/chat/`.

## Gates

| gate | result |
| --- | --- |
| `go generate ./internal/manual`, `go test -tags=aforge_packed_manual ./internal/manual`, `git diff --exit-code` | pass |
| `go test ./internal/manual/` and `go test -run Manual ./internal/tui3/ ./internal/session/` | pass |
| `make test-laws` | pass |
| touched packages (`./internal/manual`) | pass |
| `./cmd/aforge` in full — it is what reads `docs/HEADLESS.md`, and CI's touched-package set does not include it for a change with no Go file in it | pass, 96.8s |

## What I could not verify

Nothing here needs a model or a network. The behaviour these pages describe is #593's
and was verified there; this change only makes the documents agree with it.

## Note on this branch

This branch was cut to fix #514 and its engine work is superseded: #593 landed the same
fix while it was in flight, with a different and in places better shape (the retry's
30-second wall margin and its error-class filter). That work has been dropped rather
than rebased — the issue is closed and stays closed by #593. What is left here is the
residue #593 did not carry: the machine-facing document and the retrieval gate.


santoshkumarradha (dismissed):
Reviewed on a detached worktree at the head (`31f6007`), rebased onto `dev` `cd502ff1`. It **applies clean** and every gate is green there: `go build ./...`, `go vet ./internal/manual/... ./cmd/aforge/...`, `go test ./internal/manual/` (the three new probes pass), `TestHeadlessDocumentsTheLadderAndTheEnvelopeItActuallyHas`, `scripts/laws.sh` (44 files, 14 packages, all ok), `make changelog-check`, and the prefix budget unchanged at 47531/48000 — 469 under, as on dev. No file under `internal/manual/chat/` is touched, so no new `## ` heading and no BM25 hazard; the change entry is well formed and its three `invalidates` lines are each accurate as written.

The work is right. Three asks before it lands, and the first is in the same table this change edits.

### 1. The exit-3 row is still wrong, in the table this PR corrects

`docs/HEADLESS.md:166`:

```
| `3` | `price`, `deadline` | A limit you set stopped it — the wall (`--timeout`), or a plan price that crossed the consent threshold with no `--yes-spend`. |
```

`exitLadder`'s exit-3 rung is `Stops: []stopReason{stopBudget, stopTurnCap, stopDeadline, stopPrice}` (`cmd/aforge/envelope.go:152-158`), and its own `Meaning` names all four — "the wall, the token budget, the turn cap, the price". So the `stop` column omits `budget` and `turn-cap`, and the Means cell omits the token budget and the turn cap.

This matters here rather than as a separate ticket because it makes the PR's own contract line false in its own file. The description says "**The documented headless contract names every `stop` word the binary emits**", and the change entry's first `invalidates` line rests on the same claim. As it stands the page names eight of ten. It is also a disagreement between two documents that both claim to be authoritative: the chat manual's row already reads correctly — "the wall, the token budget, the turn cap, the price" (`internal/manual/chat/running-from-the-terminal.md:44`).

Please fix both cells of that row.

### 2. `no-progress` is emitted and appears nowhere on the page

`StopNoProgress StopReason = "no-progress"` (`internal/exec/noprogress.go:70`) is set on a leaf the guard terminates (`internal/exec/linear.go:1249`), and it reaches `execStop`'s `default:` branch (`cmd/aforge/exec.go:322`), so the word survives into `stop` and lands on exit `2` exactly like `overrun`. `docs/HEADLESS.md` never mentions it: the pass-through list at `:451` reads `empty`, `overrun`, `promote`, `paused`, `cancelled`, and the second at `:490` reads `cancelled`, `paused`, `promote`, `split`, `empty`, `overrun`. Neither has it. (The enumerating comment at `cmd/aforge/exec.go:322-323` omits it too, which is presumably where both lists came from.)

Pre-existing, and I would normally send it back as its own issue — but it is one word in two lists on the page this PR exists to reconcile with the binary, and leaving it makes the same claim false a second way.

### 3. Nothing gates the claim, so this drift recurs unseen

`TestHeadlessDocumentsTheLadderAndTheEnvelopeItActuallyHas` (`cmd/aforge/vocabulary_test.go:740`) reads `docs/HEADLESS.md` and asserts ten hard-coded quoted field names, three required phrases, three remembered-stale strings and five retired flag spellings. It never iterates `exitLadder` and never compares against the `stopReason` constants. That is why asks 1 and 2 were able to happen at all — and this PR does not add `unchecked` or `unjudged` to any of those lists either, so the exact rot it is repairing is ungated the moment it lands.

Minimum: add `"unchecked"` and `"unjudged"` to the required strings, so this change holds its own ground.

Better, and about ten lines: loop the rungs — for each `exitLadder` entry assert its code and every word in its `Stops` appears in the exit table, and assert every `stopReason` constant appears somewhere on the page. That turns a table somebody has to remember to update into one the build updates for them, and it is the difference between fixing this instance and fixing the class.

### Not blocking

The body paragraph "the two documents a person and a harness read about it now agree with the binary" is broader than what lands; asks 1 and 2 are exactly the gap. Worth softening or, better, making true.

The three new probes carry no coupling to the `docs/HEADLESS.md` half of the change — they are a retrieval gate over text that landed with #593. That is a legitimate thing to add and the entry says so plainly; noting it only so the record is straight about what they prove. They resolve to `running-from-the-terminal` and pass despite the competing heading in `adaptive-runs.md:918`.

santoshkumarradha:
I pushed the three asks as one commit (`1f0cd33`) rather than leave them sitting on a review, since the coordinator wants this landed today. What is in it, and why each:

**1. `docs/HEADLESS.md:166`, the exit-3 row.** It now reads `` `budget`, `turn-cap`, `deadline`, `price` ``, and the Means cell names the token budget and the turn cap beside the wall and the price — which is what `exitLadder`'s own `Meaning` string has said all along. Without this the description's contract line, "the documented headless contract names every `stop` word the binary emits", was false in the table this change edits.

**2. `no-progress`, in both pass-through lists** (`:451` and `:490`). `StopNoProgress` (`internal/exec/noprogress.go:70`) is set at `internal/exec/linear.go:1249` and survives `execStop`'s `default:` branch (`cmd/aforge/exec.go:322`), so it lands on exit `2` under its own name exactly like `overrun` — and the page never mentioned it. I also added `split` to the first list, which was missing it while the second list had it; same defect, same sentence.

**3. The gate now reads the ladder instead of remembering it.** This is the part worth arguing about, so here is the argument. `TestHeadlessDocumentsTheLadderAndTheEnvelopeItActuallyHas` was ten quoted field names, three required phrases, three stale strings and five retired flags — every one of them typed there by somebody who had just noticed a drift. That is a gate that catches the drift it was written for and no other, which is precisely how the exit-3 row stayed wrong through every green run, and this change was adding `unchecked` and `unjudged` to a page with nothing holding them either.

So it now does two things instead:
- reads each row out of the "one ladder" section and compares it against `exitLadder` itself — every rung's code must have a row, and every `stopReason` in its `Stops` must be named in that row. It reads that one section deliberately, so `exec`'s narrower restatement further down (which correctly omits `price`) is not held to the shared table's standard.
- sweeps the `stopReason` and `exec.StopReason` const blocks with `go/ast` and requires every declared word to appear somewhere on the page. That is what would have caught `no-progress` the day it was added, and it is why an ending added next month is on this gate next month without anybody remembering to type it here.

`unjudged` gets a named check of its own, since it is a field rather than a stop word and the sweep does not reach it.

Reverting the three doc fixes fails the gate by name on all three:

```
vocabulary_test.go:804: docs/HEADLESS.md's exit-3 row does not name `budget`, which exitLadder says produces that rung:
      | `3` | `price`, `deadline` | A limit you set stopped it — the wall (`--timeout`), ...
vocabulary_test.go:804: docs/HEADLESS.md's exit-3 row does not name `turn-cap`, which exitLadder says produces that rung:
      | `3` | `price`, `deadline` | ...
vocabulary_test.go:822: docs/HEADLESS.md never names the `no-progress` ending, which ../../internal/exec/noprogress.go declares
```

The change entry gained two `invalidates` lines covering both halves.

Green on the branch: `go build ./...`, `go vet ./cmd/aforge/`, `go test ./cmd/aforge/` ok 47.8s, `scripts/laws.sh` 14/14 ok (`vocabulary_test.go` already imports `go/ast`, so this is on the laws gate at no new cost), `make changelog-check` well formed, prefix budget unchanged at 47531/48000.

One thing I did **not** touch, for the record: `cmd/aforge/exec.go:322`'s comment enumerates the pass-through endings and omits `no-progress` too — presumably where both doc lists came from. It is a comment, it is out of this change's scope, and the new sweep makes the page correct regardless.

Nothing else in the original review stands; the three new probes and the change entry were right as they were. Dismissing my changes-requested now.

PR #605, opened 2026-09-03
chat: an interrupt from outside reaches Close, so nothing the session owned is orphaned

### LINKED ISSUE #471: chat: an external interrupt never reaches Close, so every task and job the session owned is orphaned
## Problem

An external interrupt of the chat door never reaches `Agent.Close`. The chat path has
no `signal.Notify` (only `chatv3_at.go`, `exec.go` and `run.go` install one), so a
`kill -INT`, a terminal hangup, or a service manager's stop takes the default
disposition and the process dies with every `defer` skipped, `Close` included.

The TUI's own quit keys go through `Close`, and after #448 that stops every running
task and refuses late starts. The interrupt road bypasses all of it: every node, child
agent and background job the session owned is orphaned, and a task mid-call keeps
spending until its own process notices the parent is gone, if it ever does.

Found while writing #448's real-model line, which had to use `--once` returning from
its loop to reach `Close`, because an interrupt could not.

## The law

Every door a person can leave through reaches `Close`. A signal is a door.

## Fix shape

The chat door installs the same `signal.Notify` the other doors have, for `SIGINT`,
`SIGTERM` and `SIGHUP`, and routes it to the same quit the TUI's keys use so `Close`
runs with its graces. A second signal during the grace forces exit, so a person is
never trapped behind a stop that hangs.

## Replication

```sh
AFORGE_HOME=$(mktemp -d) bin/aforge chat --once "start a task that takes a minute" &
sleep 20; kill -INT $!
```

The task's child session under that home keeps writing after the parent is gone.

## Acceptance

1. `kill -INT` on a chat process with a task running leaves the task settled in the
   checkpoint and nothing written under the home after the process exits.
2. A second interrupt during the grace exits at once.
3. The manual's quitting section says an interrupt stops work the same way a quit does.

AbirAbbas:
## What was wrong

Kill an aforge chat from outside — `kill -INT`, `kill -TERM`, a closed terminal
window, a service manager's stop — and the process could die with every `defer`
skipped, `Agent.Close` among them. After #448 the quit keys stop every running
task and refuse late starts; the signal road reached none of it, so every node,
child agent and background job the session owned was orphaned, and a task
mid-call kept spending until its own process noticed the parent was gone. Three
holes made it up. The terminal's own hangup (`SIGHUP`) was caught by nothing in
the tree at all. The surface read one signal off its channel and then stopped
reading while the channel stayed registered, so a person watching a slow close
could press `ctrl+c` from another terminal all day and be answered by nothing.
And the headless door — `aforge chat --once`, the issue's own replication —
watched no signal whatsoever.

## What is true now

The set that means "somebody outside this process asked it to stop" is written
down once, in a new `internal/leave`, with the two-signal rule beside it: the
first signal takes the ordinary leaving road, and a second one arriving while
that road is still going ends the process at once on the shell's own number for
it (128 + the signal). The second watch is armed **before** the first signal's
leaving work starts, because that work blocks — arming it afterwards is exactly
the shape that swallowed everything sent during a close taking its time. A door
that owns the terminal gets one bounded chance to hand it back before the forced
exit, and a hand-back that hangs cannot hold it.

Both chat doors read that one road. `internal/tui3`'s `forwardSignals` keeps its
own account of why the surface catches signals at all and delegates the set and
the rule; `runChatV3Once` installs the road before it opens the session, cancels
the turn on the first signal, and leaves through the `agent.Close` it already
deferred — which, by defer order, runs before the road stands down, so a second
signal stays live for the whole of a close that can take the turn's grace plus
two job rounds.

| Behaviour | Proven by |
| --- | --- |
| `SIGINT` reaches the leaving work | `TestASignalReachesTheLeavingWork` (`internal/leave`) |
| `SIGTERM` reaches the same road | same test, second case |
| `SIGHUP` reaches the same road — a closed window is a door now | same test, third case |
| The headless `--once` door leaves through `Close`, installed before the session opens | `TestTheHeadlessChatDoorLeavesThroughClose` (`cmd/aforge`, structural) |
| A second signal ends it at once, on 128 + the signal, without waiting for the first leaving | `TestASecondSignalEndsItWhileTheFirstIsStillLeaving` (130 and 143) |
| The terminal is handed back before that exit, and a hand-back that blocks cannot hold it | `TestTheTerminalIsHandedBackBeforeTheForcedExit`, both halves |
| One source of truth: no door names a signal of its own | `TestTheSurfaceNamesNoSignalOfItsOwn` (`internal/tui3`, structural) + the `cmd/aforge` law above |
| The road stands down when the door closes | `TestTheLeavingRoadStandsDownWhenItIsStopped` |
| The manual says an interrupt stops work the way a quit does | three new probes in `internal/manual`'s retrieval table |
| The turn interrupt (`esc`, #358) and the two-press `ctrl+c` rule are untouched | the full `internal/tui3` and `cmd/aforge` suites |

That a closed session actually stops its tasks and settles its checkpoint is
#448's law and is proven by its own tests (`TestCloseCutsEveryNodeAndWaitsForIt`
and the rest of `task_close_test.go`); what is new here is only that a signal
reaches it.

The manual's quitting section said in as many words that closing the terminal
window is not a quit aforge sees. It is one now, so that sentence is gone —
along with the section's length, which had grown past what one retrieval can
carry whole and was why the new questions could not reach it. Four
self-contained sections, the same facts.

## Gates

Run from a worktree on this branch, rebased onto `dev`:

- `go build ./...`, `go vet ./...`, `make fmt-check` — pass
- `GOOS=windows go build ./...` — pass (`SIGHUP` is defined and never delivered
  there; no build-tagged file was needed)
- `go generate ./internal/manual`, `go test -tags=aforge_packed_manual
  ./internal/manual`, `git diff --exit-code` — pass, the packed corpora agree
  with their folders
- `go test ./internal/manual/`, `go test -run Manual ./internal/tui3/
  ./internal/session/` — pass
- `make test-laws` — pass, 45 law files in 14 packages
- touched packages, `-count=1 -p 1`: `./cmd/aforge ./internal/leave
  ./internal/manual ./internal/tui3` — pass
- `go test -race -count=3 ./internal/leave/` and `-count=5` without race — pass

## What is not verified here

- **No real-model end-to-end run.** The tagged e2e suite needs a key this
  machine does not have set, so the "task settled in the checkpoint after a
  `kill -INT`" line is proven by composition — the signal reaching the leaving
  work, the leaving work being `Close`, and #448's own tests for what `Close`
  does — rather than by one run of the whole road.
- **A bare `aforge engine` serving one pipe is deliberately untouched.** A torn
  pipe on a non-persistent engine already reaches `Close` through
  `server.leave()`, and `aforge engine --daemon` already answers SIGINT and
  SIGTERM in `internal/enginehost`; a signal to a pipe engine whose pipe is
  still open remains a hole, on the far side of a different door, with close
  semantics that turn on `Session.persistent`. It wants its own change.
- One local red, reproduced on a pristine checkout of the base commit and so not
  this change: `internal/tui3`'s `TestADropTypedSlowlyStillLands` fails when
  `TMPDIR` is a path with a real file at a prefix of it, because the test types
  a path one character at a time and attaches whatever the first character
  matches. It is a property of the box, not of the tree, and the package is
  green with any other temporary directory.

Closes #471


santoshkumarradha (dismissed):
Reviewed at `2d6b9fd99`, rebased onto dev (clean, five commits, zero conflicts). Green on the rebased tip: `gofmt` clean on all seven Go files, `go build ./...`, `go vet` on the four packages, touched packages all pass — `leave 0.186s`, `tui3 459.536s`, `cmd/aforge 46.886s`, `manual 1.229s` — `go test -race ./internal/leave/` clean on the concurrency seam, `make test-laws` 46 files / 14 packages all ok, prefix budget unmoved at 47531/48000.

I checked the three "what was wrong" claims against dev rather than taking them: `grep -rn SIGHUP --include='*.go'` on dev returns zero hits; dev's `forwardSignals` really does take one value and return with `signal.Notify` still registered; and dev's `runChatV3Once` has no signal call anywhere in it. All three hold.

**And the road works end to end.** No tagged e2e touches signals at all — `grep -rni 'sigint|sigterm|sighup|syscall.Kill|signal\.' internal/e2e/` is empty — so I ran a real `chat --once` on `deepseek/deepseek-v4-flash` from the worktree binary and sent it SIGHUP 43 seconds in, mid-stream:

```
EXITSTATUS=0
{"ts":"2026-09-03T20:40:35.034-04:00","tag":"turn","model":"deepseek/deepseek-v4-flash",
 "served":"DeepInfra","status":200,"ms":42616,"ttft_ms":1069,"error":"decode stream: context canceled"}
```

`ttft_ms:1069` says the provider stream was genuinely live, `decode stream: context canceled` says the SIGHUP reached the new `leaving()` closure's cancel, and exit 0 says the deferred `agent.Close` ran to completion. On dev that same signal takes the default disposition and the shell reports 129. The transcript and `meta.json` were flushed and closed, and no `presence.json` was left behind.

The race argument in `leave.go` also holds, and for a better reason than the entry gives. "The second watch is armed before the leaving work starts" is loose — creating a goroutine is not scheduling it — but nothing can be lost regardless, because `signals := make(chan os.Signal, 2)` (`leave.go:37`) is registered before either watcher and `os/signal` sends non-blockingly into that buffer. It is the buffered channel, not the ordering, that makes it safe. Worth saying in the comment, since the ordering claim is the one a later reader would try to preserve.

Two asks, and the first is inside the boundary this change draws for itself.

### 1. `runHostOnce` is a chat door with no leaving road at all

`cmd/aforge/chatv3_host.go:1110`:

```go
// runHostOnce is --once over a connection: one message, the reply on stdout,
// everything else on stderr. It is [runChatV3Once] with a remote agent, and it
// is deliberately the same shape — a probe that compares stdout with the
// sentence it asked for must not have to know which machine answered.
func runHostOnce(agent *remote.Agent, text string) error {
	events, err := agent.Submit(context.Background(), text)
```

A hardcoded `context.Background()`, no `leave.On`, no cancel. The change entry says "`internal/leave` is the one statement of it **for the chat doors**", and this is a chat door — its own comment says it is deliberately the same shape as the door you just fixed. After this change the two shapes have diverged, and `--host --once` orphans exactly the way `--once` did before.

The engine hole you wrote down and left is a fair call — it is a different door with a different owner. This one is not: it is the same function with a remote agent. Either give it the road (it looks like the same three lines) or name it in the entry beside the engine, so the next person does not read "the chat doors" as covering it.

Related, and the reason I would not just leave this as a note: `keys.md:483-485` now tells a person that a real signal or a closed window leaves "through the same clean exit: draft written, session closed, status 0", without qualifying the door. That sentence is false for `--host --once` today. The manual law is the one that bites here — a page that is confidently wrong is worse than one that is silent.

### 2. One assertion in the new `cmd/aforge` test proves nothing

`cmd/aforge/leaveroad_test.go:74-76` asserts that `defer agent.Close()` is present — but that line is unchanged context in this diff; it is already on dev. The test's other two assertions (`leave.On` is present, and it comes before `openV3Agent`) do fail on revert and are the real content. Please drop the Close assertion or make it say something this change is responsible for, so the test's name means what it claims.

### Smaller, none blocking

- The forced-exit path skips real cleanup by design, and it is more than the terminal: `chatv3.go:111`'s `defer trace.Announce`, `:374`'s `defer proc.closeAll()` — which is `agent.Close`, `session.FlushUsage`, `recall.Close` and `Memory.Close` (`chatv3_process.go:310-331`) — and `chatv3_surface.go:37-38`'s `closeWire`, plus the `chat.log` close. The entry states the trade honestly; the title's unqualified "nothing the session owned is orphaned" does not. A person who signals twice has asked for this, so I would not change the behaviour — just do not let the headline outrun it.
- After `leaving()` returns, the second watcher stays armed until the caller's deferred `stop()`. In tui3 that window runs from `app.quit()` finishing to `p.Run()` returning (`tui3.go:1048`), and a second signal landing in it exits 130 on a process that had already closed cleanly. The CAS makes it exclusive, so there is no double-close — it is a cosmetic wrong exit code in a narrow window.
- The tui3 law forbids the identifiers and syntactically-direct `signal.Notify`/`NotifyContext`, so an aliased import, a dot import or a function value gets past it. Fine as a ratchet against the accident; worth knowing it is not a wall. (For the record, a build-tagged file *is* scanned — `parser.ParseDir` ignores build constraints — so that particular worry does not apply.)
- The entry's list of doors that spell their own signal set omits `cmd/harness-design/main.go:186`.

### Confirmed clean

Both new `## ` headings in `keys.md` carry high-traffic tokens (*terminal*, *closed*, *running*, *picker*, *kill*) and are exactly the BM25-hijack shape — but the risk did not materialise: the whole probe table passes, including the neighbours I would have expected to lose ("does it keep working when I close the terminal", "does it run when my terminal is closed", "the window closed before a second", "kill a server I started"). The deleted sentence is genuinely gone — `grep -rn "not a quit"` over the corpus returns nothing, and the three surviving "terminal window" hits are all consistent with the new behaviour. The 300ms debounce survives under "Your unsent draft is kept", as claimed.

Shared-box safety, which I checked because signal tests can be dangerous here: every signal in `leave_test.go:177-180` is `syscall.Kill(os.Getpid(), …)` against its own test binary, nothing signals a pid it did not create, and `TestTheLeavingRoadStandsDownWhenItIsStopped` installs a SIGHUP guard before standing the road down so the deliberately-unhandled signal cannot kill the runner.

santoshkumarradha:
Pushed `8f4cb25` with both asks, plus one thing the verification run turned up that neither of us had.

**`runHostOnce` is on the road now.** It takes the same `leave.On` closure `runChatV3Once` takes, so the signal set still lives in one place. Measured over a real ssh pipe, HUP'd mid-stream:

| | exit | stderr | `presence.json` |
| --- | --- | --- | --- |
| with the fix | **0** | empty | none |
| without | **129** | `Hangup` | left behind |

and the control's call log is the part worth reading: the surface died at 21:09:59 and the far turn ran on to 21:11:08 — **sixty-nine seconds of spending with nobody attached, and the session never closed.** That is the orphaning your change is about, on the one door that had been left off it.

**But the cancel alone is not the road on this door, and that is the new thing.** `remote.Agent.Submit`'s own comment says it plainly — "THE CONTEXT BOUNDS THE CALL AND NOT THE TURN" — and names `remote.Agent.Interrupt` as the surface's cancel, "a frame of its own", which travels. With only the `cancel()` copied over from `runChatV3Once`, the SIGHUP does not stop the turn: the process kept printing for another minute after the signal and then exited 0. What it bought was that the process survives the signal at all, so the loop reaches `agent.Close()` and the far session is closed instead of orphaned — which is exactly the `presence.json` difference above, and is worth more than a fast exit.

Still, a person pressing ^C on `--host --once` waiting out the whole turn is not what `keys.md` now promises them. So the closure calls `agent.Interrupt()` beside the `cancel()`: Interrupt is what reaches the turn, and the cancel stays because it is what unblocks this side when the link is already gone. The comment and the change entry carry the measurement, because a `cancel()` that provably does nothing on the door it is written for should not sit there unremarked.

**The no-op assertion is replaced by a law that fails on revert.** `TestEveryHeadlessChatDoorTakesTheOneLeavingRoad` walks both headless doors and refuses one with no `leave.On` or one submitting under a context nothing can cancel. On a clean revert of the `runHostOnce` half:

```
--- FAIL: TestEveryHeadlessChatDoorTakesTheOneLeavingRoad
    runHostOnce does not take the shared leaving road, so a signal orphans whatever it is holding
    runHostOnce submits under an uncancellable context at chatv3_host.go:1119:9
```

It forgives a `context.Background()` that is being wrapped by `WithCancel`/`WithTimeout`/`WithDeadline`, since that is how both doors legitimately get their context.

**`internal/leave`'s comment now credits the right thing.** Your reasoning is sound and the code is correct, but the sentence attributes the safety to arming order, and starting a goroutine is not running it — a second signal can arrive before the watcher is scheduled. It is lost by nothing because `signals` is buffered and registered with `signal.Notify` before either watcher exists, and `os/signal` delivers into that buffer without blocking. I kept the ordering paragraph and added which of the two is load-bearing, because the ordering is the thing a later reader would try to preserve while rearranging.

The entry also gains `cmd/harness-design/main.go`, which was missing from the list of doors that spell their own set.

Green on the branch: `gofmt` clean, `go build ./...`, `go vet`, and all four touched packages — `cmd/aforge 46.142s`, `internal/leave 0.188s`, `internal/tui3 460.446s`, `internal/manual 1.232s` — plus `scripts/laws.sh` 14/14, `make changelog-check` well formed, prefix budget unchanged.

**Two traps from the verification, in case they cost you an hour later.** The control run *falsely passed* at first: `setsid nohup` sets SIGHUP to `SIG_IGN`, a Go child inherits it, and `signal.Notify` in a fixed build overrides that ignore while an unfixed build never sees the signal — so `nohup` silently inverts a SIGHUP test. And `presence.json` is not evidence you can collect late: the far engine host tidies it when it idles out at around ninety seconds, so the check has to be immediate.

**What I did not touch, and what stands.** The engine hole is still yours to have written down rather than fixed, and I agree with that call — different door, different owner. The forced-exit path still skips `trace.Announce`, `proc.closeAll` and `closeWire` by design, which the entry states honestly; only the title's unqualified "nothing the session owned is orphaned" outruns it, and I left the title alone since a person who signals twice has asked for that. And both new `keys.md` headings are retrieval magnets that displaced nothing — the whole probe table passes, including the four neighbours I expected to lose.

PR #606, opened 2026-09-03
plan: a finished leaf's ending is journaled, so a restart restores it as done

### LINKED ISSUE #502: plan: a finished leaf's outcome is written to the in-memory node and never journaled, so a restart restores the plan as planned
A leaf's measured ending is written onto its plan node and never journaled, so
the persisted graph is the plan **as planned** and a restart restores none of
what happened.

Both doors do it, in the same shape:

| Door | Writes | Journals |
| --- | --- | --- |
| `cmd/aforge/chat.go` `recordOutcome` | `Turns`, `Tokens`, `Cost`, `Stop`, `Verdict`, `Artifacts`, `Result`, `Checked`, `Calibration` | **no** |
| `internal/exec/schedule.go` (~700) | `Turns`, `Tokens`, `Cost`, `Stop`, `Verdict`, `Artifacts`, `Result`, `Checked` | **no** — the file contains no `RecordPlanGraph` call at all |

`plans.journal` fires in exactly three places in `cmd/aforge/chat.go`: plan
registration (~3962) and two revision paths (~4255, ~4290). None of them is a
node finishing. So `store.RecordPlanGraph` only ever serialises a graph whose
nodes carry what they were *given*, never what they *returned* — unless a
revision edit happens to land afterwards and sweep the fields in as a side
effect, which is not a guarantee anything relies on.

`internal/store/planjournal.go` says the journal exists so a job's remainder can
be revised after a restart. The remainder is decided from node results, and the
node results are the fields above.

## Replication

Any finished `aforge do` run with `--keep`. Using the run from #496's
acceptance (`deepseek/deepseek-v4-flash`, a two-file change, three nodes, tests
green, `settled: true`):

```sh
STORE=<the store path the run printed>

# 169 events, and not one of them carries a `checked` key
sqlite3 "$STORE/graph.db" "select count(*) from events;"                        # 169
sqlite3 "$STORE/graph.db" \
  "select count(*) from events where payload like '%\"checked\"%';"             # 0
```

The control that makes this a persistence defect rather than a missing value:
in that same run, `Outcome.Account` **was** populated — two derived patch files
sit beside the database with the real diff in them:

```sh
ls "$STORE/graph-scratch/.aforge/trace/"*.patch                                 # 2 files
```

So `Account.Summary()` had something to return, `node.Checked` was set in
memory, and the store still holds nothing. The same is true of `Result`, `Cost`,
`Turns`, `Tokens`, `Artifacts` and `Verdict` on the persisted node.

## Guessed cause

`recordOutcome` predates the plan journal and nobody added the write when the
journal arrived. The headless scheduler never had one.

## Acceptance

1. A run is interrupted after at least one leaf has finished, and restarted from
   the same store: the restored graph's finished node carries its `checked`,
   `result` and `cost` — not the blank node it was planned as.
2. Both doors, not one: `cmd/aforge/chat.go` and `internal/exec/schedule.go`
   reach whatever seam does the writing, and a structural test keeps it that way
   rather than a comment asking the next door to remember.
3. The write is best-effort in the sense the surrounding code already means:
   losing it costs the remainder's accuracy and never the run.
4. It does not journal the whole graph once per node completion if that is
   measurably expensive on a wide fan-out — say what it costs, in `PERF.md`, if
   a bound is introduced.

## Not a duplicate of #492

#492 restored the *writer* of `exec.Outcome.Account`, which is why
`node.Checked` now has a value at all. This is the seam one step later: the
value never reaches disk. #496's acceptance line proves the account through the
patch file precisely because it could not prove it through the node.

AbirAbbas:
A leaf's measured ending was written onto its plan node in memory and never
reached the journal, so `store.PlanGraph` only ever held the plan **as planned**.
Reopen a job after a restart and every finished step came back blank: no
`checked` clause, no cost, no turns, no tokens, no stop, no verdict, no
artifacts. The revision pass that decides a job's remainder reads exactly those
fields — `Node.Checked` exists so it can tell "this step changed files and its
suite came back green" from "this step has a title" — so after a restart it
reasoned as though no leaf had ever checked itself, and went on proposing
children to re-run tests that had already passed. The `plan` tool had the same
hole from the other side: it prints a step's turns off the journaled document,
and that column was always empty because the number never got written down.

Two doors wrote the same nine fields, in the same shape, and neither journalled:
`cmd/aforge/chat.go`'s `recordOutcome` and `internal/exec/schedule.go`'s `apply`
— the second contained no `RecordPlanGraph` call at all. `plans.journal` fired
only at plan registration and on a revision that edited something; a node
finishing was not one of them.

## What is true now

There is **one seam**. `exec.Settle` writes a leaf's ending onto its plan node —
the nine fields and the done-or-failed word both doors spelled separately — and
then hands the document to the journal. The chat door passes a journal that
persists the plan under the namespace its nodes were minted under, so a restart
restores the job as it happened. The headless scheduler reaches the same seam
and passes no journal, because `aforge plan run` mints its nodes under no store
namespace and persists its finished document to the file `--output` names; what
it gets from the seam is that its ending can no longer drift from the chat
door's — and it gains `Calibration`, which it never wrote, so a headless run's
recalibration now carries what the worker said about its own fit.

The store still owns what happened. A rehydrated plan re-derives each node's
state, result and failure from the durable rows exactly as before, so a journal
that says "done" cannot overrule a row that says another process has claimed
that node.

The write is best-effort in the sense the surrounding code already means: a
journal the store refuses is logged and costs the remainder's accuracy, never
the run.

No bound was introduced, so `PERF.md` is unchanged — the issue's fourth
acceptance line asks for a line there only if one is. What it costs, measured on
an eight-core box under load: journalling a 60-node plan once per landed leaf is
60 appends of a 104 KB document, 2.7 seconds of sqlite for the whole run; a
20-node plan is 25 milliseconds for the whole run. That is spread across a run
whose every leaf takes minutes, and it lands on the same cadence the revision
pass already journals on.

- **A finished leaf comes back finished.** A new registry built over the same
  store restores the node with its `checked`, `result`, `cost`, `turns`,
  `tokens`, `stop`, `verdict`, `artifacts` and `calibration` —
  `TestAFinishedLeafComesBackFinishedFromTheJournal`. It fails on `dev`.
- **The ending survives with no durable rows at all** —
  `TestAJobWithNoDurableRowsKeepsWhatTheJournalCarried`.
- **The durable row still overrules the journal**: a row completed with a
  different summary wins on `result` while the journal's measurements stand, and
  a row another worker has claimed restores as running rather than as the
  journal's done — `TestTheDurableRowStillOverrulesTheJournaledEnding`.
- **The ending leaves the process.** The issue's replication was a count of
  stored events carrying a `checked` key, and it was zero; the newest
  `plan_graph` payload now carries the leaf's checked clause and its cost —
  `TestASettledEndingReachesTheStoreAsAnEvent`.
- **One seam writes an ending, and both doors reach it.** No file outside
  `internal/exec/settle.go` assigns a plan node's `Checked`, and `recordOutcome`
  and `apply` each call the seam —
  `TestOnlyTheSettlingSeamWritesALeafsEnding` and
  `TestBothDoorsSettleThroughTheSeam`, both on the laws gate.
- **The seam writes the whole ending and then journals** —
  `TestSettleWritesTheWholeEndingAndThenJournals`; the done-or-failed rule is
  `TestSettleWithNoJournalStillSettlesTheNode`.
- **Losing the journal never costs the run.** A store that refuses the write and
  a registry with no store at all both leave the node settled —
  `TestAJournalTheStoreRefusesDoesNotCostTheEnding`,
  `TestARegistryWithNoStoreStillSettlesTheNode`.
- **A requeue is still not an ending thrown away**, and the headless door's
  failure sentence and events are unchanged — the existing scheduler suites
  (`internal/exec/schedule_test.go`, `verdict_test.go`, `exec_test.go`,
  `jobs_test.go`) hold that, and were run untouched.

## Gates

| Gate | Result |
| --- | --- |
| packed corpora build from their folders | pass |
| the manual law (`./internal/manual`, `-run Manual` on tui3 and session) | pass |
| `make test-laws` | pass — 44 law files in 14 packages |
| touched packages `./cmd/aforge ./internal/exec ./internal/head` | pass |
| `go run ./cmd/aforge-changes check` | pass |

## What I could not verify

The end-to-end replication in the issue — an interrupted `aforge do --keep` run
restarted from the same store — needs a real model and was not run; the restart
is proved at the registry seam instead, over a real store, by the tests above.
No manual page changed: this adds no command, key, tool, default, limit or
refusal, and no page stated the claim that has been falsified. The one place
that did state it was a code comment above `head.renderPlan`, and it is
corrected here.

Closes #502


santoshkumarradha (commented):
Reviewed at `04f8a5eaa`, rebased onto dev `cd502ff1` (clean, no conflicts — dev's three intervening commits touch nothing this PR does). Green on the rebased tip: `gofmt -l` clean on all seven Go files, `go build ./...`, `go vet` on the three packages, `go test -count=1 -timeout 15m ./internal/exec/ ./internal/head/ ./cmd/aforge/` → `ok internal/exec 22.958s`, `ok internal/head 21.142s`, `ok cmd/aforge 47.388s`, `make test-laws` 45 files / 14 packages all ok, prefix budget unchanged at 47531/48000. Change entry present and conforming.

**The claim holds end to end.** I ran the errand twice on a real model (`deepseek/deepseek-v4-flash`), once on this branch and once on clean dev as a control, through `aforge do --db` so the store outlives the process, then read the journal back in a **separate process** through `Store.PlanGraphFor` + `plan.Load` — the same read `jobPlans.hydrate` does. That read-back is the restart.

This branch, landed leaf `task-2`:

```
state = "done"   checked = "1 file changed, +0 +0 lines"   stop = "done"
cost = 0.0004823608   turns = 2   tokens = 9409   verdict = "unverified_success"
artifacts = [haiku.txt]
```

Clean dev, same errand, also finished green and wrote its file:

```
state = "pending"   checked = ""   result = ""   cost = 0
turns = 0   tokens = 0   stop = ""   verdict = ""   artifacts = []
```

On disk this branch writes three `plan_graph` events — seq 10 at plan time with no measurement keys, then seq 30 and 37 carrying `artifacts checked cost result state stop tokens turns verdict`; dev writes exactly one. And `head.renderPlan` over the two real stores differs by exactly the segment you said it would: `| 2 turns |` is present here and absent on dev. Total spend for both runs, about $0.0032.

So: the consolidation is real, both doors reach the one seam, and the durable-row precedence you claim is not just asserted — `syncPlanState` overwrites `State` from `planStateOf(stored.Status)` unconditionally (`chat.go:3984`) and `TestTheDurableRowStillOverrulesTheJournaledEnding` exercises it directly. Good change.

Four things for the record. None of them blocks it in my view, and I am saying so to whoever lands it, but the first is new information you do not have.

### 1. It journals twice per landing

Seq 30 and seq 37 above are identical payloads for one settled leaf: both the leaf runner and the landing path reach `recordOutcome`. Append-only and newest-wins, so it is harmless to correctness — but it is two whole-document store writes per settled leaf where the change entry's cost argument assumes one, which doubles the figure in the description. Worth a look; if the second is load-bearing, the comment should say why.

### 2. The write is O(N²) in a shape PERF.md names

`internal/store/planjournal.go:56` appends the WHOLE `PlanGraph`. An N-node plan now writes N documents of N nodes each — quadratic bytes into a never-pruned `events` table, about 6 MB per sixty-node job, and eight per-leaf writes rather than four with #1 above. PERF.md's letter is satisfied (no documented number changed), but `PERF.md:1227` names this exact shape: *"a per-call cost linear in the transcript is a per-run cost quadratic in the run."* And the "2.7 seconds of sqlite" measurement is wall-clock, which `PERF.md:9` — "gate on work, never on time" — declines as a measure.

The in-house alternative already exists two functions away: `EventNodeBrief` / `RecordNodeBrief` (`planjournal.go:98-136`) is *"one row per briefed node, written beside the blob rather than in place of it."* A `node_settled` event carrying the nine fields would be that idiom, would drop the quadratic term, and would close item 4 below at the same time. I would take this as a follow-up rather than hold the fix hostage to it — the bug being fixed is worse than the cost — but it should be written down somewhere other than a reviewer's head.

### 3. The "one place" claim outruns what the law enforces

`internal/exec/settle_test.go:58` filters `selector.Sel.Name != "Checked"` and `token.ASSIGN` only. So a future door that writes `Cost`, `Turns` or `Verdict` without touching `Checked` — or writes any of them in a composite literal — leaves the law green, while the change entry says "`exec.Settle` is the one place that writes one". Either widen the law to the nine field names (and composite literals), or soften the sentence. Widening is the better answer and is a few lines, since the field set is already spelled once in `settle.go`.

### 4. `journalSettled` has a silent drop path

`cmd/aforge/chat.go:4233` no-ops when `entry.graph != graph`. If `takeIfRoot` has already retired the job, `j.get(prefix)` rehydrates a new graph pointer and the settling leaf's ending is dropped with nothing logged. Benign where it happens today, but the change entry says the best-effort write "is logged and costs the remainder's accuracy, never the run" — this path is the one that is not logged. One line.

Also worth knowing, both pre-existing and neither yours: `head.renderPlan` still prints `$0.00` for a sub-cent leaf (from `store.Impact`), so the journaled `cost = 0.00048` is invisible in the row it now populates; and two of the new tests — `TestAJournalTheStoreRefusesDoesNotCostTheEnding` and `TestARegistryWithNoStoreStillSettlesTheNode` — assert only in-memory fields that the old `recordOutcome` already wrote, so they pass against pre-PR code. They are fine as guard tests; they just are not regression proof, and the other six are.

Issue #572, opened 2026-09-03
handover: a reservation counts what it could not check and names none of it


santoshkumarradha:
## Observable failure

A handover's reservation **counts** the behaviours it could not check and **names none of them**, so nothing in it can be acted on. The sentence is built to introduce a list and the list is empty — it ends on a colon.

`aforge do` on #520, second hand-off, in its deliverable:

> I'm handing this over with a reservation — a review found this still missing: 16 behaviours the request states have no check that exercises them. Nothing in this project's own verification would fail if each of these were absent or wrong, so nothing that has been run says whether the work does them:

`aforge do` on #515, third hand-off, on the person-facing stream at 45m0s:

> partial — gate: 23 behaviours the request states have no check that exercises them. Nothing in this project's own verification would fail if each of these were absent or wrong, so nothing that has been run says whether the work does them: (not repaired: nothing further was started)

Both trail off after `them:`. A person reading either one learns a number and cannot act without redoing the analysis that produced it — which is the whole of what the reservation was for. Sixteen unchecked behaviours out of a request is either fine or fatal depending entirely on **which** sixteen, and that is the one thing the line does not say.

## The product already has the right shape, one lane over

The same fact, rendered by the gate on the stream during a different run (#510, first hand-off, 41m17s):

> no check exercises — Count timed-out tool calls that are identical (same command text, same timeout) within a round, in the leaf harness. — and 27 more

Fact, the quoted thing, the count. That line is actionable: it names the behaviour verbatim and says how many more there are. The two above are the same underlying finding with the naming removed. So this is not a feature to invent — it is one rendering to bring into line with another that already exists.

## Law

**A RESERVATION NAMES WHAT IT COVERS.** A handover that says work was not checked names at least one of the behaviours it means, verbatim, and counts the rest — the shape the stream's `no check exercises` line already uses. A bare count is not a reservation; it is the shape of one.

## A second failure of the same kind, at the other end

A reservation that is never uttered at all. On #515 the worker met a genuine, load-bearing constraint — the caller's refusal reaches `repair` as a plain `error` and `why()` (`internal/shaped/shaped.go:601`) flattens it to its first line, so there is **no typed rule identity at that seam to branch on**. Fixing the issue properly needs one created at the caller's contract, which is wider than the issue scoped. Instead of saying so, the worker matched the refusal's English with `strings.Contains(lower, "quote")` and returned a hardcoded field name, and its handover said nothing about the constraint.

[...]

PR #607, opened 2026-09-03
exec: a leaf's done carries the commands it ran, or says no check was run (#550)

### LINKED ISSUE #550: leaf: done without the check the brief named, and no line saying none was run
## Observable failure

In the `aforge-as-worker` trial (`bench/worker-trial/REPORT.md`, branch `trial/aforge-as-worker` @ `89786a4d`, section "2026-09-03 · #515"), run 1's second leaf `task-2-x1` wrote two new tests into `internal/shaped/shaped_test.go`, confirmed they failed, implemented its fix across `shaped.go` and `prose.go`, then ran **one** command — `go build ./...` — and landed. The node was recorded **done**. Its summary, complete apart from the file list, is:

> Now build and test

The brief named the check twice. The request's own checklist (`acceptance`, seq 33) carries the point:

```json
{"behaviour":"go build ./... and go test ./internal/shaped/ must be run when done",
 "quote":"When you are done, run `go build ./...` and `go test ./internal/shaped/`"}
```

and the leaf's runtime contract (its turn-0 note) ends:

> Finish once: go build ./..., go test ./internal/shaped/. … Final message names tests, the exact -run command, the changelog filename, and what each run returned.

`go test` was never run. After the last edit there is exactly **one** tool call in the whole transcript, and it is the build. The next leaf ran the suite in its third turn and got:

```
exit: exit status 1
Go test: 1 passed, 1 failed in 1 packages
shaped (1 passed, 1 failed)
  [FAIL] TestReaskPromptNamesBrokenRule
     shaped_test.go:524: the object-shaped answer must be repaired
```

A test the same leaf had written, in the package it had just edited. `go test ./internal/shaped/` runs in **0.005 s** on this tree (REPORT.md, "Where it ended").

The harness had already measured that red and said nothing. The leaf's closing photograph is in the store:

```
seq 197 | 11:24:22.152Z | task-2-x1 | verification |
  {"when":"on the finished tree","scope":"touched packages (3 files)","named":4558,"red":1,…}
```

One red check on the finished tree — and the node completed three minutes later with "Now build and test".

## Law

A leaf's "done" carries the check it ran, or says it ran none. A deliverable naming no check is not a quieter deliverable than one naming a failing check; it is the same claim with the evidence removed, and the gate, the growth round and the person are all entitled to tell those apart.

## What the finish path checks today

`internal/exec/linear.go:1317` — `land()` is the one exit every landing passes through. It records what the leaf left on disk (`outcome.Artifacts`), takes the closing photograph, reads the meter and grades. **It never asks what the leaf ran.** `verdictFor` (`internal/exec/linear.go:1372`) grades from `Stop` alone: a leaf that stopped under its own power is `provider.VerdictUnverifiedSuccess`, whatever it did or did not execute.

**The "what did you run" field exists and nothing a leaf does fills it.** `exec.Outcome.Account` (`internal/exec/executor.go:426`) is documented as "the files it changed, **the checks it ran**, and what each one found". It is filled by `AccountFor` (`internal/exec/accountfor.go:83`) at PhotographAfter, and its check list comes from `ranChecks` (`internal/exec/accountfor.go:202`), which reads `outcome.Verification` — *the harness's own* reading of the finished tree. A leaf that ran the suite twice and a leaf that ran nothing produce an identical `Account`. There is no field anywhere recording the commands the leaf itself issued.

**The mutation-without-check signal exists, and was switched off for exactly this leaf.** `SelfCloseFindings` (`internal/exec/selfclose.go:90`) raises `SelfCloseOwnChecks` — "the checks this work wrote are red" (`internal/exec/selfclose.go:61`) — and `internal/exec/linear.go:1003` puts it to the leaf before it lands. The close is gated on `RoomLeft`, which ends (`internal/exec/selfclose.go:210-215`):

```go
if room.Left && outcome.Exhausted != "" {
    room.Left, room.Why = false, "it had already been told to land"
}
```

`task-2-x1` was exhausted — `seq 199 | leaf_exhausted | {"bound":"budget","turns":20,"meter":"cost","reached":214953,"allowance":150070}`, having taken the landing reserve at turn 16 ("budget exhausted — landing reserve granted"). So the one mechanism that would have named the red never ran, and the silence went out as a delivery.

Nor did the red reach the next leaf. The handover carried the files and not the finding:

```
seq 220 | job_growth   | {"reason":"overrun","moved":[3 files],"red":8,"finding":{}}
seq 221 | leaf_resumed | {"turns":20,"files":[3 files]}
```

`finding` is empty. `task-2-x2` rediscovered the failure itself, at its own cost, ten turns in.

## Fix shape

1. **A landing carries what the leaf ran.** The loop already sees every `sh` call; the commands the leaf issued belong on the outcome beside the files it changed, so `Account` can tell "ran the suite" from "ran nothing". Where a leaf ran no check, the deliverable says so in words — not by omission.
2. **A red the photograph measured is never dropped in silence.** Where the self-close cannot fire because the leaf had already been told to land, the finding goes into the handover — the `job_growth` row's `finding`, the resumed leaf's note, and the person's line — instead of into a `"red":1` nobody reads.

Named tests: a scripted leaf that edits a package, writes a failing check and lands under a budget reserve delivers a line saying no check was run, and is not reported as done without one; the same leaf's measured red reaches the next leaf's brief.

## Replication

Store: `~/.aforge/runs/aforge-do-4188158208/graph.db` (run 1 of item #515, 2026-09-03T10:45–11:27Z). Queried on a **copy** — a run store has a half-life; see `bench/worker-trial/captures/README.md`.

```sql
-- the brief's test instruction
select payload from events where kind='acceptance';                 -- seq 33

-- the leaf's own contract, composed at runtime and nowhere in the tree
select body from transcript where node_id='task-2-x1' and kind='note' limit 1;

-- every call after the last edit: one, and it is not go test
with c as (select node_id,seq,idx,tool,body from transcript where kind='tool_call'),
     le as (select node_id, max(seq*10000+idx) k from c
            where tool in ('edit','write') group by node_id)
select c.tool, c.body from c join le on c.node_id=le.node_id
 and (c.seq*10000+c.idx) > le.k where c.node_id='task-2-x1';
-- sh | {"cmd": "cd /home/santosh/af-trial-515 && go build ./... 2>&1"}

-- the done message
select summary from nodes where id='task-2-x1';
select payload from events where kind='node_completed' and node_id='task-2-x1';   -- seq 224

-- the red the photograph measured, and the exhaustion that switched the self-close off
select payload from events where kind='verification'  and node_id='task-2-x1';    -- seq 197, "red":1
select payload from events where kind='leaf_exhausted' and node_id='task-2-x1';   -- seq 199

-- the failing test, as the NEXT leaf saw it
select body from transcript where node_id='task-2-x2' and kind='tool_result' and idx=10;
```

Failing test: `TestReaskPromptNamesBrokenRule`, `internal/shaped/shaped_test.go:524` — "the object-shaped answer must be repaired".

From the hand-offs of #515. Report: `bench/worker-trial/REPORT.md` on branch `trial/aforge-as-worker` @ `89786a4d` (local only, at `~/af-trial`). Sibling findings from the same three runs: #549, and the #514 row below it.

Queued for after the WIP freeze; do not start without the coordinator.

AbirAbbas:
## What was wrong

A leaf edited a package, wrote two new tests, confirmed they failed, implemented its fix, and then ran exactly one command — `go build ./...` — and landed. It was recorded **done**, with "Now build and test" as its summary. The brief had named `go test ./internal/shaped/` twice and it was never run; the next leaf ran the suite in its third turn and found the failing test the first leaf had written, rediscovering the whole diagnosis at its own cost. Nothing anywhere in the record said which commands the leaf itself had issued, so a leaf that ran the suite twice and a leaf that ran nothing produced an identical `exec.Account` — its check list is the harness's own reading of the finished tree, never the leaf's claim to have checked its work. And the closing photograph *had* measured the red; the one mechanism that would have put it to the leaf is gated on the leaf having room left, this leaf had already been told to land, so the finding went into a `"red":1` nobody reads and the handover carried the files without it.

**A deliverable naming no check is not a quieter deliverable than one naming a failing check; it is the same claim with the evidence removed.**

## What is true now

A leaf's landing carries what the leaf itself ran, and a red it was never asked to settle travels with the work instead of stopping at the landing.

- The loop records every shell command the leaf issues, as the `sh` tool itself spells it — one function now reads that text and both the tool and the record call it, so the two cannot disagree. `Outcome.Commands` keeps the newest 20, clipped, and `Outcome.CommandsRun` counts them all, so a bounded list can never read as the whole run. A batch refused unexecuted at the output limit records nothing, because nothing ran.
- The account carries them, under `What the work ran itself:`, kept apart from the checks the closing photograph took. A command is never evidence that anything passed — nothing parsed its output — and `Account.Verified()` still answers from `Checks` alone.
- Where the work changed files and no check was run, the account says so in words: `no check was run on the finished tree`, spelled once and used both in the full report and in the one clause the plan's state view carries for that node. That is the one deliberate exception to the emptiness law in this change, and it is commented as such: this is a measured fact — the work changed the tree and nothing checked it — not an unknown rendered as something.
- An account with nothing to account for is still nil. Commands ride an account that exists because the leaf changed files or a reading was taken; they never bring one into being.
- `Outcome.Standing` is what the leaf's own closing photograph found and the leaf still held when it landed, settled at the one seam every exit passes through. `SelfCloseFinding` grew a `Fact` — the finding without the instruction addressed to a worker that has already stopped — and every existing sentence is composed from it, so the two cannot drift and no person-facing wording moved. The handover a dead leaf gives whoever picks the work up now names those facts.

## The validation contract, and the test that proves each line

| behaviour | test |
| --- | --- |
| A landing records the commands the leaf itself issued, in order; a write is not a command, and a batch refused unexecuted records none | `exec.TestALandingCarriesTheCommandsTheLeafRan` |
| The record is bounded and says how many earlier commands it left out | `exec.TestTheCommandRecordIsBoundedAndSaysWhatItLeftOut` |
| The account names those commands, separately from the closing photograph's checks | `exec.TestAnAccountNamesWhatTheWorkRanItself` |
| An account of changed files with no check says so, in words; one that carries a check does not | `exec.TestAnAccountWithNoCheckSaysSoRatherThanLeavingItOut` |
| A leaf that changed files and ran no check delivers that line from the landing itself — and one that ran a build names the build and still says no check was run | `exec.TestALeafThatChangedFilesAndRanNoCheckSaysSoWhenItLands` |
| A leaf's own commands are never a check | `exec.TestCommandsAreNotChecks` |
| A leaf with nothing to account for still leaves the account nil, however many commands it ran | `exec.TestALeafWithNothingToAccountForStillLeavesTheAccountNil` |
| A finding the leaf lands holding is on the outcome, not only in the journal | `exec.TestASpentMeterLandsTheFindingWithNoClose` |
| A leaf that settled its own finding lands holding none | `exec.TestALeafRestoresThePublicNameItDeletedBeforeItLands` |
| Every finding's instruction is composed from the fact a handover carries | `exec.TestEveryFindingALeafRaisesAgainstItselfNamesWhatItIsAbout` |
| The finding reaches the next leaf's brief | `resident.TestLeafStateNamesTheFindingTheLeafLandedHolding` |

## The manual

`internal/manual/chat/adaptive-runs.md` gains its own section — *When a worker says it is done and never ran the check* — covering the command list, the bound, the exact sentence a person will read where nothing checked the work, and the finding that travels to whoever picks the work up. One probe added to the retrieval table.

## Gates

Run from the worktree, outside any sandbox, with `GOFLAGS=-buildvcs=false`:

- `go build ./... && go vet ./... && make fmt-check` — green
- `go generate ./internal/manual && go test -tags=aforge_packed_manual ./internal/manual && git diff --exit-code` — green, no tracked change from generation
- `go test ./internal/manual/` and `go test -run Manual ./internal/tui3/ ./internal/session/` — green
- `make test-laws` — green, 43 files in 14 packages
- touched packages, `make test TEST_FLAGS='-count=1 -p 1' PKGS='./internal/exec ./internal/manual ./internal/resident'` — green
- and, because they read the account's renderers although the change does not touch them: `./internal/revision ./internal/plan ./internal/verify ./internal/store` and `./cmd/aforge` — all green

## What could not be verified here

- The issue's Replication queries a run store from the trial machine, which is not reachable from this checkout; the failure is reproduced instead by the scripted landing tests above, which is what its Named tests ask for.
- The issue's fix shape also asks for the finding to reach the `job_growth` row's `finding` field. That is deliberately **not** in this change: the only way to set it is `resident.WithFinding`, which feeds the governor's per-name round-spending arithmetic, so it changes what buys a repair round — a different seam with its own laws, and one the issue's Named tests do not name. Worth its own change.
- The reservation that counts what it could not check and names none of it is #572 and is untouched here.

Closes #550


santoshkumarradha (dismissed):
Reviewed at `d1bf04866`, rebased onto dev (clean, four commits, no conflicts). Green on the rebased tip: `gofmt` clean on all nine Go files, `go build ./...`, `go vet` on the three packages, `go test -count=1 -timeout 15m` gives `ok internal/exec 22.365s`, `ok internal/resident 25.963s`, `ok internal/manual 1.101s`, `make test-laws` 44 files / 14 packages all ok, prefix budget unmoved at 47531/48000 — the `tools.go` change adds no schema surface, so it costs zero.

**Both headline claims are proven in one real run** on `deepseek/deepseek-v4-flash` through `aforge do`, read out of the gate request body:

```
What the work changed:
1 file changed, +0 +0 lines
  added greet.sh (+0 +0)

What the work ran itself:
  cat greet.sh
  sh greet.sh

What was run to check this work:
  no check was run on the finished tree
```

That is a leaf that ran two commands and still says no check was run — the exact case the change exists for, and the block kept properly apart from the photograph's. Five runs, $0.0052. Worth recording for the next lane: `node.Checked` is **not** persisted to the `--db` store, so the observable door for this is the gate prompt with `AFORGE_CALL_LOG_BODIES=1`; grepping the store for either string finds nothing.

I verified the other structural claims rather than taking them. `noCheckWords` really is spelled once (`account.go:432`) and reaches all four destinations through the single `rows(signoff)` renderer. `Verified()` still answers from `Checks` alone (`account.go:244-254`) and no production reader treats a command as passing evidence. And every exit of `Linear.Run` passes through `land()` — I had the returns at 792, 796, 852, 873, 902, 923, 1209, 1258 and 1269 enumerated individually, plus 1018 returning the `land` from :988 — so `Standing` is settled on all of them.

One thing to fix before it lands, and two smaller ones.

### The emptiness-law exception fires on runs where a check DID run

This is the finding. The renderer's only condition is `len(a.Checks) == 0 && len(a.Files) > 0` (`account.go:458-466`, and again at `:499-500` in `Summary()`). It never consults `Verification.Taken`, `AfterTaken` or `Unread`. But `ranChecks` returns nil (`accountfor.go:207`) whenever `!reading.AfterTaken || len(after.Reported) == 0`, and `photograph.go` reaches that state from four different worlds:

| `photograph.go` | what actually happened | what the account now says |
| --- | --- | --- |
| `:378 case !ok` | "the finished tree could not be read: `<cmd>` could not be started a second time" | `no check was run on the finished tree` |
| `:381 case after.Uncollected` | "`<cmd>` ran on the finished tree and its suite failed to collect" | same |
| `:393 case after.TimedOut && len(after.Reported)==0` | "`<cmd>` ran on the finished tree and was killed at its ceiling of `<d>` without naming a single check" | same |
| `:315 !reading.Taken` | no reading was ever taken | same |

In the first three a command demonstrably ran on the finished tree. The same `Outcome` carries both sentences and they reach a reader together — `revision/acceptance.go:1132` renders `Verification.Unread` while `revision/judge.go:875` renders `Account.Lines()` — so the record now says "nothing checked it" beside "the thing that checked it could not be read".

The change entry's justification is that this is "a MEASURED fact, not an unknown", and on the fourth row that is true. On the first three it is precisely an unknown, wearing the words of a measurement — which is the failure the emptiness law exists to prevent, inverted. `accountfor.go`'s own comments already say this in as many words: *"an incomplete observation"*, *"nobody looked is not everything passed"*.

Please gate the branch on the reading actually having been taken — `reading.Taken && reading.AfterTaken && Unread == ""` or equivalent — and let the other three fall back to silence or to `Unread`'s own sentence. `adaptive-runs.md:808-812` states it unconditionally too and inherits the same defect, so it moves in the same change.

My own e2e is, incidentally, a live instance of the honest row: the scratch dir has no repository and no suite, so nothing was ever attempted. That is the case the sentence is right about.

### Smaller

**`snip` slices bytes, and this string is now person-facing.** `internal/exec/trace.go:473-478` does `return text[:limit] + "…"`, so a command clipped at `ranArgumentBytes` (200) can split a UTF-8 rune and put invalid bytes into a line a person reads. Pre-existing for `Ran`, newly person-facing for `Commands`. The rune-safe twin already exists — `plan/graph.go:1244`'s `clipRunes`, whose comment says *"a mangled character here would ride the sentinel's whole prompt"* — so this is one call away.

**The "one reading" claim is `internal/exec`-scoped.** `internal/tui/node.go:975`'s `toolCallBlock` maps `"sh" → "cmd"` and pulls it with `extractStringField`, so a second reader of an `sh` call's arguments survives — and it reads only the `"cmd"` form, where `shellCommand` also accepts the array and serial-script forms (`tools.go:1356-1367`), falling back to raw JSON. The entry's sentence is true where it matters (the tool and the landing record cannot disagree, which is the stated goal) and false as a repo-wide statement. It is a v1 surface, so I would fix the sentence rather than the code.

**Two wording overreaches while you are in there.** "records every shell command the leaf issues" — `noteCommand` runs unconditionally beside `record(call, ...)` (`linear.go:1111`), so a call that failed is counted as issued; I read `Toolbox.sh` (`tools.go:1388-1429`) and there is no policy-refusal path, so in practice issued does equal ran, but `noteCommand`'s doc comment "after its call has executed" is stronger than the code guarantees. And the batch-refused-at-the-output-limit case is correct as claimed (`linear.go:1050-1061` records none).

### Confirmed clean

One `## ` heading added — `## When a worker says it is done and never ran the check — a done that names nothing it ran` (`adaptive-runs.md:800`) — and it displaces nothing: the whole probe table passes, the new probe lands, and the tui3/session name gates are green. No surviving claim in the corpus that the record cannot name a worker's commands. `SelfCloseFinding.Fact` sits beside the existing fields with every `Sentence` composed from it, so no person-facing wording moved, and `Outcome` has no JSON tags so nothing serialises — an older outcome has an empty `Standing` and renders as it did.

santoshkumarradha:
Pushed `7c059d5` with the blocker and the two smaller items. What is in it, and one correction to my own review.

**The exception now fires only where it is true.** `Account.Unread` carries `Verification.Unread`'s sentence into the account, and `rows`/`Summary` say that instead of `no check was run on the finished tree` when it is set.

**My review got the boundary wrong, and your test caught it.** I first keyed the split on `Unread` being non-empty, which broke `TestALeafThatChangedFilesAndRanNoCheckSaysSoWhenItLands`:

```
landing account did not say no check was run:
  What was run to check this work:
    this project declares no way of checking itself, so there is no reading to take
```

`verify.Reading.Unread` holds two different kinds of sentence. Before a baseline exists it explains why nothing WILL be read — no way of checking this project, a wall too short to afford a reading — and that is the measured absence your exception was written for, not a failure to read anything. After one exists, `exec/photograph.go` uses the same field to explain why the finished tree could not be read. So `AccountFor` fills `Account.Unread` only when `Verification.Taken`, and that boolean is the real line between the two. The commit says this and so does the entry; the manual page names both worlds.

**The tests fail on revert**, which I checked rather than assumed. `TestAReadingThatCouldNotBeReadSaysThatAndNotThatNothingRan` walks the three unread worlds and the absence:

```
--- FAIL: TestAReadingThatCouldNotBeReadSaysThatAndNotThatNothingRan
    the strategy would not start a second time: the report calls an unread reading a measured absence:
      What was run to check this work:
        no check was run on the finished tree
    the strategy would not start a second time: the report does not say why the tree could not be read
```

**`snip` no longer cuts through a character.** It sliced bytes, which was harmless while it only reached a trace's call tail and is not now that it reaches `Account.Commands` under "What the work ran itself:" — a command clipped at `ranArgumentBytes` put invalid UTF-8 in front of a person. `TestABoundedCommandIsNotCutThroughACharacter` clips a run of three-byte runes and checks `utf8.ValidString`.

On where that back-off lives: I kept it local rather than reaching for `internal/plan`'s `clipRunes`, because the tree's own convention is unambiguous — `internal/plan`, `internal/search`, `internal/tui3`'s reveal, `internal/resident`'s revise and five places in `internal/session` all spell these two lines themselves rather than depend on each other. Exporting one of them for `internal/exec` would make this the only cross-package dependency of its kind, and moving them all somewhere shared is a tree-wide change that does not belong in this PR. The comment names the siblings so the next person can see it is a convention and not an oversight.

**The "one reading" sentence is scoped.** The entry now says `exec.shellCommand` is that one reading *within `internal/exec`*, and names the survivor: `internal/tui/node.go`'s `toolCallBlock` still pulls a `"cmd"` field out of the raw arguments for the v1 surface, and handles only that spelling where `shellCommand` also takes the array and serial-script forms. True where it matters — the tool and the landing record cannot disagree — and it was reading as a repo-wide claim.

Green on the branch: `gofmt` clean, `go build ./...`, `go test -count=1 -timeout 15m ./internal/exec/` → `ok 22.394s`, `./internal/manual/` → `ok 1.154s`, `scripts/laws.sh` 14/14, `make changelog-check` well formed, prefix budget unchanged.

**A fact for whoever verifies this next, not an ask:** `node.Checked` is not persisted to the `--db` store, so neither `no check was run on the finished tree` nor the new sentence can be found by grepping it. The observable door is the gate prompt with `AFORGE_CALL_LOG_BODIES=1` — that is where I read both blocks out of a real run.

**What I did not touch.** The commands list, the bounds, the `Standing` settling, `SelfCloseFinding.Fact`, and `resident.LeafState`. I checked those rather than trusting them — `noCheckWords` is spelled once and reaches all four destinations through the single `rows(signoff)` renderer, `Verified()` still answers from `Checks` alone, and every exit of `Linear.Run` passes through `land()` (the returns at 792, 796, 852, 873, 902, 923, 1209, 1258, 1269, plus 1018 returning the `land` from :988). They are right.

Two wording notes I left alone: `noteCommand`'s doc comment says "after its call has executed" while it runs unconditionally beside `record(call, ...)`, and there is no policy-refusal path in `Toolbox.sh` today, so issued does equal ran — the comment is just stronger than the code guarantees. And #610 also edits `adaptive-runs.md` and `internal/exec/linear.go`, so whichever of the two lands second will need a rebase; the manual insertions are close enough to conflict textually.

PR #608, opened 2026-09-03
session: a landing time is a fact of the record, carried across the restore

### LINKED ISSUE #530: session: a restored task has no landing time on its record, so it is dated in the future and the tasks place drops it
## What happened

2026-09-03, base `dev@f38d6c01a`.

**A task restored from a checkpoint is dated in the future, and the tasks place then drops
it off the page.** A piece of work that ran for twelve minutes and finished twenty minutes
before the window was closed comes back, on the next open, stamped twelve minutes from
*now* — and the place's own date window reads that as tomorrow and does not draw the row
at all. The task that needed a person went missing from the tasks place entirely while its
shorter siblings sat above it saying `now`.

Both halves of that come from the same hole: **a landing time is being derived from when a
window opened, because the record does not carry one.**

The derivation, in `internal/tui3/place_tasks.go`'s `taskNodeEnded`:

```go
at := node.began
if at.IsZero() {
	at = node.met
}
...
return at.Add(node.elapsed)
```

`met` is the moment *this terminal* first heard of the node. For a node replayed out of a
checkpoint that is the moment the person sat down, so `met + elapsed` is a landing time in
the future. The drop is one line in `internal/tui3/tasksplace.go`'s `readTasks`:

```go
if !r.win.Holds(tasksEntryAt(item.entry, now)) {
	continue
}
```

`UsageWindow.Holds` (`internal/session/usage_spend.go`) is `!at.Before(From) &&
at.Before(next bucket after To)`, so an instant past the end of today is not held and the
row is skipped — not filed under `earlier`, not folded: gone.

The `now` on the siblings is the same absence seen from the other side: `tasksEntryAt`
answers *the moment of the reading* for a row with no stamp, which is right for a window
that cannot judge an undated row and wrong for `sinceAt`, which then prints `now` over
work that nothing is doing.

## This issue is the ENGINE half only

The surface half is already landing in the `ui/polish-v0` wave: `taskNodeEnded` stops
falling back to `met` for a restored node, and the age field reads a stamp rather than
substituting the reading's clock, so a row with no recorded landing time draws **nothing**
where it used to draw `now` (the emptiness law). That change makes the surface honest and
makes it read the record only. It cannot make the record right, and that is what is left.

The coordinator's ruling is that **a landing time is a fact of the record, carried across
the restore, and never derived from the window's open time.** So this issue is the whole of
what remains, and none of it is in `internal/tui3`.

## Where

Everything below is `internal/session`.

- **`TaskNode.started`** (search `started  time.Time` in `task_run.go`) is the only real
  start, and it is set exactly once, at the transition to `TaskRunning` in
  `TaskGraph.startLocked` (search `node.started = time.Now()`). **It is not written to the
  checkpoint.** `taskRecord` in `task_store.go` carries `ElapsedMS` and no stamp at all,
  and the comment on the neighbouring `elapsed` field concedes the consequence in so many
  words: *"a node rehydrated from a checkpoint has no started to measure from and this is
  the age it had."*
- **`TaskNode.runStart()`** in `taskground.go` (search `func (n *TaskNode) runStart`)
  already makes the same guess the surface was making — `time.Now().Add(-n.elapsed)` — and
  says so: *"its window is measured back from now, which lands LATER than the truth."* It
  chooses that direction deliberately for its own purpose (a ground overlap: a false alarm
  costs more than a quiet miss), so it is evidence that the fact is missing rather than a
  second bug, but it is a second reader that would rather have the stamp.
- **`TaskIndexEntry.EndedAt`** in `task_index.go` is the durable landing time — *"when it
  landed, and it is zero for a row merged in live"* — and it survives a restart **only for
  a node whose row actually reached the project's `tasks.jsonl`**. For a row rebuilt from
  the graph, `(*TaskNode).indexEntryLocked` (search `A landed node's EndedAt is now minus
  nothing`) stamps it `entry.EndedAt = time.Now()`, and its own comment concedes the shape:
  the original stamp is *"where the file's row is"*, which is precisely what a node whose
  session died before the row was written does not have.
- **`TaskNotice`** in `task_contract.go` carries `Elapsed time.Duration` and no landing or
  starting stamp — its only `time.Time` is `Deadline`, a proposal field — so even where the
  engine does know when the work landed, the surface has nothing to read it out of.

## The fix

Three changes, all in `internal/session`, and the shape is one sentence: **the record keeps
the stamps, and everything downstream prefers them to `time.Now()`.**

1. `taskRecord` grows a landing stamp (and, for the same money, a start), written when the
   node is snapshotted and restored onto `TaskNode` when it is rehydrated. It is an added,
   omitted-when-empty JSON field, so `taskFileVersion` does not move and an older
   checkpoint decodes exactly as it always did — the same discipline `runRecord` already
   follows for `Kind`.
2. `(*TaskNode).indexEntryLocked` prefers the node's own restored stamp over `time.Now()`,
   and stamps `time.Now()` only where it genuinely is the landing instant — the live
   transition, which is what `(*Agent).recordTaskIndex` is for.
3. `TaskNotice` grows `EndedAt` (and `StartedAt`) beside `Elapsed`, so a surface drawing a
   restored row has a fact to draw instead of a subtraction to perform.

`TaskNode.runStart()` keeps its guess as a fallback and should use the restored stamp when
there is one; the comment there should stop describing the guess as the only option.

What a person sees: a task that landed at 23:28 says it landed at 23:28 whether they are
looking at it at 23:30 or the following week, it sits in the tasks place under the day it
actually landed on, and a task whose landing time genuinely was never recorded draws **no
age at all** rather than `now`.

## Replication

**Deterministic (no model, no key that is ever used).** Everything is in the repository.

1. `make build`, then make a throwaway home with a conversation in it:
   `make demo-home DEMO_HOME=/tmp/aforge-stamp` and quit the surface it opens.
2. Beside any conversation's `transcript.jsonl` under
   `/tmp/aforge-stamp/.aforge/v3/projects/*/*/`, write the checkpoint a task that ran for
   twelve minutes and landed leaves behind — this is exactly what `taskRecord` writes
   today, and the thing to notice is that there is nowhere in it to put *when*:

   ```json
   {"type":"tasks","version":1,"seq":10,
    "nodes":[{"id":1,"title":"Rotate the staging certificate","state":"unverified",
              "elapsed_ms":720000}],
    "runs":[]}
   ```

   Leave that project's `tasks.jsonl` absent, which is what a session that died before its
   index row was written looks like.
3. Reopen that conversation
   (`cd /tmp/aforge-stamp/<project> && HOME=/tmp/aforge-stamp <repo>/bin/aforge`, enter on
   the card) and open the tasks place with `/history`.

**What a developer sees today:** the row is not on the page. The head sentence counts the
work the window holds and this task is not among it, because the surface has dated it
`<the moment you pressed enter> + 12m` and the window ends at the end of today. Widen the
window with `shift+←` as far as it goes and it does not come back — a future stamp is past
every window's far edge, not behind its near one. Meanwhile any row in that place with no
stamp at all reads `now`.

**Field (real models, needs `OPENROUTER_API_KEY`, a few cents and a few minutes).** The
product version, and the one that was actually hit: in a conversation, hand aforge a task
and let it land — one that ends `needs your look` is the case that hurts, because that row
is the one somebody has to come back to. Quit with `ctrl+c` twice. Reopen the same
conversation and open the tasks place. The task is not on it. Its shorter siblings, if any
are undated, say `now`.

## Acceptance

- **e2e:** through the real surface on the deterministic home above — after reopening the
  conversation, the tasks place draws the row `Rotate the staging certificate`, and its age
  cell reads the age measured from the landing time on its own record (`12m` past the
  recorded landing, not `now` and not a future instant). Today the string
  `Rotate the staging certificate` does not appear on the page at all.
- **e2e:** the control — a checkpoint carrying `elapsed_ms` and **no** landing stamp (every
  file written before this change) still resumes, and its row draws **no age cell at all**:
  the page must not contain `now` for that row, and the row must still be on the page. That
  is the emptiness law, and it is what the surface half in `ui/polish-v0` already does once
  the record stops lying to it.
- **Unit:** a checkpoint written by `taskStore` and read back reproduces the node's landing
  time exactly — `restored.EndedAt == original.EndedAt` — and the round trip of a
  checkpoint written *before* this field existed still decodes at `taskFileVersion` 1 with
  a zero stamp and a preserved `ElapsedMS`.
- **Unit:** the structural law — every path that stamps a `TaskIndexEntry.EndedAt` with
  `time.Now()` is the live landing transition. A rebuilt row takes the record's stamp when
  it has one, so `(*TaskNode).indexEntryLocked` on a restored node returns the recorded
  instant rather than the instant it was called.
- The manual pages that describe reading a task's age
  (`internal/manual/chat/reading-a-task-page.md`, `internal/manual/chat/tasks.md`) quote
  the new behaviour — a landing time is what the record says, and no recorded time draws
  nothing — and the change entry's `invalidates` names what people believed before.

AbirAbbas:
## What was wrong

A task restored from a checkpoint was dated in the future, and the tasks place then dropped
it off the page. Work that ran for twelve minutes and finished twenty minutes before the
window was closed came back, on the next open, stamped twelve minutes from *now* — the
place's date window read that as tomorrow and did not draw the row at all, so the one piece
of work that needed a person went missing while its shorter siblings sat above it saying
`now`. Both halves came from one hole: **a landing time was being derived, because the
record did not carry one.** `taskRecord` wrote `elapsed_ms` and no instant at all, so every
reader that wanted a moment invented one — the surface from when the window opened, and
`(*TaskNode).indexEntryLocked` from `time.Now()` at the instant somebody read the row.

The surface half landed in #518: v3 stopped doing the arithmetic and now draws nothing where
the record has no stamp. This is the engine half — the record itself.

## What is true now

A landing time is a **fact of the record**: written when the node is snapshotted, restored
when it is rehydrated, and never derived from when a window opened. A task that landed at
23:28 says 23:28 whether you look at it at 23:30 or the following week, it sits under the day
it actually landed on, and a task whose landing time genuinely was never recorded draws no
age at all rather than `now`.

- `taskRecord` carries `startedAt` and `endedAt`, added and omitted-when-zero, so
  `taskFileVersion` does not move and an older checkpoint decodes exactly as it always did.
- `TaskNode.ended` is stamped at the three live transitions into a settled state and nowhere
  else, and `(*TaskNode).indexEntryLocked` reads it instead of the wall clock — leaving the
  row's stamp zero where the record has none rather than substituting the reading's clock.
- `TaskNotice` carries `StartedAt` and `EndedAt` beside `Elapsed`, and v3 reads them for the
  row's age, the completion card's `started 14:02` and its `ran · 14:02 → 14:14`.
- `(*TaskNode).runStart` reads the recorded start; its `now - elapsed` guess is the fallback
  for an older record and nothing else.
- One `(*TaskNode).ageLocked` now holds the rule the checkpoint, the index row and the notice
  had each written out for themselves: a settled node's age is the frozen fact its record
  carries, and only unsettled work consults the moving clock.

| Behaviour | Test |
| --- | --- |
| A landed node's start and landing time survive the process, through the JSON a later process reads | `TestACheckpointCarriesTheLandingTimeAcrossARestore` |
| A checkpoint written before the stamps existed still decodes at version 1, keeps its `elapsed_ms`, rebuilds with two zero instants, and costs no new bytes when written again | `TestACheckpointWrittenBeforeStampsExistedStillResumes` |
| A row rebuilt from a restored node carries the RECORDED landing instant, not the instant it was built | `TestARestoredNodesRowKeepsTheRecordedLandingTime` |
| A settled node with no recorded landing time gets a zero stamp — still a row, no age drawn | `TestANodeWithNoRecordedLandingTimeDrawsNoStamp` |
| A landing that happens in this process still stamps the moment it landed, in the project's index | `TestALiveLandingStampsTheMomentItLanded` |
| A settled node's age is frozen and never re-measured from its restored start | `TestARestoredSettledNodesZeroAgeStaysFrozen` |
| The notice carries both recorded instants, and zero where nothing recorded them | `TestANoticeCarriesTheRecordedStamps` |
| The ground overlap reads the recorded start rather than guessing backwards from now | `TestTheGroundOverlapReadsTheRestoredStart` |
| Every place in `internal/session` that stamps a row's `EndedAt` with `time.Now` is a live landing transition, and the row builder reads no clock at all | `TestOnlyALiveLandingStampsARowWithNow` (a law, on the fast gate) |
| A restored task's own stamps outrank the moment the surface met it | `TestARestoredTaskReadsTheRecordsOwnStamps` |
| A restored task with no stamps is still undated, and a watched one keeps its live clock | `TestARestoredTaskIsNeverDatedInTheFuture`, `TestAWatchedTaskKeepsItsOwnLandingTime` |

## Gates

| Gate | Result |
| --- | --- |
| `go generate ./internal/manual && go test -tags=aforge_packed_manual ./internal/manual` | pass |
| `go test ./internal/manual/` and `go test -run Manual ./internal/tui3/ ./internal/session/` | pass |
| `make test-laws` | pass |
| touched packages: `make test PKGS="./internal/session ./internal/tui3"` | pass except `TestADropTypedSlowlyStillLands` |
| `go run ./cmd/aforge-changes check` | pass |

## What I could not verify

`TestADropTypedSlowlyStillLands` fails in the full `internal/tui3` package run and passes on
its own. It was reproduced failing at the same assertion on a clean tree at this branch's
base with the change removed, so it is a pre-existing full-suite interaction and not this
change; no line of it or of the drop-key road is touched here, and no known-red entry was
added for it.

The issue's two e2e acceptances drive the real binary against a real model. This machine has
no `OPENROUTER_API_KEY`, so neither was run. Both are covered end-to-end inside Go instead:
the engine test proves the row a restored node hands the tasks place carries the recorded
instant, and the surface tests prove what the place and the card draw from it, with and
without a stamp.

## The manual

`internal/manual/chat/tasks.md` owns the tasks place — where a settled row's age comes from,
that work reopened from a previous session keeps the time it landed, and that a row whose
older record never carried one carries no age while staying on the page.
`internal/manual/chat/reading-a-task-page.md` owns one task's own page — its recorded start,
and the `started 14:02` segment the card omits when there is none. Both topics have a
retrieval probe in `internal/manual/chat_test.go`.

Closes #530


santoshkumarradha (dismissed):
Reviewed at `9329d07a2`, rebased onto dev (clean, four commits, no conflicts). Green on the rebased tip: `gofmt` clean on all twelve Go files, `go build ./...`, `go vet` on the three packages, `go test -count=1 -timeout 15m ./internal/session/ ./internal/tui3/ ./internal/manual/` → `ok 127.6s` / `ok 459.3s` / `ok 1.2s`, `make test-laws` 46 files / 14 packages all ok, `make test-packed-manual` ok, changelog well formed, prefix budget unmoved at 47531/48000 — this change adds zero bytes.

Worth saying first: `TestADropTypedSlowlyStillLands`, which your body lists under "what I could not verify" as failing in the full tui3 run, **passed** in the full package run after the rebase. That was base drift, not yours.

The design is right and I checked the two things that would have sunk it.

**Backward compatibility holds in both directions, against a real artifact rather than a fixture.** A pre-#608 `internal/session` (clean dev) decoded a checkpoint written by this branch's binary:

```
old binary decoded a #608 checkpoint: version=1 nodes=1
  id=1 state=running elapsed_ms=9050 title="Create landing.txt"
```

Decode is plain `json.Unmarshal` with no `DisallowUnknownFields`, the version check is unchanged, and `omitzero` is live on `go 1.26.5`. So a downgrade reads a new record and a new binary reads an old one.

**No zero time reaches a surface as a date.** I had all 25 non-test readers of `EndedAt` traced: every one guards `IsZero()` or goes through `sinceAt`, which returns `""`. And no zero-stamped row can reach the durable file at all — both `appendTaskIndex` callers backstop it. The zeros live only in the in-memory merged view.

**And the `startedAt` half is proven through a real door.** `aforge do` cannot show this — it runs the plan engine into `graph.db` and writes no `tasks.json` anywhere — so this went through `aforge chat --once` with a `propose_task`, which wrote a real checkpoint:

```json
"state": "running", "elapsed_ms": 9050,
"startedAt": "2026-09-03T21:08:39.663355052-04:00"
```

with `endedAt` absent because the node was still running — `omitzero` demonstrated on a real record. A separate process **10m38s later** read it back:

```
READ INSTANT (this fresh process) = 2026-09-04T01:19:17.685Z
  record.StartedAt = 2026-09-04T01:08:39.663Z   <- the actual run start
```

The recorded instant, not the reading's clock. That is the change. Two runs, $0.0053.

Three things to fix. The first is the one I would hold the merge for.

### 1. A page still says the thing this PR just made untrue

`internal/manual/chat/tasks.md:1109-1111`:

> Where nothing knows when the work started — **a task replayed out of a checkpoint, which keeps how long it ran and not when it began** — the stamp is **absent** rather than invented.

A task replayed out of a current checkpoint now *does* know when it began, and `started 14:02` is drawn. You corrected the equivalent sentence in `reading-a-task-page.md` and in `tasks.md`'s own restore section at :1613, and this one was missed.

This is the hazard `CLAUDE.md` names in as many words — "when you make something possible, hunt down the page that says it isn't" — and no gate can catch it, because the gates check that a name is mentioned and never that the claim around it is true. It is one sentence.

### 2. The law has two real holes, and one is already occupied

I ran the law's own `containsTimeNow` against each evasion:

```
e.EndedAt = time.Now().UTC()          caught
now := time.Now(); e.EndedAt = now    NOT caught
e.EndedAt = stamp()                   NOT caught
```

(The `.UTC()` case is caught — `ast.Inspect` descends — so that one is fine.) The variable and helper holes are real, and the tree already contains the variable form: `task_index.go:671,679` does `now := time.Now()` then `closed.EndedAt = now`. It passes the law by not being seen, and its allowlist entry at `task_landing_law_test.go:23` is therefore dead weight — remove the entry today and the law still goes green.

So the body's "names any other place in `internal/session` that stamps a row's `EndedAt` with `time.Now`" is stronger than what runs. The `indexEntryLocked` half — refusing one in the row builder at all — is sound and is the half that matters most. Either widen the check to a variable assigned from `time.Now` in the same function, or soften the claim to what it actually guarantees. I would widen it; the allowlist proves you already know the sites.

### 3. An interrupted design or subharness now loses its age, and two records disagree

`interrupt` (`task_store.go:1267`) sets `record.State = TaskFailed` for `TaskKindHarness` (:1304) and `TaskKindSubharness` (:1323) and never sets `record.EndedAt`, so those restore settled-with-zero and rebuild undated. Meanwhile `closeInflightTaskIndexRows` writes a durable row dated at reopen, and `Agent.TaskIndex` (`task_index.go:548-556`) has the live graph row **replace** the file row for the same node — so the file's better stamp is discarded in favour of the undated one. Before this change the graph row carried `time.Now()`.

Ordinary tasks are unaffected; they restore `TaskQueued`. And drawing nothing is defensible under the emptiness law — I am not sure it is wrong. But it is a behaviour change with no test and no line in the entry, and the same mechanic blanks the age for any old-checkpoint settled node whose durable row does carry a real instant. Please either stamp `record.EndedAt` on those two paths or write the choice down.

### Smaller

- `internal/session/taskground.go:71-74` still says "a rebuilt row's `EndedAt` is **NOW** — so every task this session has ever finished would look like it landed during this run." That is precisely what this PR falsifies. The conclusion still stands; the reason given is now wrong.
- `task_landing_stamp_test.go:88`'s `time.Sleep(time.Millisecond)` proves nothing — `ageLocked` returns the frozen `elapsed` deterministically for a settled node. Dead weight rather than a flake; worth deleting so nobody reads it as load-bearing.
- The body's "three live transitions into a settled state" is imprecise — `task_run.go:1660-1661` is settled→settled. The invariant holds (three sites assign `node.ended`, three assign a settled state, and the sets match); only the wording is off.
- The `tasks.md` heading rename drops "task history", "old and past tasks" and "work from other sessions". The existing `{"where is my task history", "tasks"}` probe still resolves, but that vocabulary now survives only in `commands.md:1288`. Worth a look next time that page is edited.

### On `ageLocked`

Verified properly: all three old copies are gone, and the new one is equivalent for unsettled, settled-with-record and settled-without-record. The one intentional difference — settled with a nonzero start and `elapsed == 0` now returns 0 rather than `time.Since(started)` — is strictly better and is pinned by `TestARestoredSettledNodesZeroAgeStaysFrozen`. Good consolidation.

santoshkumarradha:
Pushed `515bf58` with all six items, and bought the half of the verification that was missing. Two of these turned out differently from how I wrote them up, so both are below.

**The `endedAt` half is now proven through a real door.** My review could only show `startedAt`, because `chat --once` closes the agent at end of turn and that *pauses* an ordinary task rather than landing it. Resuming with `--session <transcript path>` lands it:

```
"state": "done",
"startedAt":  "2026-09-03T21:24:54.830704929-04:00",
"endedAt":    "2026-09-03T21:25:13.119543315-04:00",
"elapsed_ms": 18288
```

read back from a separate process at `21:26:53.314` — **100.19 seconds after the stamp**, and the stamp is still the landing instant. `endedAt − startedAt` is 18.2888s, exactly the recorded `elapsed_ms`, so both come off the same run; `tasks.json`'s mtime equals `endedAt` to the millisecond, which incidentally proves the `omitzero` half too — the last write is the landing checkpoint, and that same record read while running had no `endedAt` at all. One caveat worth stating: the resume *restarted* the paused node rather than continuing it, so this proves the pair for the attempt that actually landed rather than across the pause.

**The stale page.** `tasks.md:1109` now says a task replayed out of a checkpoint carries the instant it began, and reserves the absent stamp for a checkpoint written before the record carried one.

**An interrupt is an ending, for the two kinds that settle on one.** `interrupt` stamps `record.EndedAt` for `TaskKindHarness` and `TaskKindSubharness`, through a small `interruptedAt` that keeps an instant the record already had — a node that landed and was then caught by the close ended when it landed, not when the process did. `TestARestoredInterruptedDesignCarriesTheMomentItStopped` covers all four cases including the one that must NOT move: an ordinary task goes back on the frontier queued and is not stamped, because it has not ended.

**The law, and where I did not do what was asked.** The instruction I had was to widen it *and* delete the `closeInflightTaskIndexRows` allowlist entry as dead weight. Widening makes that entry **live**: it really does write `EndedAt` from a variable bound to `time.Now` (`task_index.go:671,679`), and it was only invisible because the narrow law never looked. Deleting it after widening turns the test red on a legitimate writer, so I widened and kept it. It is documentation of a real site now rather than permission for a question nobody asked.

`TestTheLandingLawFindsTheClockThroughANameAndAHelper` pins all four spellings — written out, through `.UTC()`, through a name, through a helper — and asserts the law does not refuse a row stamped **from the record**, which is the shape this whole change installs.

**And the widening immediately found one more.** `newOrchestrateFamily` (`orchestrate.go:1559`) writes `EndedAt: family.started` on a row whose `Status` is `TaskRunning` — an adaptive family's opening row carries its *start* instant in the ending field, so a run that has not ended is dated by when it began. Reasonable on the tasks place, which needs something to sort by; confusing everywhere else. It is older than this PR and out of its scope to move, so it is allowlisted with exactly that said, and filed separately.

**A correction to my own review:** I passed on a claim that `e.EndedAt = time.Now().UTC()` evades the law. It does not — `ast.Inspect` descends into the selector. Only the name and helper forms were real holes.

The three small ones: `taskground.go:71`'s comment (the conclusion stands, the reason moved); the `time.Sleep` in the frozen-age test, gone, with a line saying why there is deliberately no wait; and "task history", "old and past tasks" and "work from other sessions" back in a sentence under the renamed heading, since the rename had taken that vocabulary out of the search index and it survived only in `commands.md`.

Green on the branch: `gofmt` clean, `go build ./...`, `go vet`, `ok internal/session 122.981s`, `ok internal/tui3 459.768s`, `ok internal/manual 1.240s`, `scripts/laws.sh` 44 files / 14 packages all ok, `make changelog-check` well formed, prefix budget unchanged at 47531/48000.

Two things I checked and left alone because they are right: backward compatibility holds in **both** directions — a pre-#608 `internal/session` decoded a checkpoint written by this branch's binary at version 1 with `elapsed_ms` intact — and no zero time reaches a surface as a date; all 25 non-test readers of `EndedAt` guard `IsZero()` or go through `sinceAt`, and both `appendTaskIndex` callers backstop the durable file. The `ageLocked` consolidation is a genuine three-into-one and its one intentional difference is strictly better.

PR #610, opened 2026-09-03
exec: a leaf knows its clock, and reading forever is not progress

### LINKED ISSUE #540: do: nothing paces a leaf to finish before the wall, and 45 minutes of recon wrote nothing
Three `aforge do --timeout 2700` runs each spent **2700.05 s of a 2700 s wall**. Run 1
made 156 shell calls, every one of them read-only, and wrote **nothing** in 45 minutes.
The no-progress guard never fired once across any of the three runs.

This is filed as a **harness gap**. The model's behaviour is the evidence; the missing
law is the harness's. The split is stated explicitly below.

## The three runs

| run | store | root start (UTC) | last transcript row | wall | sh calls | first mutation |
| --- | --- | --- | --- | --- | --- | --- |
| 1 | `aforge-do-1923110067` | 08:24:30 | 09:09:30 | 2700 s | 156 | **never** |
| 2 | `aforge-do-1792941490` | 09:11:50 | 09:56:50 | 2700 s | 142 | 09:51:59 — **40m09s in** |
| 3 | `aforge-do-4060048231` | 09:57:44 | 10:42:45 | 2700 s | 108 | 10:14:41 — 16m56s in |

Run 1's verb histogram, and the query that makes it:

```sh
sqlite3 graph.db "select v,count(*) from (
  select substr(json_extract(body,'\$.cmd'),1,
                instr(json_extract(body,'\$.cmd')||' ',' ')-1) v
  from transcript where kind='tool_call' and tool='sh') group by v order by 2 desc;"
```

```
grep 82 | sed 54 | cat 11 | find 5 | ls 3 | gsed 1     (= 156)
```

Tool histogram for run 1: `sh` 156, `read` 1 (refused — see #538). **Zero `write`, zero
`edit`.** Runs 2 and 3, which were handed verified file-and-line facts in the prompt,
still managed only 2 and 5 mutating calls respectively.

## What the harness said to the leaf, and when

```sh
sqlite3 graph.db "select node_id,ts,substr(body,1,60) from transcript
                  where kind='note' order by seq,idx;"
```

- Run 1: `budget exhausted — landing reserve granted` at 08:36:32, 08:49:53, 09:02:56.
  **No deadline warning at all.** The leaves died on the token budget, the executor
  spawned the next attempt, and the run rode that cycle into the wall.
- Run 2: `budget exhausted` at 09:36:48 and 09:51:59; `deadline close — landing reserve
  started` on `task-2-x2` at 09:56:50 — the wall instant itself, on a leaf with no time
  left to land in.
- Run 3: `budget exhausted` at 10:14:41; twice `board: did not finish — context deadline
  exceeded`.
- **`conclude directive injected` appears zero times in all three stores.**

```sh
for d in 1923110067 1792941490 4060048231; do
  sqlite3 ~/.aforge/runs/aforge-do-$d/graph.db \
    "select count(*) from transcript where kind='note' and body like '%conclude directive%';"
done   # 0 0 0
```

## Why the existing guards cannot see this — the harness half

`internal/exec/noprogress.go` is the only law that watches a leaf work, and both of its
reachable signals are definitionally blind to recon:

1. **The stagnant signal needs zero mutations AND zero new information**
   (`noProgressStagnantCap = 6`, `noprogress.go:203`). `readTurn` at
   `noprogress.go:196-224` marks `newInfo = true` whenever a result's SHA-256 content hash
   is one it has not seen. 156 *different* greps return 156 *different* bodies, so
   `newInfo` was true on essentially every turn and `stagnantTurns` never left zero.
   Reading something new is counted as progress; **reading forever is therefore
   indistinguishable from working**.
2. **The turn floor is unreachable.** `noProgressTurnFloor = 60` (`noprogress.go:213`),
   and it additionally requires fewer than 5 productive turns in the last 20 — where
   "productive" is the same `mutation || newInfo`. Measured max turns per leaf attempt
   across all three runs: **19, 19 / 23, 21, 9 / 16, 25, 14.** Every leaf lands on its
   token budget at roughly 20 turns, so the 60-turn floor cannot fire on this shape of
   run at all.
3. **The repeat signal** (`noProgressRepeatCap = 4`) needs the same tool, same arguments
   *and* same result four times running. 156 distinct commands never came close.

And nothing anywhere paces against the **wall**:

- The only wall-derived message to a leaf is at `internal/exec/linear.go:798-812`, and it
  fires only once `time.Until(deadline) <= landingReserve`, where
  `deadlineLandingReserve` (`linear.go:1698-1704`) is `deadline/10` capped at **two
  minutes**. Two minutes of a 45-minute wall is 4.4 %, and in run 1 it never arrived
  at all.
- **The leaf is never told how much wall it has, or how much is left, at any earlier
  point.** `RoomLeft` (`internal/exec/selfclose.go:186-216`) does compute `room.Wall`,
  but it is consumed by the self-close pass, not surfaced to the model as a budget it
  can pace against.
- There is no law of the form "N turns in, nothing has been written". `mutationTools`
  (`noprogress.go:147-152`) is `write` / `edit` / `apply_patch` — the harness already
  knows exactly what a mutation is; nothing counts how long a leaf has gone without one
  while the wall runs down.
- The `do` wall default is `defaultDoSeconds = 900` (`cmd/aforge/do.go:45`); the trial
  passed `--timeout 2700`. Note the unit churn landed today — #379, reverted by #494,
  re-landed by #495 — so anything touching the wall should rebase on #495.

## Model vs harness — the attribution, stated

**Model (evidence, not the bug filed here):** `deepseek/deepseek-v4-pro` chose 82 greps
and 54 `sed -n`s before writing anything, and in run 1 never wrote at all despite reading
`CLAUDE.md` four times. Even handed verified line numbers and an explicit "start writing
within your first few turns" in the run-2 prompt, it still went 40 minutes before its
first `edit`. Advisory prompt text did not move it — consistent with what the vs-pi wave
already recorded.

**Harness (the gap to fix):** a leaf may burn an entire wall read-only and no mechanism
notices, because (a) "new information" counts as progress without bound, (b) the only
turn-count backstop sits above the turn count any leaf reaches, (c) the leaf is given no
running account of its wall, and (d) the sole wall-derived nudge arrives inside the last
two minutes or, as in run 1, not at all. A pacing or no-progress-toward-a-mutation law
belongs to the harness — the model's behaviour above is the load it has to hold under.

## Replication

Verb histogram and timeline: the two `sqlite3` queries above, against
`~/.aforge/runs/aforge-do-{1923110067,1792941490,4060048231}/graph.db` (copy before
querying; do not touch the originals). Wall figures: `nodes.started_at` on `root` versus
`max(transcript.ts)`, and the report's own `2700.05s` per run. Code:
`internal/exec/noprogress.go:147-224`, `internal/exec/linear.go:798-812` and `:1698-1704`,
`internal/exec/selfclose.go:186-216`, `cmd/aforge/do.go:45`.

Trial report: `bench/worker-trial/REPORT.md` on the local branch `trial/aforge-as-worker`
(worktree `~/af-trial`; not pushed to origin). Binary built from `dev` at `b95f8c29f`.
Related: #510 (the issue handed to `aforge do` in all three runs), #538 (the refused
`read` call that opened run 1), #495 (the wall's unit).

Queued for after the WIP freeze; do not start without the coordinator.

AbirAbbas:
Workers now receive their available wall time before their first turn, one halfway clock reading, and an advisory reminder after each ten tool-calling turns without a workspace change. Fresh research remains allowed; the existing repeated-call and stagnant-result stops retain their thresholds.

Review corrected the clock to use the effective caller deadline. A worker with ten seconds left under a one-hour nominal lease now hears ten seconds, without inventing fifty-nine minutes of elapsed work. The regression reproduces that failure. Manual and PERF.md describe the actual behavior.

The touched-package CI run exposed an existing stalled-check test whose 200-millisecond fixture window expired during real checker setup. The failure reproduced repeatedly on unchanged baseline code. Its fixture now allows ten seconds while retaining the assertions that the first call times out and the second succeeds; production limits are unchanged.

Validation: full exec and manual suites passed after the correction; focused clock, pacing, and manual tests passed after integrating dev; canonical make build passed. The repaired stalled-check test passed ten consecutive real-door runs under concurrent package load (57.127s). The wall-clock trace fixture also leaves setup room under load; its real trace assertion passed five consecutive runs (25.819s). No private paid-model trial was replayed, and advisory pacing cannot guarantee a model begins writing.

Closes #540

Final GitHub light gate, touched packages, and check pass at the reviewed head. One local full session run hit an unchanged identical-write checkpoint failure (task done instead of failed); 50 baseline repetitions and 200 paired repetitions did not reproduce it. That separate observation remains unresolved and no skip or production change was added for it.


santoshkumarradha (changes requested):
Reviewed at `8a837c953`, rebased onto dev (clean, three commits, zero conflicts). Green: `gofmt` clean, `go build ./...`, `go vet`, `go test -count=1 -timeout 15m ./internal/exec/ ./internal/manual/` → `ok 33.940s` / `ok 1.173s`, **`go test -race -count=2 ./internal/exec/` → `ok 109.815s`** with nothing failing on the second run only, `make test-laws` 44 files / 14 packages all ok, the packed-manual variant ok, prefix budget unmoved at 47531/48000.

**Both mechanisms are proven through a real door**, `aforge exec` on `deepseek/deepseek-v4-flash` with a 30s wall and one read per turn, straight out of the run's own trace:

```
── turn 10  finish=tool_calls  in=4478 out=61 cached=4352 hit=97% ──
call sh {"cmd": "cat /tmp/af-abir/610-run/scratch2/note-10.txt"}
  → 40B: file 10: alpha beta gamma delta line 10⏎
no-progress guard: 10 consecutive tool-calling turns with no workspace mutation — result asked for
wall pace — 18s gone, 12s left
```

and `deadline close — landing reserve started` at the end. The control at a 60s wall crossed no half and paced nothing. Two runs, $0.0027. I also could not reproduce the flakiness I expected: the two real-clock tests at `-count=10` while `./internal/tui3` loaded the box came back `ok 107.461s`, 10 for 10.

That same line is the problem.

### That leaf was doing exactly what it was told

The brief said one read per turn, do not write any files, and the deliverable is the answer. At turn 10 it was told it had made "10 consecutive tool-calling turns without changing the workspace. Produce the result now." It ignored the notice and read for eight more turns until the wall — so on this evidence the advisory span bought nothing, and had the wall been longer it would have hit turn 20 and been concluded.

The predicate (`noprogress.go:339-350, 403-411`) is: a turn with at least one tool call, `mutationsAfter == mutationsBefore`, no write-tool call, increments `reconTurns`; any mutation resets it; ten pace, twenty conclude, then four landing turns terminate. **A leaf whose deliverable is its answer never mutates anything, by design.** A research errand, a review, a question answered out of a repository — these are not stalls, and this counts them as one. The long build and the slow provider both survive (one command is one turn; recon counts turns, not time), so this is the one shape that does not.

I am not asking you to drop the guard. I am asking for the case to be *distinguished* — a leaf that has been asked for an answer rather than a change should not be walked up this ladder, and if the signal for that is not available where the guard stands, then the guard should not escalate past the advisory for it. As it stands the PR's own e2e is a leaf being told to hurry up while doing its job correctly.

### The manual says the opposite of what the code does

`adaptive-runs.md:917` — **"Reading is never by itself a reason a worker is stopped"** — and the same claim in the change entry and the PR body's "no new way for a leaf to die". Your own test asserts the reverse: `pace_test.go:186` `TestALeafThatOnlyReadsIsConcludedThoughEveryResultIsNew` scripts a leaf whose every turn is a distinct read and asserts it stops `StopNoProgress`. The qualifier is in the next sentence, but the bolded one is the retrievable claim and it is false as written. This is the manual law's exact failure mode: a page that is confidently wrong in the words somebody searches.

### PERF.md carries the identifiers without the numbers

`PERF.md:1704-1706` names `noProgressReconTurns`, `2 × noProgressReconTurns` and `wallPaceAt` and gives **no value for any of them**, while every neighbouring row in the same table spells its figure — `noProgressTurnFloor` (60), `maxTurnBackstop` (400), `cachedTokenWeightPercent` (10), `reuseCeiling` (240,000). `PERF.md:20` is "Any change to a number below must change this file in the same commit", and 10 and 0.5 are nowhere in the file. The one-second display rounding at `linear.go:1720` is undocumented too.

Worth saying clearly, because it cuts the other way: **no number is spelled twice in the code.** Every production and test reference interpolates from the constant, and `noProgressReconConclude = 2 * noProgressReconTurns` is derived. The only literal copies are in the changelog's prose, which is a record rather than a source. So this is the reverse of drift — the doc has the names and not the figures, which is the one thing that table exists to hold.

### Smaller

- **"A mutation of any kind" is overstated.** `produced.go:165-174` snapshots only `w.root`, skips dot entries and dependency and cache trees, and diffs endpoints. A write outside `--dir`, into a skipped tree, or a create-then-delete inside one shell call is invisible to `MutationCount` and does not clear the count. The changelog line should say what is actually watched.
- **`TestAProducingLeafIsNeverPaced` (`pace_test.go:159`) passes without the production change** — it asserts `StopDone` plus the absence of a string that does not exist on dev.
- The two real-clock tests cost `./internal/exec` **+11.5s, 22.357s → 33.940s, a 52% increase**, essentially all sleep. Precedent exists (`linear_test.go:136` on dev already sleeps), so this extends a pattern rather than inventing one, and the `-race -count=2` run was clean — but it is worth knowing what the package now costs everyone.
- The new section spells the flag `-timeout` while `linear.go:1716` and the PR body say `--timeout`.
- Untested claims in the body: no test shows an answer-only leaf making more than twenty distinct reads and then *succeeding*; "a leaf already told to land is never paced" is covered for ordering but not for suppression when landing was already active.

### Confirmed clean

One `## ` heading added (`adaptive-runs.md:902`), none renamed, and it is a retrieval magnet — but nothing broke: the probe table passes embedded and packed, and the three probes I would have expected to lose still land correctly. `how-tasks-run.md:1012`'s `no_progress` reset describes the session-task counter, a different subsystem — not falsified, though the vocabulary now collides.

### One thing for whoever lands this

This PR and #607 both edit `internal/exec/linear.go` and `internal/manual/chat/adaptive-runs.md`. `linear.go:816-834` (the pace injection between the deadline-landing check and `readSteering`) and `linear.go:1262-1286` (the `progress.observe` switch) are ordering-sensitive, and a silent mis-merge there changes when the guard fires. The `adaptive-runs.md` insertions are close enough to conflict textually. Whichever lands second wants a real read of the conflict, not a retry.

santoshkumarradha:
Coordinator's ruling on the false positive, and it comes with a sharper ask than my review had — because reading `observe` again, the machinery you need is already sitting in the same function.

**The ruling: this is a law violation rather than a tuning question.** aforge is general-purpose, and a leaf whose deliverable is its *answer* — a research errand, a review, a question answered out of a repository — works by reading and mutates nothing by design. Concluding such a leaf for the absence of mutation is not a threshold set too low; it is the wrong predicate.

**What the guard may do:** advise at ten turns as it does now, and **conclude only on repetition** — a tool-calling turn whose call and result the leaf has already seen, same call and same content. Never on the absence of mutation. A fresh read is progress.

**And you have already written that.** `noprogress.go`'s Signal 1 is exactly it (`lastCallKey` / `lastResultKey` / `repeatCount`, with errors excluded because a retry is troubleshooting rather than spinning), and Signal 2's `stagnantTurns` already resets on `newInfo` for the same reason. Signal 4 is the new one, and it is the only one that deliberately looks past that — your own comment says so:

```go
// Signal 4: mutation-free recon. A tool-calling turn advances the count
// whenever it leaves the workspace unchanged, even when its result is new;
```

"Even when its result is new" is the whole defect in five words. The other three signals already agree that new information is progress; this one overrules them. Resetting `reconTurns` on `newInfo` as well as on `mutated` would keep everything the escalation ladder does for a leaf genuinely going in circles — Signals 1 and 2 catch that leaf and catch it earlier — while a leaf reading twenty different files stops being counted as stalled.

`pace_test.go:186`'s `TestALeafThatOnlyReadsIsConcludedThoughEveryResultIsNew` inverts under this: a leaf whose every turn is a *distinct* read should reach `StopDone`, and its name should say so. That test is currently the clearest statement in the tree that the behaviour is intended, which is why it has to move with the code.

The three record fixes from my review come with it, plus two small ones:

1. `adaptive-runs.md:917` — "Reading is never by itself a reason a worker is stopped" — becomes true under this change rather than needing rewording. Worth keeping the sentence and making the code match it, which I think was the intent all along.
2. `PERF.md:1704-1706` names `noProgressReconTurns`, `2 × noProgressReconTurns` and `wallPaceAt` with no value for any of them, where every neighbouring row spells its figure. `PERF.md:20` puts the number in that file. The one-second display rounding at `linear.go:1720` is undocumented too. (Credit where it is due: no number is spelled twice in the code — every reference interpolates from the constant and `noProgressReconConclude` is derived. The gap is the reverse of drift.)
3. "A mutation of any kind" overstates what is watched: `produced.go:165-174` snapshots only `w.root`, skips dot entries and dependency and cache trees, and diffs endpoints, so a write outside `--dir`, into a skipped tree, or a create-then-delete inside one shell call never clears the count.
4. The new section spells `-timeout` where `linear.go:1716` and the body say `--timeout`.
5. `TestAProducingLeafIsNeverPaced` (`pace_test.go:159`) passes without the production change — it asserts `StopDone` plus the absence of a string that does not exist on dev.

Everything else in this PR is good and the wall pacing is not in question: I proved both mechanisms in one real run, the `-race -count=2` pass is clean, and I could not make the two real-clock tests flake at `-count=10` under a loaded box.

**On landing.** #607 goes in first (it and this PR both touch `internal/exec/linear.go` and `internal/manual/chat/adaptive-runs.md`, and `linear.go:816-834` and `:1262-1286` are ordering-sensitive). Once it does, **I will rebase this branch onto it and push the rebase only** — I wrote the #607 side of every one of those conflicts, so resolving them is mine rather than yours. The mechanism change above stays with you, and this review stays open until it lands.

AbirAbbas:
Ruling taken as written. The mechanism is changed, not tuned: **no count of mutation-free turns can conclude a leaf any more, at any count.** Three commits on top of the rebase, one per ask.

## The mechanism — `74bbd216`

Signal 4 keeps the advisory and loses the escalation entirely. `noProgressReconConclude` is deleted, the `reconTurns >= …` conclude branch is deleted, `reconNoticeDue` is deleted, and the `case g.reconTurns >= noProgressReconConclude: return g.reconReason()` arm of `noProgressReason` is deleted — so a recon count can no longer appear as the reason a leaf was stopped, because there is no path that stops one.

I did not take the suggested edit literally, and this is the one place I diverged, so it is worth stating plainly. Resetting `reconTurns` on `newInfo` as well as on `mutated` makes it a copy of `stagnantTurns` — same increment, same reset, and the stagnant cap fires at 6 — so the advisory at ten could never be reached again and Signal 4 would be dead code wearing a comment. The ruling's other half was "advise at ten turns as it does now", so what I kept is the counter and what I removed is the only place it could conclude. The behaviour matches the ruling exactly; the counter reads what it always read and nobody may act on it but the notice.

What that means for the two leaves in question:

- **A leaf reading twenty different files** is asked for its result at 10, asked again at 20, asked again at 30, and finishes when it finishes. `TestALeafWhoseEveryReadIsFreshIsNeverConcluded` — the old `TestALeafThatOnlyReadsIsConcludedThoughEveryResultIsNew`, inverted and renamed as you asked — scripts thirty distinct reads and asserts `StopDone`, `Exhausted` empty, `outcome.Turns > 2 × noProgressReconTurns`, no conclude directive anywhere in the transcript, and neither `conclude directive injected` nor `no-progress` in the run's record. It is the reviewer's false-positive shape and it also closes the untested claim you flagged: an answer-only leaf making more than twenty distinct reads and then *succeeding*.
- **A leaf genuinely going in circles** is still concluded, and by the predicate the ruling names. `TestALeafRepeatingItselfIsStillConcluded` runs the whole ladder twice through `Linear.Run`: once on the same call returning the same result (Signal 1, at 4), once on turns that keep returning content already seen (Signal 2, at 6). Both end `StopNoProgress` with the right cause in the record.
- **`TestProgressGuardReconCountNeverConcludes`** feeds the guard fifty mutation-free tool-calling turns and asserts every verdict is `continue` except `pace` at each multiple of ten — never `conclude`, never `terminate`.

I checked these are not vacuous by re-adding the concluding branch and re-running: `TestProgressGuardReconCountNeverConcludes`, `TestAReadOnlyLeafIsAskedForItsResultAndAskedAgain` and `TestALeafWhoseEveryReadIsFreshIsNeverConcluded` all go red, on the right assertions. Reverted, green again.

One person-facing string moved that you did not ask for, and I think it has to. The advisory's trace note read `no-progress guard: 10 consecutive tool-calling turns with no workspace mutation — result asked for`. A leaf reading twenty different files is making progress; a line beginning "no-progress" tells whoever reads that run the opposite of what the code now believes. It is now `reading pace — 10 consecutive tool-calling turns with no workspace mutation, result asked for`, which pairs with the `wall pace —` note beside it. The sentence the *model* sees is byte-for-byte unchanged — that is the one your run quoted, and it already offers the answer-in-the-reply path.

## `TestAProducingLeafIsNeverPaced` passing without the production change — `74bbd216`

Correct, and thank you: it asserted the absence of a string that does not exist on `dev`. It now carries its own read-only control in the same test — same helper, same length, same brief shape — which asserts the notice **is** there with `itoa(noProgressReconTurns)` in it. The positive half cannot pass without the production code, so the test fails on `dev` for the right reason.

## The suppression case — `74bbd216`

`TestALeafAlreadyLandingIsNeverPaced` is new: a leaf whose budget landing began at the foot of turn 1 crosses half of a two-second wall at the top of turn 2 and receives no clock reading. It asserts the budget landing message *is* in the transcript (so the landing really was active), that no message contains the clock reading, and that the record holds `budget exhausted — landing reserve granted` and no `wall pace —`. Ordering was already covered; suppression is now covered too. ~1.2 s, no new sleep beyond that.

## The manual — `2379559a`

Kept the sentence and made the code match it, as you read the intent. **"Reading is never by itself a reason a worker is stopped."** is now simply true, and the paragraph after it says what *does* stop a worker — repeating itself: the same call returning the same answer, or turns that keep coming back with nothing the worker has not already seen — and that a worker whose result is the answer itself may read for its whole time and finish by giving that answer. The paragraph before it no longer describes a second span, a conclusion, or "a few last calls"; it says the reminder is a reminder, nothing is stopped and nothing is taken away, and it comes again with the larger count if the reading goes on. The heading is unchanged, and the probe table passes embedded and packed.

`--timeout` where the section had `-timeout`, matching `linear.go` and the rest of the corpus. I left the neighbouring heading at `:903` alone — it is a spelling somebody types into search, not a claim, and `commands.md` already says either spelling is accepted.

Regenerating the packed corpus surfaced one collision I would not have predicted: the phrase "several turns" in the new paragraph outranked the tasks page for *"why did one task turn into several workers"*. Reworded to "turn after turn"; the probe lands correctly again.

## `PERF.md` — `2379559a`

Every surviving row now spells its figure: `noProgressRepeatCap` (4), `noProgressStagnantCap` (6), `noProgressTurnFloor` (60), the reminder at `noProgressReconTurns` (10), `wallPaceAt` (0.5), and "read to the second" for the display rounding at `linear.go`. The escalation row is deleted with the escalation. The paragraph under the table no longer argues for a fourth ceiling — it says the absence of a write is not evidence of a stall, and that the repeat and stagnant signals reach a circling leaf at 4 and 6 turns.

Your credit on the other side is preserved in the code: every reference still interpolates from its constant, and the two tests that had used the derived `noProgressReconConclude` now say `2*noProgressReconTurns` rather than a literal 20.

## The change entry — `e7abaafb`

Rewritten. "A mutation of any kind" is gone; what is there instead is what `MutationCount` can actually see — a per-leaf revision that moves when a write tool files an artifact or when a bounded before-and-after walk of the leaf's **own** workspace root differs, a walk that skips dot entries and dependency and cache trees, stops at 6000 entries, and compares endpoints. So a write outside `--dir`, a write into a skipped tree, and a create-then-delete inside one shell call never clear the count. The title said "reading forever stopped counting as progress", which is now false; it says the reminder never stops a leaf.

## The rebase, and the merge you offered to take

You offered to do this one and I have taken it out of your hands, so here is exactly what happened, in case you want to redo it. When I fetched, `origin` held your rebase onto #607 at `e9cf745b`. Six things have merged into `dev` since (#622, #589, #608, #605, #607 …), so I rebased onto `dev`'s current head `b3922479` — **clean, zero conflicts**, and `linear.go`'s ordering is intact: the deadline landing check, then the pace injection, then `readSteering`. Before pushing I byte-compared the two patch sets — `git diff 85ceded8 origin/fix/…` against `git diff b3922479 <my rebase>` — and they are **identical**, so your conflict resolutions are preserved exactly and nothing of yours was discarded. The push was `--force-with-lease` pinned to `e9cf745b`.

## What I did not verify

- **No real-model run.** Yours is the only end-to-end evidence either mechanism has. `OPENROUTER_API_KEY` is not in this environment, and the shape you would want re-run — an answer-only leaf reading past turn 20 and finishing on its own — is exactly the run your review produced, so it is worth thirty seconds of your `deepseek-v4-flash` to see it now finish instead of being told to hurry.
- **The advisory now repeats without bound.** A leaf reading for 400 turns hears the reminder forty times. It is one sentence of about fifty tokens each, and it is the only pressure left in the system on a leaf that ought to be writing — but it is a cost that did not exist when the second span ended the run, and I would rather name it than have you find it. If you want it capped at two spans, say so and it is a two-line change.
- **The package's cost is unchanged from your measurement** apart from the new landing-suppression test's 1.2 s. `./internal/exec` runs at 44 s here.

Gates, from the worktree on the rebase: `go build`, `go vet`, `make fmt-check`, `go generate ./internal/manual` + the packed-manual test + a clean `git diff --exit-code`, `go test ./internal/manual/`, `go test -run Manual ./internal/tui3/ ./internal/session/`, `make test-laws` (49 files, 14 packages), `make test PKGS=./internal/exec` (44 s), the untagged `./internal/e2e/` word gate, `go test -count=2 ./internal/exec/`, and `aforge-changes check` — all green.

Issue #609, opened 2026-09-03
checks re-run the whole suite from scratch and no check knows how much wall is left, so a task finishing with less than two suite lengths to go can never end done


santoshkumarradha:
## Status (checked 2026-09-12)

**Landed**
- Session-side memo + wall-bounded skip: [PR #633](https://github.com/Agent-Field/aforge-v2/pull/633) (`internal/session/checkmemo.go`, `wallBoundedWindow`, `Unread`). Change entry: `docs/changes/unreleased/633-green-check-not-rerun-at-the-wall.md`.
- Separately, checks binding onto the task copy (wrong-checkout half of the old story): [PR #911](https://github.com/Agent-Field/aforge-v2/pull/911) / closed #886.

**Still open**
- Content-identity across restore ↔ deliverable trees so a memo can hit when roots differ but bytes match — called out as still owed in the #633 change entry.

---

## What happened

The #546 DOE run by the canary, 2026-09-03, builds `b07c3dd51` (share 3) and `15a2b6ef2` (share 2), tables on #407. Cells reef / attrs / tox, chat door, `aforge chat --yolo -one-model -model deepseek/deepseek-v4-flash -max-hours 0.25 -max-cost 1.0` — a 900 s wall. In 3 of the 6 walled cells the task had finished and the check reading it was still running when the wall came down. The plainest row, `Human-Agent-Society-reef-145` under `turnWallShare=3`:

```
task_done_s 588 · done_to_wall_s 315 · wall 903s · $0.0853 · carry_ons 6 · reason "wall · tests green"
```

Tests green, work finished at 588 s, 315 s of wall left, and the run still ended on the wall.

The suite on these projects takes about 150 s, and it is run **from scratch by each reader of the same tree**: once by the baseline reading at session start, once by the task's own auditor, once more by the session's terminal check on whichever ending road the run takes. Same command, same tree, no reuse between them. A task that finishes with less than two suite lengths of wall left cannot reach a done row whatever it did, because everything that reads its result starts the suite again.

## Replication

[...]

PR #633, opened 2026-09-04
session: a green check is not re-run at the wall, and none is started that the wall cannot hold

AbirAbbas:
Session baseline and terminal checks now reuse a completed answer only for the same command and unchanged tree, with a bounded 32-entry memory. Fresh session checks are sized by remaining wall time and an unstarted check is reported as unchecked rather than red. Cancellation or process-start failure supplies neither a reusable answer nor a complete execution time; review added a regression for the prior cached-cancellation failure.

The branch is refreshed against current dev. Original focused check/cache/ending tests and the new cancellation regression pass, with manual retrieval tests. The review follow-up documents the memory bound in PERF.md.

This only partially addresses #609. Task-checker results are still not shared across restored worktrees and the final deliverable tree; doing that requires an appropriate content identity and shared receipt. Task audits also retain the existing reading-time floor when the session has less wall remaining. Those original acceptance items remain outstanding. No new live-model canary was run.

Held for secondary review. Related original PR implementation and local exact-head regression evidence are preserved in the audit report.

The process helper preserves its original task-baseline call contract for #611, while session callers pass an explicit wall-sized window. Compatibility/check-memory/manual regressions pass on the final head.


santoshkumarradha:
Coordination with draft #688: your check reuse directly affects cost/time, and your `checkpoint.go` changes touch the same end-of-turn flow. #688 distinguishes a reporting turn waiting on its existing request owner from final goal completion. On that wait path, do not run/reuse terminal checks or label the global request done merely because the current report ended. On a genuine goal-ending path, preserve your same-command/unchanged-tree receipt contract and keep unstarted checks unknown.

Please include both controls after rebase: background receipt with a live owner produces no duplicate task or terminal-check spend; a genuinely finished goal still performs its required fresh-or-valid-reused checks. Our Spark cohort is frozen and excludes both branches; future timing comparisons must identify the exact combined commit. #688 is draft and not yet performance validated. No merge requested here.

santoshkumarradha:
The completion path is being simplified in #741 (https://github.com/Agent-Field/aforge-v2/pull/741). `decideRemains` now consults the principal once after actual declared checks; `decideOverTheChecks` returns `Decision`, and `terminalAudit` returns checks/stash state without implicit file cleanup. Please adapt check reuse/remaining-wall logic to that path without restoring a completion reader, file-change witness or provisional done decision. The last point matters: deciding done before running a failing check resets the repeated-failure fingerprint. Existing budget-stop precedence and stopSpent remain required.

PR #667, opened 2026-09-09
chat: the status row's state word is painted from the reading it was measured with

santoshkumarradha:
## What was wrong

Every width decision the status row makes is made from the **plain** cluster —
the gap between the identity and the telemetry, the right-aligned row's indent,
and the columns the keeping and money doors are pressed on (`standdoor.go`,
`moneydoor.go`). The painted cluster was never held to the same measurement:
two segments rebuilt their own text inside `paintPart` instead of painting the
text the row had been laid out against, and one of them — the state word —
rebuilds it from `app.now`, which is `time.Now()` in the running program.

So the layout's reading of the turn clock and the paint's are two reads of a
moving clock a few microseconds apart. A turn crossing `9s` into `10s` between
them is drawn one cell wider than it was measured:

```
"…· ⟲ 42% cached · 160 tok/s avg · YOLO · ⠋ working · 10s"   211 cells on a 210-column frame
```

The renderer under us composes into a cell grid exactly the terminal's width,
so the row is not wrapped — it is **cut**. The last cell goes: `⠋ working · 10s`
is drawn `⠋ working · 10`. Confirmed in a real tmux pane with a stub bubbletea
program: a 61-cell bottom row on a 60-column pane drew `tick-002` as `tick-00`.
The keeping and money doors are meanwhile a column left of where they look.

## What is true now

The state word hands its painting over on the `hudPart` it was assembled into,
so the text and the painting are one reading of the clock. The phone deck's own
state segment is built the same way for the same reason.

## The test

`TestAClockThatTicksMidFrameCannotPushTheStatusRowPastTheFrame` walks the
crossing through every pair of reads the row makes, at five widths, and holds
both halves of the law: no row wider than the frame, and a painted cluster that
measures exactly what the row was laid out against. It fails on `dev` (211 cells
at 210 columns) and passes here.

## Scope

This came out of an investigation into a reported **doubled status row** (a
stale HUD row stranded one row above the live one). It is **not** that defect —
an over-wide row is cut by the renderer, not wrapped, so it cannot scroll the
alt screen. It is a real defect found in the same code, with its own reproduction.
The doubled row is still open; what was ruled out is in the investigation notes.


santoshkumarradha:
The gate's only red is inherited from `dev`, not caused here.

`TestCrewAppliesAPresetAndConfirmsInOneLine` and `TestStatusNamesTheCrewUnderTheModel` both fail on a clean detached worktree at `origin/dev` (5376c1e23, #665 "Leave model generation controls to providers by default") with no changes from this branch:

```
--- FAIL: TestCrewAppliesAPresetAndConfirmsInOneLine
    crew_test.go:150: the confirmation is missing "brain kimi-k3:high"
--- FAIL: TestStatusNamesTheCrewUnderTheModel
    crew_test.go:324: the crew line lost "brain kimi-k3:high"
```

The whole `./internal/tui3` suite is green at this branch's original base (d40d3e4ed, #664), so the reasoning rider stopped reaching the crew line with #665. That is #665's lane to settle — either the line owes the rider or the tests owe an update. The light gate here is green and nothing else in the touched package is red.

PR #679, opened 2026-09-09
Task states: the chat manual learns the three tiers and one question

santoshkumarradha:
Lane M of the task-states wave. The engine contract landed in #668 and the automatic attempts in #673; this is the corpus that tells the chat — and therefore the person asking it — what the new words mean.

**What was true in the pages.** They spelled a task's state five ways, because the surfaces did: `needs your look` on the card, the roster and home; `awaiting review` after a handoff to the chat; `failed` under a run that lost its connection; `unverified` for files nobody built; `delivery needs attention` on an expanded card. Two pages read side by side gave two vocabularies, and the chat — whose only account of this program is these pages — said whichever it was handed. The answers row was documented as `[a] accept · [l] look again · [n] not right · [d] decide these for me`, where `l` offered a check the engine had already run twice and `d` quietly flipped a standing setting.

**What is true now.** Every task row answers one question before anything else — *do I need to do anything* — and the pages say so: **moving · over · your call**, one glyph each, with the reason always beside the word (`waiting on task 4`, `incomplete · ran out of steps`, `your call · conflicts with your branch: parser.go`). `failed` is gone from everything a person reads, and `incomplete` carries the reason table out of `TaskReasonOf`, ending by ending. The six `your call` questions each carry their own two verbs, so `[a]` is always yes to what the row is asking and `[n]` always no — and the page says which three of the six a **landing card** asks, because a proposal, a design and a spend gate ask the other three from their own cards. `[s] tell it` opens the task's page rather than answering, so "looks good" typed on a card is not an accept; `[d] let aforge decide this one` hands over one card and changes no setting; `[t] take it back` is the way out of `aforge is deciding`; and the auto-settle floor — a task never stays unowned past the end of a turn — is why a card can never go quiet. A conflict is never the model's to accept, said where the conflict is described. The three things the engine tries first (a second check on another model, one merge round, one rerun from the branch) are written out, which is why no `check again` chip exists.

**Pages changed:** `tasks.md`, `how-tasks-run.md`, `home.md`, `screen.md`, `places.md`, `reading-a-task-page.md`, `task-rooms-after-restart.md`, `keys.md` (one sentence).

**Headings added** in the asker's own words: *the three tiers, one glyph each … and what happened to finished but needs your look* · *why does it say incomplete — what incomplete means vs stopped* · *what your call can be asking — the six questions* · *how do I accept a task* · *the task has a conflict* · *tell it something instead of answering* · *why did it stop asking me — it says aforge is deciding, and how do I take a task back* · *why is there no check again — what the engine tries first* · *the four words a task can land with* · *why does it say not accepted under my task* · *being quiet is not going in circles*.

**Twelve new probes** in `internal/manual/chat_test.go`. Two oversized sections were split so the question about a refused check reaches the page that answers it instead of the page about accepting.

**`keys.md`'s answers-row section is left to #689**, which rewrites it beside its own code; this branch keeps only the one sentence there that described a state word which no longer exists. `internal/e2e/tuiwords_test.go` is likewise #689's — nothing here needed reconciling, and the untagged `go test ./internal/e2e/` gate is green.

**Where the code still spells an old word, the page follows the code:**
- `internal/session/task_audit.go:397` — a landing REPORT still opens `finished, but needs your look — `, so `how-tasks-run.md` quotes it exactly while the row above it says `your call`.
- `internal/session/principal_wire.go:376` — `waitWord` still returns `needs your look` (pinned by `principal_audit_test.go:1137`), so `starting-aforge.md` keeps it.
- `internal/session/taskgrade.go:426` — the ratings file a person may open still writes `needs your look`, so `models-and-cost.md` keeps it.
- `internal/tui3/homestanding.go` — a **standing order** waiting on a person reads `needs your look` on `dev`; `standing-orders.md`, `keeping-an-eye.md` and home's standing lines were left alone. (`feat/task-states-rail` moves this to `your call`; when it lands those three want one more pass.)
- No `rerun from its branch` chip is documented: there is no such verb, and the rerun the engine spends is automatic.

**What lands with the surface lanes.** `feat/task-states-card` (#689) is open and this branch's card prose is written against the strings it actually spells. `feat/task-states-rail` has not opened a pull request; the roster's `your call` section, home's `? your call` / `✗ incomplete · a fault` rows, the task page's header word and the roster hint are described as that branch already spells them.

Gates: `go test ./internal/manual/ ./internal/tui3/ -run 'Manual|Corpus|Probe'`, `go test ./internal/e2e/`, `go test ./internal/session/ -run Manual`, `make test-laws`, `make test-packed-manual` — all green.


santoshkumarradha:
Manual coordination with draft #688: both branches edit `how-tasks-run.md`. Please retain the new section “A background job finishes while a task is still working on my request” when resolving the overlap, while applying this PR's authoritative task-state vocabulary.

The distinction to keep is that the conversation may finish a reporting turn while the task and overall request remain unfinished. A job receipt must not start a duplicate owner; task results may still require repair/integration, and new human requests remain parallelizable. Please avoid describing that waiting turn as task completion or a `your call` decision. #688 remains draft until tests and benchmark review finish.

PR #683, opened 2026-09-09
provider: a check that answers while the send keeps failing is a fault, and a fault waits

### LINKED ISSUE #675: provider: a check that answers while every send fails spins the request without a pause, and the media road's wait names no model so nothing is drawn
## What happened

Two roads into the same recovery primitive, and both of them spin.

A person asks for a picture while the wifi is going in and out. The tool row
sits there with nothing on it — no `waiting for connection`, no count-up, no
figure at all — for up to two minutes, and underneath it the process is sending
requests as fast as the transport can refuse them.

On `dev@5ffb15b1`, a twenty-second media fixture (HEAD answers, every POST
fails with a DNS error, the reachability check answers on its third try):

```
B7 posts=3728757 probes=3728759 phase_total=3728759 named_model=0 blank_model=3728759 err=context deadline exceeded
```

Every one of those 3.7 million phase reports carries an empty model, and
`tui3.PostPhaseNews` drops a phase for a model nobody named — which is why the
screen shows nothing at all while the machine is at full tilt.

The chat road has the same shape without the invisibility. A two-minute fixture
whose reachability check always answers and whose every POST fails pre-send:

```
B25 posts=8573814 inferred_calllog_rows=17147628 median_gap_ms=0.009560 elapsed=2m0.001s
err=connection is still unavailable; try again when connected
```

Eight and a half million sends and seventeen million model-call log rows in two
minutes, at a gap of ten microseconds.

## Replication

**Deterministic (no model).** `go test -run TestConnectionRecovery -count=1
./internal/provider/`, with a fixture in that package built like this:

- *media:* `NewMediaClient` over a `RoundTripper` that answers HEAD 405 and
  fails every POST with `&net.DNSError{IsNotFound: true}`; set
  `client.connection.connectionProbe` to fail twice and then return nil;
  register `OnPhase` and count the news; call
  `doEndpoint(ctx, http.MethodPost, base+"/images", body, true)` under a
  twenty-second context. Today: millions of posts, every phase news with an
  empty `Model`.
- *chat:* `NewClient` over a `RoundTripper` that fails every POST the same way;
  set `client.connectionProbe` to return nil; call `CompleteWithMessages`.
  Today: millions of posts inside the two-minute window, with no pause between
  passes.

Either fixture with `client.wait` replaced by a recorder shows the same thing
in under a second: no delay is ever asked for between one failed send and the
next.

**Field (real models).** A captive portal or a transparent proxy — a café or a
hotel network that answers a HEAD on the origin instantly and drops the POST —
is the real shape of it. Needs `OPENROUTER_API_KEY`; ask for an image, then
join the portal without signing in. No money is spent: nothing is ever
accepted.

## Where

- `doEndpoint` in `internal/provider/media.go` — both `waitConnection` calls
  pass an empty model (search `waitConnection(ctx, ""`), and the loop's
  `continue` after a successful wait has no pause in front of it.
- `send` in `internal/provider/retry.go` — search `attempt--`. It cancels the
  loop's own increment, so `attempt` never grows, `outOfPatience` is guarded by
  `attempt > 0` and never fires, and `reconnected` skips the backoff on the way
  round.
- `probeConnection` in `internal/provider/connectivity.go` is what makes it
  reachable in the field: any HTTP response proves reachability, 401 and 405
  included, which is correct and is exactly what a portal answers.
- `PostPhaseNews` in `internal/tui3/phase.go` drops a phase whose `Model` is
  empty, which is also correct — a phase for nobody belongs to nobody.

## The fix

One law for both roads: **a send that failed before it left waits with a pause
that grows, and the wait says who it is waiting for.**

- The media road hands its model to `waitConnection`, so the surface draws
  `waiting for connection` against the model that was asked for, exactly as the
  chat road already does.
- After the first recovered pass — which still resumes immediately, because a
  connection that came back should not be punished with an old delay — every
  further pass waits `backoffFor` its own count before re-sending. A
  reachability check that answers while the request still cannot go out is a
  fault, and faults back off.
- The passes are counted and capped as the arithmetic backstop for the
  degenerate case where every wait is instant. The two-minute recovery window
  is still what ends it in practice, and the refusal is still
  `connection is still unavailable; try again when connected`.

Nothing about routing budgets, pins, model defaults or the probe's own cadence
changes.

## Acceptance

- **e2e:** a picture asked for over a dead connection draws
  `waiting for connection` on the tool row within a second, counting up, and
  the count-up is against the model that was asked for — not a blank row.
- **e2e:** the control — a healthy picture request makes no reachability check
  at all and its numbers are unchanged.
- **Unit:** the twenty-second media fixture above sends fewer than 100 requests
  and posts fewer than 100 phase reports, every one of them naming the model;
  the two-minute chat fixture sends fewer than 100 requests and writes two
  model-call log rows per request; the delays asked for between passes strictly
  grow; a caller cancelled mid-pause ends with the caller's own error.
- **Unit:** the existing
  `TestConnectionRecoveryProbesWithoutGenerationAndResumes` numbers are
  unchanged — one failed send, one accepted generation, three probe waits.
- The manual page that covers this quotes the new wording
  (`internal/manual/chat/models-and-cost.md`,
  `internal/manual/chat/when-the-connection-drops.md`), and the change entry's
  `invalidates` names what people believed before.

AbirAbbas:
## What was true

A reachability check that answers while the request itself still cannot go out —
a captive portal, a transparent proxy, a network that resolves nothing but
answers a HEAD — put both roads into `internal/provider`'s connection recovery
into a loop with nothing at all between one send and the next.

`send` in `retry.go` cancelled its own increment (`attempt--`) and set
`reconnected`, so nothing grew and nothing waited. `doEndpoint` in `media.go`
simply `continue`d. And the media road passed an EMPTY model to
`waitConnection`, so every phase report it made was dropped at
`internal/tui3`'s door — a phase for a model nobody named belongs to nobody.

Measured on `dev@5ffb15b18`, with the fixtures from the report:

```
B7  media, 20s   posts=3728757 probes=3728759 phase_total=3728759 named_model=0 blank_model=3728759
B25 chat,  2m    posts=7975780 inferred_calllog_rows=15951560 elapsed=2m0.001s
                 err=connection is still unavailable; try again when connected
```

## What is true now

One primitive, `recoverBeforeSend` in `connectivity.go`, is what both roads call.
It holds the call's own two-minute window, counts its passes, and waits
`backoffFor` its own count before every pass after the first. The first
recovered pass is still immediate — a connection that genuinely came back should
not be made to wait out an old delay — and connectivity still spends none of the
provider retry ladder. `connectionRetryPasses` is the arithmetic backstop for
the degenerate case the window cannot bound, where the wall clock barely moves
and only a count is finite.

The media road hands over the model it is drawing, speaking, filming or
transcribing with, so the wait is attributed to the model that was asked for.

The same two fixtures, on this branch:

```
B7  media, 20s   posts=6  probes=8  phase_total=8  named_model=8 blank_model=0
B25 chat,  2m    posts=16 inferred_calllog_rows=32 elapsed=2m0.013s
                 err=connection is still unavailable; try again when connected
```

## Validation contract

1. When a media request cannot be sent because DNS or the dial failed, every
   phase report the wait makes names the model that was asked for.
2. When the check answers while the request still cannot be sent, the next send
   waits, and the pause grows with each further pass; the first recovered pass
   is still immediate.
3. A media request failing pre-send under a twenty-second deadline sends fewer
   than 100 requests and posts fewer than 100 phase reports.
4. A chat completion in that state sends fewer than 100 requests inside the
   two-minute window and writes exactly two model-call log rows per send.
5. Recovery still ends, and still says
   `connection is still unavailable; try again when connected`.
6. A caller cancelled during the pause ends promptly with its own error.

**Unchanged, and asserted:** a healthy request makes no check;
`TestConnectionRecoveryProbesWithoutGenerationAndResumes` keeps `posts=2
heads=4 waits=3`; an accepted media job is never resubmitted; the probe cadence,
the two-minute window, the certificate escape and the shared-wait semantics; and
no routing budget, pin or model default is touched.

## One thing this does not do

The wait now NAMES its model, which is what `internal/tui3`'s `PostPhaseNews`
requires before it will keep a phase at all. It does not make the surface draw
it: `livePhase` in `internal/tui3/phase.go` reads the phase of the window's OWN
model, deliberately — "only this window's own work is drawn on this window's
row" — and a picture is drawn by a different model from the one the conversation
talks to. Making a connection loss cross that line is a surface decision with a
stated law on both sides of it (the same rule keeps a task page from showing its
parent's connection state, per `docs/changes/unreleased/653-inline-wait.md`), so
it is not taken here. The resource storm and the attribution are.

Fixes #675

## Gates

| Command | Last line |
| --- | --- |
| `go build ./...` | *(no output)* |
| `go vet ./...` | *(no output)* |
| `make fmt-check` | *(no output)* |
| `make test-packed-manual` | `ok github.com/Agent-Field/aforge-v2/internal/manual 1.617s` |
| `go test -count=1 ./internal/manual/` | `ok github.com/Agent-Field/aforge-v2/internal/manual 2.084s` |
| `go test -run 'Manual' ./internal/tui3/ ./internal/session/` | `ok …/internal/tui3 0.197s` · `ok …/internal/session 0.311s` |
| `make test-laws` | `ok github.com/Agent-Field/aforge-v2/internal/workspace 0.029s` |
| `go test -count=1 -timeout 15m ./internal/provider/` | `ok github.com/Agent-Field/aforge-v2/internal/provider 56.153s` |
| `make build` | `-o bin/aforge ./cmd/aforge` |

`internal/provider` is the only package with a Go file in this change; the two
manual pages are covered by the manual gates above. `internal/tui3`'s full suite
was NOT run, because no file in it changed.

## Driven in a terminal

`bin/aforge chat --no-host --yolo` against a local stub that answers the model
catalog and one chat request with `bash sleep 8` followed by `generate_image`,
and then stops listening — so the picture request meets an origin that is gone.
Both builds, same box, same stub:

```
dev@5ffb15b18   ✗ Drawing it now · failed · 2m08s · ended with an error — branch kept
                "it ended with an error: connection is still unavailable; try again when connected"
this branch     ✗ Drawing it now · failed · 2m08s · ended with an error — branch kept
                "it ended with an error: connection is still unavailable; try again when connected"
```

and the conversation's own row still draws the wait it always drew:

```
Working waiting for connection · 56s
```

That run is a TOTAL outage, where the check fails too — so it is the control,
and it shows the ending and the timing are untouched (process CPU over the same
60 seconds of waiting: 1.82s before, 1.64s after). The defect's own shape — a
check that answers while the send fails — needs a captive portal and cannot be
made in a terminal, which is why the numbers above come from the package
fixtures instead.


santoshkumarradha:
Coordination with draft #688 and the ongoing Pareto benchmark: this bounded connectivity-recovery change is relevant to time/cost attribution even though it has no ownership-file conflict. Please keep local connection recovery separate from provider retries/routing budget, keep cancellation prompt, and preserve immediate first recovery. Your existing checks appear to cover those contracts; no extra feature or delay requested.

Our comparison must label this commit separately from #688's duplicate-work correction. DNS/probe failure time is not counted as orchestration overhead, and a recovered request must not replay executed tool effects or create another task owner. We will run the connectivity controls on the eventual combined candidate rather than attributing any speed difference solely to #688. Frozen current trials remain unchanged. No merge requested here.

PR #682, opened 2026-09-09
delivery: a changed list nothing can be compared against is not proof the work landed

AbirAbbas:
## What was true

`retainedDelivery` decides whether a finished task whose branch was KEPT has
actually reached the person's workspace. It ran, from the workspace root:

```
git diff --quiet refs/heads/<branch> -- :(literal)<path1> :(literal)<path2> …
```

and read exit 0 as "the workspace already holds this work".

**Exit 0 is two different facts.** It means "every listed path is identical on
both sides" — and it also means "none of these pathspecs matched anything at
all". So a finished task whose changed list named paths that exist neither on
its kept branch nor in the workspace was reported as *delivered*: `Retained`
empty, `Delivered` true, `needsDelivery()` false, and an unattended conversation
ending on `finishing here · what was asked is done` over work still sitting
unmerged on a branch.

The list is a claim and not a receipt, which is why this is reachable rather
than theoretical: `TaskNode.finish` writes a node's leavings with no
reconciliation, and `taskTree.comeHome` throws away the paths `commitTaskWork`
actually saved. A subdirectory ground — a conversation standing in a
subdirectory of a project, whose task worktree is cut at the repository root —
spells every path against a root the delivery reading does not stand in, and a
deliverable a worker named but never wrote does the same.

## What is true now

A clean diff releases the kept branch only when at least one listed path is
really there. The two sides the diff itself compares are read with the pathspecs
the diff took — `git ls-tree -r -z --name-only refs/heads/<branch>` for what the
branch carries, `git ls-files -z` for what the workspace tracks — and a listing
counts only when it names a path that was asked about, never on mere output,
because `git` here hands back both streams. It runs only after the diff came
back clean, so the ordinary undelivered road costs exactly what it cost before.

Where nothing can be compared, the person reads the line that was already
written for it:

```
<work> is on retained branch <branch>; its changes have not reached the requested workspace
```

The safe direction is deliberate: work on a branch, reported as on a branch,
costs a person one look; work reported delivered that is not costs them the
work.

Fixes #674

## Validation contract

1. A kept branch **none** of whose listed changed paths can be found — on the
   branch or among the workspace's tracked files — stays retained, and the
   conversation says so instead of `finishing here · what was asked is done`.
2. A list spelled against a subdirectory ground (`actual.txt` where the tree
   holds `subproject/actual.txt`) is such a list, and stays retained.
3. Nothing else moves: a listed path that really exists and really differs still
   stays retained; a workspace that genuinely holds the branch's content — after
[...]

PR #662, opened 2026-09-09
Build the personal AI foundation and reliable background work

santoshkumarradha:
This draft builds the personal AI workspace using persistent logical folders, explicit governing placement, shared references and the existing chat/work/artifact owners. It stays draft against `dev`; no release, deployment or replacement of the owner's installed binary is authorized.

## Current checkpoint — 2026-09-14

Maintained head **`09c139473`** includes the journey-driven checklist and both reviewed TUI checkpoints. Checkpoint 2 runtime is **`92ba84022`**; the merge build at `218dbb086` and head build passed on Spark job `20260914-220626-000455`, with identical runtime tree `111496efc9005ba3`. The `19bb937d3` product-experience document and W5-B history are preserved. The quarantined integration tree was not used or modified.

**Checkpoint 1 is ready to try in a pinned isolated demo.** The Folders place reads the engine's actual owner records, reserves a right inspector on wide terminals, supports narrow details, opens existing chats/work/artifacts, and preserves the browsing position. A chat shared by Product and Marketing opens as the same history. Home/direct chat remain. The inspector is initially read-only.

Opus implemented and reviewed this under Spark Fleet **`20260914-192937-000446`**. Actual SSH/tmux keyboard/mouse/resize/open/return interactions at `36922486c`: **48 PASS / 0 FAIL**, ended 20:28:55 UTC. The review's special-file-preview and slow-read issues were fixed before this final run. Focused validation, known baseline failures, and deferred broad suites are documented. This is a usable demo, not full merge acceptance.

Pinned source: `/home/santosh/src/af-pai-demo-36922486c`; isolated profile: `/home/santosh/aforge-pai-demo-36922486c`. The tested launch command has been shared with the owner. These remain unchanged while development continues elsewhere.

[Checkpoint 1 record and launch](https://github.com/Agent-Field/aforge-v2/blob/codex/personal-ai-backend/docs/design/workspace-foundation/grooming/BUILD-TUI-01.md) · [Terminal acceptance](https://github.com/Agent-Field/aforge-v2/blob/codex/personal-ai-backend/docs/design/workspace-foundation/grooming/validation/tui01-accept-checks.txt) · [Product experience path](https://github.com/Agent-Field/aforge-v2/blob/codex/personal-ai-backend/docs/design/workspace-foundation/grooming/PRODUCT-EXPERIENCE-PATH.md)

[...]

PR #688, opened 2026-09-09
Keep background reports with their existing work owner

santoshkumarradha:
A background-job receipt could wake the conversation while its original request already had a live worker. Global “unfinished” feedback then became another full implementation assignment: the proxy trial ran two owners concurrently for 43m11s and delivered neither implementation before its deadline. A rewritten handoff escaped the existing task-name check.

This draft separates the reporting turn from ownership of the unfinished request. Task admission and background operations retain their originating human-request identity, scoped to the admitting session. A background-only turn reports and waits only when every receipt belongs to an existing moving root; checkpoint handoff, stopped-turn reopening and explicit proposal use the same reading. New human requests, unknown or mixed origins without matching owners, and actual task-result repair/integration remain actionable. Batched receipts retain every origin and outcome; a forced reporting turn includes the outcomes before waiting. No provider tables, semantic matching, extra model calls or generation caps.

Validation: original dev233 fails the generic receipt regression (two owners instead of one) and buys four unnecessary remainder readings for a short report. Candidate8b43e0a41 passes those regressions, mixed/unknown origin and new-user/task-result controls, focused race/steering/promotion tests, the manual suite, quick checks and native/static `make build`. All GitHub gates, including full touched-package tests, are green on that exact commit. A local full-session run passed its tests but its final state guard detected new journals from the owner's concurrent live engine; the complete rerun with isolated state passed in275.120s, with no test or known-red change.

Matched Spark rebenchmark: fixed dev233 control versus candidate8b43e0a41, writing/data/proxy seed1, one sequential worker per arm, original acceptance, same exact DeepSeek model for all roles, provider-default generation settings. Clean binary/fixture/guard hashes and offline container preflight are frozen; Fleet384 is running its launcher and waiting for both frozen baseline completion markers before starting model calls. Fleet383 failed at submission-command parsing before any cell or model call; no trial was repeated. This is not a Pareto claim; remain draft until benchmark evidence supports it. The original30-cell cohort remains frozen at5376.

Dev subsequently merged manual-only #679 (c667db69d). Retain both its task-state vocabulary and this PR's background-report section during integration; the paired experiment intentionally retains the same233 base on both sides.

Cross-PR coordination: #662 touches completion evidence and wake-time context; #633 touches terminal-check reuse; #682 strengthens delivery truth; #683 changes connectivity recovery; #679 changes task-state documentation. Preserve their independent contracts while integrating this separation between a finished reporting turn and an unfinished global goal. Detailed compatibility notes will be posted on affected PRs. No other PR is merged by this draft.


santoshkumarradha:
Holding this draft for the context-first audit requested by the owner. Passing the overlap canary alone is no longer sufficient to merge this implementation.

Review identified that originating request identity is broader than work ownership: one request can contain independent operations A and B. If main owns A and a live task owns B, matching their human request does not establish that B owns the actionable completion of A. The current request-wide guard needs that counterexample assessed before being retained.

The log audit also found smaller context defects that should be repaired first: reused provider call IDs can pair a prior result with a later command; checker input uses the shortened display report despite the whole worker result already being stored; and checker ground instructions contradict its clean restore. Two of the ID-join defects, plus an old reduced-read mislabeled as a later bash call, reproduce without a model on current dev. These require accurate context assembly, not another decision layer.

The frozen paired pilot remains useful evidence and will be preserved, but its candidate will not be changed mid-run. Reconsider or simplify this draft after the context audit; do not merge it automatically on CI or a better benchmark score.

santoshkumarradha:
The frozen paired pilot is complete. Both proxy arms timed out at 3600 seconds and failed the same two held-out checks. The general controls remain mixed: writing changed fail→pass but took 138.63s vs 123.13s; data passed both and took 256.73s vs 61.92s. Missing usage receipts prevent a complete total-cost comparison. This is not evidence of a Pareto improvement.

Keep this PR held. The context-first audit and candidate are now in #717; they repair result association, read retention, current worker conclusions, and completion evidence without the request-wide ownership inference in this PR. No source or acceptance was changed in the completed pilot.

PR #695, opened 2026-09-09
sessions: a conversation's identity is written under a lock of its own, so two windows cannot overwrite each other

### LINKED ISSUE #687: conversations: two writers on one conversation folder lose each other's meta.json, and putting one away from home comes undone
## What happened

`dev@233f729d5`, 2026-09-09. Two writers on one conversation folder — a second window
on the same conversation, an engine at the far end of `--host` beside a local surface on
a shared disk, or home putting a conversation away while its own window finishes a turn —
overwrite each other's `meta.json`. The transaction is serialized on a mutex that belongs
to ONE Agent, so a second Agent, and every second process, reads and writes the same file
with nothing between them.

Two writers, one folder, 200 rounds of "one commits the name while the other records what
the turn cost":

```
losses=200/200 (title_lost=190 spend_lost=10)
```

And the shape a person meets without any exotic filesystem — home puts the conversation
away (`e`) while its own window seals a turn:

```
archived=false spent=0.25 tokens=140
```

The conversation came back out of the archive. Nothing was said about it: `meta.json` is
written and never read back, and every failure there is deliberately silence.

## Replication

**Deterministic (no model).** Drop this in `internal/session/meta_probe_test.go` on a
clean `dev` checkout and run `go test -count=1 -run TestProbeTwoWritersOneFolder ./internal/session/`.
It fails today with the line quoted above.

```go
package session

import (
	"fmt"
	"path/filepath"
	"sync"
	"sync/atomic"
	"testing"
	"time"
)

func TestProbeTwoWritersOneFolder(t *testing.T) {
	dir := filepath.Join(t.TempDir(), "0123456789abcdef")
	first, _ := newTestAgent(t, &scriptedCompleter{}, func(c *Config) {
		c.Place = Place{Dir: dir, Workspace: c.Workspace}
	})
	second, _ := newTestAgent(t, &scriptedCompleter{}, func(c *Config) {
		c.Place = Place{Dir: dir, Workspace: c.Workspace}
	})
	first.mu.Lock()
	first.stampUserLocked("diagnose parser migration failures")
	first.mu.Unlock()

	const rounds = 200
	losses, titleLost, spendLost := 0, 0, 0
	for i := 0; i < rounds; i++ {
		title := fmt.Sprintf("round %d name", i)
		spent, tokens := float64(i+1)/100, (i+1)*7
		var inside atomic.Int32
		// The rendezvous is the DETECTOR: two transactions that cannot overlap
		// never meet here, and one of them times out alone.
		meet := func() {
			inside.Add(1)
			deadline := time.Now().Add(50 * time.Millisecond)
			for inside.Load() < 2 && time.Now().Before(deadline) {
				time.Sleep(50 * time.Microsecond)
			}
		}
		var wg sync.WaitGroup
		wg.Add(2)
		go func() {
			defer wg.Done()
			first.updateMeta(dir, first.metaSnapshot(), func(meta *Meta) { meet(); meta.Title = title })
			inside.Add(-1)
		}()
		go func() {
			defer wg.Done()
			second.updateMeta(dir, second.metaSnapshot(), func(meta *Meta) {
				meet()
				meta.SpentUSD, meta.Tokens = spent, tokens
			})
			inside.Add(-1)
		}()
		wg.Wait()
		meta, err := LoadMeta(dir)
		if err != nil {
			t.Fatalf("round %d: %v", i, err)
		}
		if meta.Title != title {
			titleLost++
		}
		if meta.SpentUSD != spent || meta.Tokens != tokens {
			spendLost++
		}
		if meta.Title != title || meta.SpentUSD != spent {
			losses++
		}
	}
	if losses > 0 {
		t.Fatalf("losses=%d/%d (title_lost=%d spend_lost=%d)", losses, rounds, titleLost, spendLost)
	}
}
```

The archive half is the same probe with one side replaced: enter a spend transaction,
call `SetArchived(dir, true)` from a goroutine while the patch is inside, and read the
folder back — `archived=false` with the total on it, on a clean `dev`.

**Field (real models).** Two `aforge` windows, one conversation. In the first, ask
anything that takes a turn; while it is answering, press `e` on that conversation in the
second window's home. Needs `OPENROUTER_API_KEY`, one turn's money, under a minute. It
fires when the archive lands inside the seal's read-and-rename — which is microseconds
wide, so this is the shape rather than the reliable trigger; the probe above is.
The same window is open every time a `--host` engine and a local surface share the disk,
because there the two writers are on different machines and nothing at all is between them.

## Where

- `updateMeta` in `internal/session/placemeta.go` — the read-modify-write, serialized on
  `a.metaMu`, which is a field of one Agent.
- `SetArchived` in `internal/session/place.go` — the same read-modify-write with no lock
  at all, reached from home's `e` and from `cmd/aforge/engine.go`'s `Archive`.
- `anchorMetaLocked` in `internal/session/tools_anchor_workspace.go` — the third one.
- `internal/filelock` is already the house answer for exactly this class:
  `internal/standing/store.go`'s `underItemLock` and `internal/session/task_lock.go`'s
  `claimGitRoot` are the two worked examples.

## The fix

The conversation folder's identity file gets a lock of its own — a `meta.lock` beside it,
flocked for the length of the whole read-modify-write, with the re-read INSIDE the lock and
the temp-and-rename kept. Every writer that merges into `meta.json` goes through it, so two
windows, two engines and a `--host` pair on one disk queue instead of overwriting.

Three things the lock must not change. It never refuses: a filesystem that cannot flock, or
a lock file that cannot be created, writes exactly as it does today — `meta.json` is a
citation and not the record, and every failure here is silence. It never parks a turn: a
holder that has wedged is waited on for a bounded time and then written past, which is the
behaviour before the lock existed. And it says nothing to anybody: no new message, no new
refusal, no changed row.

## Acceptance

- **e2e:** two OS processes on one conversation folder, one naming and one recording what
  turns cost, 100 rounds through the real stamps — `meta.json` carries BOTH the name and
  `spentUSD` at the end of every round, and the run exits 0.
- **e2e:** the control, through the real door: `bin/aforge` in tmux against an isolated
  home, one turn, then `e` on the conversation in home — the row moves under `Archive`,
  and the folder's `meta.json` reads `"archived": true` with its `"spentUSD"` and
  `"title"` still on it.
- **Unit:** the failure road — a folder whose lock cannot be taken still stamps, so
  `LoadMeta` answers the new title afterwards; and a transaction whose patch is inside the
  file cannot be entered by a second one at the same instant (the rendezvous above times
  out rather than meeting).
- The manual page that covers this quotes the new wording (`internal/manual/chat/`), and
  the change entry's `invalidates` names what people believed before.

AbirAbbas:
## What was true

`meta.json` in a conversation folder is replaced atomically, so every reader sees a whole
identity — but the read BEFORE that rename was never part of the same act. A title stamp
and a spend stamp could each read the old identity, change their own field, and each
rename a complete file over the other. The only thing between them was `a.metaMu`, a mutex
on ONE `Agent`: it closed the hole for one window and left it open for a second Agent in
the same process and for every second process on the same conversation — a resumed copy,
home putting a conversation away while its own window seals a turn, and an engine at the
far end of `--host` beside a local surface on shared disk.

Measured on `dev@233f729d5`, two writers on one folder, 200 rounds of "one commits the
name while the other records what the turn cost":

```
losses=200/200 (title_lost=190 spend_lost=10)
```

And home's own put-away (`ctrl+e` → `session.SetArchived`) racing a seal:

```
archived=false spent=0.25 tokens=140
```

The conversation came back out of the archive, and nothing was said about it.

## What is true now

The serialization belongs to the FOLDER. `withMetaLock` (`internal/session/metalock.go`)
takes a per-directory process mutex — the cheap first gate for this process's goroutines,
as `task_lock.go`'s does — and then flocks a `meta.lock` beside the identity, so every
process meets on one inode before any of them reads. The whole read, patch and rename runs
inside it and the temp-and-rename stays. All three read-modify-writes go through it:
`updateMeta`, `SetArchived` and `anchorMetaLocked`. `Agent.metaMu` has no users left and
is gone.

It never refuses and it never parks a turn: a filesystem that cannot flock, a lock file
that cannot be created, and a holder that outlasts the bounded wait all run the transaction
unlocked — exactly what these stamps did before the lock existed, because `placemeta.go`'s
EVERY FAILURE IS SILENCE makes a best-effort citation better than a refused conversation.

The same probes after:

```
losses=0/200 (title_lost=0 spend_lost=0 overlaps=0)
archived=true spent=0.25 tokens=140
```

## Validation contract

Written before the tests, in `internal/session/metalock_test.go`; each test names its line.

1. Two windows on one conversation folder never overwrite each other's identity — 40
   rounds of name-against-total, and the patches cannot even overlap (the rendezvous is the
   detector: transactions that cannot meet time out alone).
2. Two PROCESSES on one folder keep both fields — 60 file-barrier rounds through the real
   stamps, each child re-executed from the test binary, each reporting its own tally.
3. Putting a conversation away from home while its turn is sealed keeps both.
4. A folder whose lock cannot be taken still stamps — an unavailable lock never refuses.
5. A holder that never lets go does not park the stamp — the wait is bounded, then it goes
   ahead.
6. A reader mid-write still sees a whole identity — the previous one, never half a file.

Must not change, and did not: no new person-facing string, no new refusal; rows, titles,
totals and the archive read as they did; `meta.json` keeps its path and its shape.

Contracts 1, 2 and 3 fail on the unlocked code — with the lock stubbed out they report
`overlapped 40 times and lost fields in 40/40 rounds`, `losses=14/60`, and the lost archive
above. Contracts 4, 5 and 6 are the guards on what must not change.

Fixes #687

## Gates

```
go build ./...                                          build ok
go vet ./...                                            vet ok
make fmt-check                                          fmt ok
make test-packed-manual                                 ok  internal/manual  1.242s
go test ./internal/manual/                              ok  internal/manual  1.933s
go test -run Manual ./internal/tui3/ ./internal/session/  ok internal/tui3 0.175s · ok internal/session 2.564s
make test-laws                                          laws: 57 files in 17 packages — every package ok
go test -count=1 -timeout 15m ./internal/session/       ok  internal/session  252.789s
make build                                              bin/aforge written
```

The touched Go package is `internal/session` alone; the other change is one manual page.
The full suite above was re-run after the rebase onto `dev@c667db69d`.

## Manual test — the real binary, in tmux, against an isolated home

`bin/aforge chat --no-host --yolo --model deepseek/deepseek-v4-flash`, one turn:

```
flock lock defined in five words · deepseek-v4-flash · via digitalocean · 1.3s   $0.0011 · 14.9k/1M · 1% · YOLO · idle
```

The folder now carries the lock beside the identity, and the identity is whole:

```
meta.json  meta.lock  presence.json  told.json  transcript.jsonl  work/
{"title":"flock lock defined in five words","shortTitle":"flock lock","spentUsd":0.0011376951,"tokens":15818}
```

**A second process holding the folder's lock never parks a turn.** Another process took
`meta.lock` and held it for twenty seconds; a turn was sent one second later. The total
reached the file thirteen seconds in — eight seconds before the holder let go — so the
bounded wait elapsed and the stamp went ahead rather than waiting on it:

```
holder started 14:51:58
 7 14:52:05 spentUsd=0.0011376951
12 14:52:11 spentUsd=0.0023240612
holder gone   14:52:25
```

and the surface stayed live throughout:

```
flock lock defined in five words · deepseek-v4-flash · via digitalocean · 0.7s   $0.0023 · 14.9k/1M · 1% ▁▁ · YOLO · idle
```

**A second aforge process put the conversation away while the first window held it.**
`ctrl+e` on the row in the second window's home; the row left the list, and the folder kept
the name and the total it already had:

```
{"title":"flock lock defined in five words","spentUsd":0.0023240612,"tokens":33957,"archived":true}
```

Then the first window took another turn: the total rose and the other window's archive
stood.

```
{"title":"flock lock defined in five words","spentUsd":0.0035092072,"tokens":51488,"archived":true}
```


santoshkumarradha:
Latency coordination with draft #688: the folder-level transaction fixes a real lost-update problem, but the new wait is on synchronous foreground paths. `stampUserLocked` calls `updateMeta` while holding `a.mu`; `sealTurn` calls `stampSpend` before returning to emit `EventTurnDone`. A held cross-process lock can therefore add up to five seconds to accepting a message or finishing even #688's report-only turn. Also, `local.Lock()` is outside the five-second deadline, so the total wait is not bounded by that deadline when several same-process writers queue.

Please add a real Submit/steer/turn-ending test with another process holding `meta.lock`, including multiple local writers, and assert responsiveness as well as preservation of metadata fields. Adapt the metadata write path so nonessential stamps do not hold the turn lock or delay its ending for lock patience; retain atomic read/patch/write ownership rather than simply removing the lock. This is a source-level contention risk, not an explanation for our existing frozen runs (which exclude this PR). Refer to #688's ownership regressions when testing the combined change.

PR #696, opened 2026-09-09
Task states: the deleted words leave the engine's own strings

santoshkumarradha:
Lane R of the task-states wave. Small and surgical: the words `docs/design/task-states/DESIGN.md` deletes were still spelled in `internal/session`'s own strings after the tiers landed.

## What was true

`needs your look` was four things at once, each correct on its own page:

- `task_audit.go`'s `needsLookLead = "finished, but needs your look — "` — the lead **every** your-call landing report opened with, whether nobody could check the work or its branch would not merge. The card quotes that sentence and the manual quoted it back.
- `principal_wire.go`'s `waitWord` — "wire the handlers is waiting on port the parser, which needs your look".
- `taskgrade.go`'s `taskGradeOutcome` — the word written into the ratings file.
- `pending.go`'s `taskUnverifiedNews`.

## What is true now

- **The report opens with the question the row is asking.** `yourCallLead(TaskFacts)` reads `taskAskOf`'s own sentence (`task_status.go`) rather than spelling a second one, so a branch that would not merge leads `conflicts with your branch — ` and every other your-call landing leads `nobody could check it — `. The report and the card can no longer be two accounts of one landing.
- **The note takes that lead off with the reading's own reason** (`status.Reason + yourCallDash`) instead of a copy of it.
- `waitWord` says `is your call`, `taskGradeOutcome` says `your call`, `taskUnverifiedNews` **is** `taskWordYourCall` — all three spelled from the one constant.
- An empty check now says `the checker never answered`, which is the design's own wording, rather than `nobody could say whether it holds` under a lead that already said as much.
- **`TestNoStringInThisPackageSpellsADeletedTaskWord`** (`internal/session/taskwords_law_test.go`) walks the package with `go/ast` and fails any string literal that spells `needs your look` or `awaiting review` again. It reads literals only — `TaskUnverified` and the engine's other Go names are untouched by the design, and several comments say on purpose which word used to be there.

## The pages this change made wrong, fixed here

Only the ones that quoted a string this change respells:

- `how-tasks-run.md` — nine quoted report leads, and the paragraph that described the old lead now describes the new rule.
- `starting-aforge.md` — the stuck-work sentence `waitWord` writes.
- `models-and-cost.md` — the ratings-file outcome words.

`CLAUDE.md`'s vocabulary law now ends `or *your call*`, and the three comments in `cmd/aforge` and `internal/exec` that restate that list follow it.

## One thing the wave should look at

A ground-shift landing (`landShifted`) reaches `taskAskOf`'s `default` arm, so its row — and now its report lead — read `nobody could check it`, when in fact the work *was* checked and held; what happened is that the ground moved. That is the projection's table, not this lane's; flagged for whoever owns `taskAskOf`.


santoshkumarradha:
Integration note with draft #688: both PRs touch task result/notification code, `mailbox_test.go`, and `how-tasks-run.md`. Please retain #688's typed background-result provenance, task admission request identity, and its background-report manual section when rebasing; adapt the wording without changing those fields or reducing a result to an untyped string.

A finished reporting turn must remain distinct from task completion or a `your call` decision. Actual task-result repair and new human requests still proceed. Run the notification/background-origin controls beside this PR's vocabulary law and task-result tests; the combined build should retain the new authoritative words without reopening a report as a new assignment. #688 remains draft while its matched benchmark runs.

PR #699, opened 2026-09-09
chat: a turn the conversation starts on its own reaches the screen on the ordinary door

### LINKED ISSUE #690: chat: over the ordinary door a turn the conversation starts on its own runs, spends and journals, and never reaches the screen
## What happened

`dev@233f729d5`, 2026-09-09. Over the ordinary door — `aforge chat --yolo --model deepseek/deepseek-v4-flash`, which takes the engine host road — a background job was started and the person stayed in the conversation:

```
› start a background job with run_bash: bash -c 'sleep 30; echo CLEANWAKE' and tell me when it exits
```

The screen stops at the "started" turn and never moves again. Ninety seconds later the status line still reads

```
notify on background job exit · deepseek-v4-flash    1 job · 1 watch ·  $0.0024 · … · YOLO · idle
```

while the conversation's own journal has the answer the session wrote about the exit:

```
{"type":"message","role":"assistant","content":"Job 1 finished — exited 0, printed `CLEANWAKE`. …"}
```

The same prompt over `aforge chat --no-host --yolo` draws the reply. So a turn the conversation starts on its own — a job exiting, a task landing, a watch firing — runs, spends and journals, and the person sitting in front of it sees nothing. It is the exact defect `internal/remote/wakelane.go`'s own header says it fixed; the engine end of that lane is right, and the surface end throws the turn away.

## Replication

**Deterministic (no model).** In `internal/tui3`, arm the follow lane the way the surface arms it at boot, let ONE ordinary turn happen, then hand the lane a turn:

```go
a, _ := watched(t, Driving{Machine: "spark"})
turns := make(chan Following, 1)
a.link.Follow = func() <-chan Following { return turns }
cmd := a.watchFollowing()          // armed once, with everything else, at boot

prev := make(chan session.Event)   // one turn this window started
close(prev)
drive(t, a, runCmd(a.takeStream(prev))...)

events := make(chan session.Event, 2)
events <- session.Event{Kind: session.EventTextDelta, Text: "CLEANWAKEREPLY"}
events <- session.Event{Kind: session.EventTurnDone}
close(events)
turns <- Following{Events: events}
drive(t, a, runCmd(cmd)...)
```

Today that prints, and the frame is empty:

```
gen after one turn = 1
msgs=[{turn:{Said: Events:0x…} gen:0}]
drawn=false follows=0 stream_nil=true
```

Drop the `takeStream` line — no turn typed first — and the same code draws it. That is the whole defect: the wait was armed with the TURN generation, one turn moved it, and the arrival is discarded by a check meant to catch a conversation the window walked away from.

**Field (real models).** `OPENROUTER_API_KEY`, about ninety seconds and under a cent:

```
mkdir wakeprobe && cd wakeprobe
aforge chat --yolo --model deepseek/deepseek-v4-flash
# › start a background job with run_bash: bash -c 'sleep 30; echo CLEANWAKE' and tell me when it exits
# stay in the conversation for ninety seconds
```

The reply never lands on screen; `tail -3` of the conversation's `transcript.jsonl` has it. Control run: the same thing with `--no-host` draws it. Keep the aforge home SHORT (`AFORGE_HOME=/tmp/h9`) or the Unix socket path is over the ceiling and the launch silently falls back to the in-process door, which is the door that already works.

## Where

`internal/tui3/watching.go`: `app.watchFollowing` captures `a.gen` and `app.followTurn` compares against it. `a.gen` is the TURN generation — `app.takeStream`, `app.startFollow` and `app.detachTurn` all bump it once per turn — so from the first typed turn onward every arrival on the lane fails the check, is dropped, and the wait is NOT re-armed, which deafens the window for the rest of its life. `app.drivingMoved`'s `msg.lane` branch reads the same field and has the same hole. The lane that a conversation switch is actually meant to invalidate is the one `app.watchWakes` keeps with `a.wakeGen`, bumped only where a lane is re-armed — which is why the local door has never had this bug.

## The fix

The two connection lanes get their own generation, bumped where they are armed and nowhere else, exactly as the local wake lane's is. A turn starting must not silence the lane that carries the conversation's own turns; a conversation switch must still discard the wait parked on the one this window left.

## Acceptance

- **e2e:** over the default door (no `--no-host`), a background job started with `run_bash … sleep 30` and left alone draws the session's reply — the transcript's `content` string on screen — within a few seconds of it reaching `transcript.jsonl`, with no `›` row above it.
- **e2e:** the `--no-host` control run is unchanged, and a second and third self-started turn in the same conversation each reach the screen.
- **Unit:** `internal/tui3` — a follow-lane arrival AFTER a turn has bumped the turn generation is drawn; an arrival from a conversation the window has switched away from is still discarded and re-arms nothing.
- The manual page that covers this quotes the wording (`internal/manual/chat/`), and the change entry's `invalidates` names what people believed before.

AbirAbbas:
## What was true

Over the ordinary door — a bare `aforge chat`, which is served by a session host — a turn
the conversation starts on its own ran, spent and journalled, and never reached the screen.
A background job exited, a task landed, a watch fired: the model wrote the answer, the
transcript kept it, and the person sitting in front of the conversation saw a still frame
and `idle`. Over `--no-host` the same prompt drew it. It is the exact defect
`internal/remote/wakelane.go`'s own header says it fixed — the engine end was right, and
the surface threw the turn away.

The follow lane (`app.watchFollowing`) and the keyboard lane (`app.watchDriving`) captured
`a.gen`, which is the **turn** generation: `takeStream`, `startFollow` and `detachTurn` each
move it once per turn. The lanes are armed once when the window opens, so the person's first
sentence made the standing wait stale; the arrival was discarded by a check written to catch
a conversation the window had walked away from, and — by that same rule — nothing re-armed
the wait. The window was deaf for the rest of its life, and `hostFollow`'s sender parked
forever on a send nobody would ever read, so every later turn stalled at the client too.

Every `internal/remote` wake-lane test passed throughout, and so did every follow-lane test
here: in a test nothing bumps `a.gen` before the arrival, which is exactly the step the
defect needs.

## What is true now

The two lanes a **connection** carries have their own generation, `linkGen`, bumped in
`clearConversation` beside every other lane's — the one place a conversation leaves the
front — and by nothing else. A turn starting no longer deafens a window to its own
conversation's turns; a wait parked on the connection the window left is still discarded and
still arms nothing, including when the conversation replacing it is local and arms no lanes
at all. This is the check `app.watchWakes` has kept with `wakeGen` since the day it was
written, which is why the door with no connection never had this.

Fixes #690

## Validation contract

- When a background job exits, a task lands or a watch fires while a person is sitting in
  the conversation, the reply is on screen within a few seconds of reaching the journal —
  over the ordinary door as well as `--no-host`.
- And the one after that, and the one after that: the lane re-arms on every arrival.
- It holds however many turns the person has typed first. Typing does not deafen the window.
- A self-started turn is still drawn with no `›` row above it — nobody typed it.
- One arriving while the previous stream is still draining is queued and drawn after it,
  never dropped.
- A turn from the conversation this window has walked away from is still discarded, and its
  wait is still not re-armed.
- The keyboard moving on another machine still reaches this window after a typed turn, and a
  take-back that failed is still always heard whatever the generation.
- Unchanged: `--no-host`, `internal/remote`, the local wake lane, and the turn generation
  still discarding a stale stream's events.

## Gates

```
go build ./...                                     BUILD-OK
go vet ./...                                       VET-OK
make fmt-check                                     FMT-OK
make test-packed-manual                            ok  …/internal/manual  (cached)
                                                   (no tracked packed corpus in the diff)
go test -count=1 ./internal/manual/                ok  …/internal/manual  1.378s
go test -run Manual ./internal/tui3/ ./internal/session/
                                                   ok  …/internal/tui3     0.181s
                                                   ok  …/internal/session  0.283s
make test-laws                                     ok  …/internal/tui3     9.918s   (LAWS-OK)
make test PKGS=./internal/tui3/                    ok  …/internal/tui3   431.277s
make build                                         bin/aforge written
```

The three new tests are real regression tests. With the two captures put back to `a.gen` and
nothing else changed:

```
--- FAIL: TestASelfStartedTurnAfterATypedTurnReachesTheScreen (0.16s)
    watching_test.go:223: the first reply after a typed turn never reached the screen:
--- FAIL: TestEverySelfStartedTurnAfterATypedTurnReachesTheScreen (0.16s)
    watching_test.go:252: the repeated replies lost "THE-SECOND-WAKE-ARRIVED":
--- FAIL: TestAKeyboardHandOverSurvivesATypedTurnAndRejectsAStaleConnection (0.15s)
    watching_test.go:417: the hand-over did not redraw and re-arm: cmd=false painting=false arms=1
```

## Driven by hand, in tmux, against a real model

`aforge chat --yolo --model deepseek/deepseek-v4-flash` — the ordinary door, no `--no-host`,
with a live engine-host socket — then `start a background job with run_bash: bash -c 'sleep
30; echo CLEANWAKE' and tell me when it exits`, and nothing else touched for two minutes.

**Before, on `dev@233f729d5`.** The last thing drawn, and the status line beneath it:

```
  Job 1 is running (`bash -c 'sleep 30; echo CLEANWAKE'`). I've set a watch on it — it'll
  report back when it exits and prints `CLEANWAKE`.

  In the meantime, anything else you need?

notify on background job exit · deepseek-v4-flash    1 job · 1 watch ·  $0.0024 · … · YOLO · idle
```

while the journal collected the turns nobody could see:

```
Job 1 completed — it ran, printed `CLEANWAKE`, then exited.
Job 1 **exited with status 0**, and its stdout produced exactly: `CLEANWAKE`.
```

**After, on this branch.** The journal wrote the self-started reply at 19:06:17Z; the frame
captured 40 seconds after the sentence was sent — about a second later — already has it,
with no `›` row above it:

```
  › start a background job with run_bash: bash -c 'sleep 30; echo CLEANWAKE' and tell me when it exits

    ▸ worked 6.2s · thought 3.0s · 1 tool call · ctrl+e

  Job 1 started — `sleep 30; echo CLEANWAKE`. It'll report here on its own when it
  finishes.

    ▸ worked 13s · thought 0.3s · ctrl+e

  Job 1 exited 0, printed `CLEANWAKE`. Clean exit.
```

**Control, `--no-host`, same branch and same prompt** — unchanged:

```
  Job 1 landed clean — exited 0, output `CLEANWAKE`. Nothing else needed.
```


santoshkumarradha:
Coordination with draft #688: these fixes are complementary. #688 makes a background-result turn report the actual outcome and stop without creating another owner while the original task is live; this PR makes that self-started reply visible through the ordinary hosted UI. Please retain our `how-tasks-run.md` background-report section alongside this PR's hosted-wake guidance.

For the combined regression, send a normal typed turn, let a job result wake a short report while its existing task continues, and assert the report/notice is drawn once, later task completion is also drawn, and no extra owner is admitted. Include a conversation switch to retain your stale-connection protection. Our current benchmark uses the persistent engine protocol and already reads those events, so this UI fix must not be claimed as the cause of a speed gain in that frozen cohort; it is directly relevant to perceived responsiveness in regular chat.

santoshkumarradha:
Superseded by #856, which landed the same two-lane fix (`linkGen`, now folded with #855's `steerGen` into one `app.convGen`) rebased onto today's dev, and carries this PR's `how-tasks-run.md` paragraph. This branch conflicts with dev in `app.go` after #855. Thank you for the diagnosis; closing.

PR #712, opened 2026-09-09
Questions: one object for every decision the engine hands a person

santoshkumarradha:
Lane **E1 — the question contract** of the questions wave. Engine side only: no
surface change, and nothing a person sees today behaves differently.

`docs/design/questions/DESIGN.md` lands here as the copy of record for the wave,
with the final exported names written back into it (one source of truth).

## The problem, from `ideation/questions-audit.md`

Thirteen engine lanes each minted an id, banked a wait, emitted an event and were
answered through a door of their own — twenty-one distinct question mechanisms
across fourteen answer surfaces, three of which could draw an answer row at all.
`ResolveConflict`, `TakeBackDecision` and `AnswerSubharness` had no caller
anywhere in the product: work could stop on a question nothing could draw, let
alone answer. And a landed task's `your call` lived in a registry the one
predicate never folded in, so a session sitting on one told home, the switcher
and the tab signal there was nothing to do.

## What lands

**`internal/session/question.go`** — `Question`, and the vocabulary around it:
`AskKind`, `QuestionForm`, `Asker`, `SubjectRef`, `Block`, `InputShape`, `Pick`,
`Stakes`, `Policy`, `Blocking`, `AnswerScope`, `Withdrawal`, `DecidedBy`,
`Exchange`, `QuestionCap`.

- **`Question.Check(records)`** is the gate. No head, no reason, no stakes, too
  few or too many answers, a pick naming an unknown key, a clock or a policy on
  irreversible stakes, and `already decided: …` where the record answers it.
  Every refusal ends in something the asker can act on. `Agent.AskQuestion` adds
  the one bound that belongs to the set — `QuestionCap` open questions against
  ONE subject.
- **`Agent.ResolveQuestion(Answer) error`** is the ONE door. Every arm calls the
  resolver a surface already calls, so answers.go's law — AN ANSWER IS APPLIED
  THROUGH THE SAME RESOLVER A SURFACE USES — is literally true across eleven
  lanes rather than nearly true across three. `Agent.applyAnswer` (the
  answers.jsonl drain) goes through it, and so do the three lanes that had no
  caller. `[s] tell it` steers the work and LEAVES THE QUESTION OPEN.
- **`Agent.OpenQuestions() []Question`** is derived from the lanes' own waits.
  No second store: `Agent.questionWords` keeps only the WORDS a lane said, and an
  entry with no wait behind it is never returned.
- **Events** `EventQuestion`, `EventQuestionWithdrawn`, `EventQuestionAnswered`,
  on **`Agent.WatchQuestions()`** — a standing lane of their own that replays
  what is already open when a surface attaches. Deliberately not the turn's
  stream (a strict sequence its readers walk to the close) and not the standing
  task lane (the roster's).
- **The record** — `decisions.jsonl`, `Agent.Decisions()`, `DecisionsSection`
  (`the record`, one line each: head → picked · with · by · when · reversible).
- **Withdrawal is automatic**: letting go of a question nobody answered
  withdraws it with a reason, so no lane has to remember; the presence beat
  sweeps whatever that misses.
- **Presence carries the whole question** in a new `PresenceQuestion.Full`, with
  the four older fields still filled — a window of any age reads what it read.
- **`NeedsPerson` counts a landed `your call`.**
- The consent, task-proposal and standing lanes now bank whole questions: the
  gate's tool and the policy's own phrasing, the proposal's clock and the pick
  that clock will take, the standing card's one-item answers. A running
  sub-harness's question is kept rather than thrown away with its event.

## Names that differ from the design's first draft

`QuestionKind` was already taken by answers.go for the LANE, and is on disk in
presence files read by other builds. So the object carries two kind fields:
`Kind QuestionKind` (the lane, extended from three values to eleven) and
`Ask AskKind` (the shape of the decision). `Option` is answers.go's
`AnswerOption` widened rather than twinned; `Answer` is answers.go's `Answer`
widened, so an older window's bare-key line still reads; `Scope` is
`AnswerScope`, since `ConsentScope` means something else.

## Verified

`go vet ./...`, `make test-laws`, `go test ./internal/session/
./internal/manual/ ./internal/tui3/`, `make changelog-check`, and the binary
under the `SIZE-BUDGET` ratchet (52.2MB of 54.6MB).

## Owed to the surface lanes

Nothing is drawn from this yet. S1–S3 read `OpenQuestions()`,
`WatchQuestions()`, `PresenceQuestion.Full` and `ResolveQuestion`; E2 owns the
`ask` tool and the ladder in `system.md`, which this deliberately does not touch.

https://claude.ai/code/session_01Sjru7qi9iUmnXZexBDRxJE


santoshkumarradha:
Integration finding from the context audit in #717: at head 926810e3, a conflict question exposes `LandingYesKey` (`a`) with the label `resolve it`, but `applyLanding` handles that key by calling `ResolveUnverified(TaskAccept)` before reaching the `QuestionConflict` fallback to `ResolveConflict` (question.go around 1448 and 1467).

For an existing merge conflict, choosing `a resolve it` therefore retries acceptance rather than entering the conflict-resolution/check round. Combined with #714, a changed but cleanly mergeable destination may be accepted without the recheck that #714 intends. Please dispatch by question kind before the shared answer key, and test the real `ResolveQuestion` path for both conflict forms. Preserve #717's current worker conclusion/rollback evidence when the resolution round rechecks; that draft contains the matching context fixes, not a replacement question protocol.

PR #714, opened 2026-09-09
Tasks: a ground shift asks the conflict's question, not "nobody could check it"

santoshkumarradha:
## What was wrong

`Agent.landShifted` is the landing for work that **held its check** and whose files
were changed on the person's branch while it ran (`groundladder.go`'s ground shift).
It landed `TaskUnverified` with a report lead, and `taskAskOf` had no case for it — so
it fell to the default arm and the row, the card and the report all read
`nobody could check it`.

Both halves of that sentence were false: it *was* checked, and it *holds*. What
happened is that two versions of the same files now exist. Lane R flagged it in #696.

## What it does now

A ground shift is the **conflict shape**, so it asks the conflict's question with the
conflict's two answers — no new ask kind:

```
? ◆ Port the parser · your call · 6m40s · 2 files · branch kept · task/parser
  your branch changed the same files while it worked: parser.go, lex.go
  [a] resolve it · [n] drop it · [s] tell it
```

- `landShifted` records the moved files through `TaskNode.shiftedBy` (which goes through
  the existing `clashesWith` door) and marks the node, so `TaskFacts.Conflicts` carries
  them and `taskAskOf` yields `TaskAskConflict`.
- **The fact tells the two roads apart, never the prose.** `TaskNotice.Shifted` /
  `TaskFacts.Shifted` — necessary because the shifted landing keeps the merge word
  `kept`: its branch *would* have fastened. `conflicts with your branch: …` stays for a
  true merge conflict.
- `Agent.groundShift` now answers the moved files beside the sentence
  (`groundShiftFound`); `groundShiftReason` stays as the sentence-only door its tests use.
- Under `task.settle = auto` it is **never handed to the model** (`handToModelOnAuto`),
  the same rule as a conflict, and the landing note says so in its own words rather than
  borrowing the conflict's — that sentence opens by claiming the branch conflicts, which
  is untrue here.
- `ResolveConflict` needed no change: it never asked whether a marker was written, so
  the round was always the right verb for this road — what it lacked was a card that
  offered it. Its doc now says so, and a test holds it.

## Tests

- `ProjectTask` table rows for the shifted facts, with and without named files.
- `TestTheTwoRoadsToTheConflictQuestionSayDifferentThings` — same kind and answers,
  different sentences.
- `TestAGroundShiftLandsAsTheConflictsQuestionWithTheFilesNamed` — the real landing over
  a real repository, and the report leads with the row's own question.
- `TestResolveConflictSpendsARoundOnAGroundThatMoved` — the door accepts it, the round
  brings both ends of the file together on the branch, nothing is rewritten in place.
- `TestAGroundShiftUnderAutoStaysWithThePerson`, `TestAShiftedNoteRefusesTheModelTheMergeAndSaysWhatMoved`.

## Manual law

`docs/design/task-states/DESIGN.md`'s your-call table gains the second sentence under the
conflict row. `internal/manual/chat/tasks.md` gains a section in the asker's own words
("someone else changed the same file while the task was running"), and
`internal/manual/chat/how-tasks-run.md`'s existing ground-shift section — which showed
`nobody could check it —` and offered `[a] accept` — is corrected in the same change.


santoshkumarradha:
Please preserve the context fixes in draft #717 when combining this change: `task_run.go`/`task_merge_round.go` now keep the current worker/resolver conclusion, clear stale empty-attempt evidence, and tell later checks when resolver work was rolled back. Your shifted-ground `ResolveConflict` path should retain that actual current-state evidence through recheck.

There is also a concrete interaction with #712: its conflict question labels the shared `a` key `resolve it`, but currently dispatches that key to `ResolveUnverified(TaskAccept)` before the conflict fallback. I left the exact finding there. A combined test should take the real question-answer door and assert the changed ground goes through resolution/recheck rather than direct acceptance. #717 is draft and awaiting benchmark validation; this is coordination, not a request to merge either draft prematurely.

PR #715, opened 2026-09-09
chat: a dead lane's half-answer comes off the page when its rescue takes over

### LINKED ISSUE #703: chat: a rescue on another lane leaves the dead lane's half-answer on the page, above the answer it kept
## What happened

`dev@57940a81`, 2026-09-09. One model id is served by many machines. When the machine
answering a turn goes quiet mid-answer, the transport sends the same question to another
one and lets the two race (`internal/provider/hedge.go`). The rescue wins, a dim line says
so — and the half-answer the dead machine wrote **stays on the page**, with the rescue's
answer glued onto the end of it.

The conversation on disk contains only the rescue's answer. The page and the saved
conversation disagree, which is the exact divergence #653 closed for a transport retry and
left open here.

Driven with a stub OpenAI-compatible server (first POST streams ~240 characters and then
drops the TCP connection without `[DONE]`; every later POST answers normally),
`aforge chat --no-host --yolo`, one message:

```
PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT
PARTIALTEXTSTUBANSWER the link is back.

  · that lane went quiet — this answer is coming from
  another one
```

`grep -rl STUBANSWER <home>` matches `transcript.jsonl`. `grep -rl PARTIALTEXT <home>`
matches nothing. The notices drawn were `· still no answer — trying another lane ·
/model switches` and `· that lane went quiet — this answer is coming from another one`;
`the request failed — asking again` never appeared, because nothing on this road
announces a discard.

## Replication

**Deterministic (no model).** The surface's own reducer, driven with the event sequence
the wire produces today — a delta, the hedge line as a plain notice, the replacement's
delta. Drop this in `internal/tui3/` and run
`go test ./internal/tui3/ -run TestProbe -count=1 -v`:

```go
a := newTestApp(&fakeAgent{model: "m"})
a.state, a.turn, a.linear = stateWorking, 1, true
a.event(session.Event{Kind: session.EventTextDelta, Text: "PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT"})
a.event(session.Event{Kind: session.EventNotice, Text: "that lane went quiet — this answer is coming from another one"})
a.event(session.Event{Kind: session.EventTextDelta, Text: "STUBANSWER the link is back."})
a.event(session.Event{Kind: session.EventAssistantDone})
// livePage(a) still contains "PARTIALTEXT"
```

Today it prints the two answers as one paragraph:

```
PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT
PARTIALTEXTSTUBANSWER the link is back.
```

The transport half is already in the tree and passes:
`go test ./internal/provider/ -run TestALaneThatStallsMidAnswerIsHedgedAndTheAnswerArrivesWhole -v`
asserts the flip raises exactly one **`StreamNotice`** carrying that line — a line to
print, and nothing that tells anybody the words above it are void.

**Field (real models).** Any turn whose serving machine stalls mid-answer long enough to
be rescued. `OPENROUTER_API_KEY`, a model with several machines behind it, a long answer;
a few cents and a minute or two. It fires on whatever fraction of turns the speed guard
acts on, so the stub above is the one to gate on.

## Where

- `internal/provider/hedge.go`, `hedgeRace.flip` — emits `StreamNotice` with `hedgeNotice`
  when `r.spoken`, then replays the winner's held events. Nothing says the earlier text
  was withdrawn.
- `internal/session/loop.go`, the `turnObserver`'s `case provider.StreamNotice` — turns it
  into `EventNotice`, a dim line and nothing else. The buffers (`partial`, `reasoning`,
  `warm`, `forming`) keep the dead lane's work, so an interrupt after a rescue would
  journal both halves as one answer.
- `internal/tui3/feed.go`, `feed.retry` / `feed.dropLive` — the withdrawal that already
  exists, reached only from `EventRetrying`.
- `internal/manual/chat/screen.md`, "Does it ever ask a second time in parallel" — says
  *"when it wins nothing is taken away"*, which is the belief this defect creates.

## The fix

A rescue that replaces an answer somebody is reading withdraws that answer, exactly as a
retry does. The transport raises a replacement rather than a bare notice, the session
discards the dead lane's partial text, its reasoning and its half-arrived calls before the
winner's events are replayed, and the surface takes its rows off the page.

The person still reads the same line, once:

```
  · that lane went quiet — this answer is coming from another one
```

A rescue that wins before any text reached the screen still takes nothing away and still
says nothing.

## Acceptance

- **e2e:** `aforge chat` against the stub above, one message: the frame after the turn
  settles contains `STUBANSWER` and does not contain `PARTIALTEXT`, and
  `grep -c PARTIALTEXT` over the profile's `transcript.jsonl` is 0 — page and conversation
  saying the same thing.
- **e2e:** the control, same binary, a stub that answers normally on the first POST: no
  notice is drawn, nothing is withdrawn, and the answer is on the page once.
- **Unit:** `internal/provider` — the flip raises the replacement kind exactly once when
  text had been shown, and raises nothing at all when the rescue wins before the first
  word. `internal/session` — a replacement mid-stream reaches the room as `EventRetrying`,
  and a turn stopped after it journals only the rescuing lane's text, never the two halves
  concatenated. `internal/tui3` — the reducer sequence above leaves `STUBANSWER` on the
  page with `PARTIALTEXT` gone and the line drawn once.
- The manual page that covers this quotes the new wording
  (`internal/manual/chat/screen.md`), and the change entry's `invalidates` names what
  people believed before.

AbirAbbas:
One model id is served by many machines. When the one answering goes quiet
mid-answer, the transport sends the same question to another one and lets the
two race (`internal/provider/hedge.go`).

**What was true.** The rescue won, a dim line said so — and the half-answer the
dead machine had already written stayed on the page, with the rescue's answer
glued onto the end of it. The saved conversation held only the rescue's answer,
so the page and the conversation disagreed. This is the divergence #653 closed
for a transport retry and left open on this road; nothing on it announced a
discard, so `the request failed — asking again` never appeared and no surface
knew the words above were void.

**What is true now.** A rescue that takes over an answer somebody is already
reading raises `provider.StreamReplaced` — the same line, plus the instruction
to throw away what is above it. The turn loop answers it the way it answers a
cut: the partial answer, the reasoning beside it, the reads it started early and
the calls that were still arriving are dropped, and the room is told on
`EventRetrying`, which every surface already reads as "throw away what you drew".
A rescue that wins before the first word is still a silent swap. The manual's two
claims that "nothing is taken away" are corrected in the same change.

Fixes #703

## Validation contract

1. When a rescue on another machine replaces an answer somebody is reading, the
   dead machine's words come off the page, and the page says the same thing as
   the saved conversation.
2. The person is still told, in the same words, once:
   `that lane went quiet — this answer is coming from another one`.
3. A rescue that wins before a single word reached the screen takes nothing away
   and says nothing.
4. The dead machine's text never reaches the conversation — a turn stopped after
   the rescue journals the rescuing machine's text alone, never the two halves
   concatenated.
5. The dead machine's private working and its half-arrived calls go with its
   text, exactly as on a transport retry.
6. The transport retry is unchanged: one `the request failed — asking again`,
   still withdrawing; a stop during backoff and an exhausted ladder still keep
   their partial reply.
7. The adapter's request-reshaping notices (`Retry 1/3: …`), the pin lines and
   the first-prompt line stay plain notices and take nothing off the page.

## Gates

```
go build ./...                                            (no output, exit 0)
go vet ./...                                              (no output, exit 0)
make fmt-check                                            (no output, exit 0)
make test-packed-manual   ok  internal/manual  (cached)
make test-laws            ok  internal/tui3    10.393s   (58 law files, 17 packages)
go test -count=1 -timeout 15m ./internal/provider/   ok   54.720s
go test -count=1 -timeout 15m ./internal/session/    ok  234.973s
go test -count=1 -timeout 15m ./internal/tui3/       ok  417.562s
go test ./internal/manual/                           ok  (cached)
go test -count=1 -run Manual ./internal/tui3/ ./internal/session/   ok 0.170s / ok 0.259s
go test -count=1 ./internal/e2e/                     ok   0.173s   (the surface-words table)
make build                                           bin/aforge written
```

## Manual test — the real binary, in tmux, against a stub router

A stub that serves an endpoints page (so the base carries routing preferences)
and, for the one completion that carries the marker, streams a run of
`PARTIALTEXT` and then goes silent with the connection still open. Every later
completion answers `STUBANSWER the link is back.` at once — the rescue taking the
answer. `aforge chat --no-host --yolo`, one message.

**Before, on `dev@57940a813`** — the two answers as one paragraph, the notices
under them:

```
  › REPROB26 answer me

  PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT
  PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT
  PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT PARTIALTEXT STUBANSWER the
  link is back.

    · still no answer — trying another lane · /model switches
    · that lane went quiet — this answer is coming from another one
```

`PARTIALTEXT` matched 0 files under the profile; `STUBANSWER` was in
`transcript.jsonl`. Page and conversation disagreed.

**After, this branch** — one answer, and the story of how it got there one row
away:

```
  › REPROB26 answer me

    ▸ worked 10s · ctrl+e

  STUBANSWER the link is back.
```

`ctrl+e` opens it, each line once, where the withdrawn text was:

```
    ▾ worked 10s · ctrl+e

    · still no answer — trying another lane · /model switches
    · that lane went quiet — this answer is coming from another one

  STUBANSWER the link is back.
```

`PARTIALTEXT` is now on neither the page nor in the profile;
`"role":"assistant","content":"STUBANSWER the link is back."` is the whole of
what the conversation kept. The stub's own log shows the race — the stalled
completion and the rescue that replaced it — and the original repro no longer
reproduces.

**Control**, same binary, same stub with nothing stalled: no notice, nothing
withdrawn, `STUBANSWER the link is back.` on the page once with no work row.

## The neighbouring claim that is not a defect

The same sweep asked whether streamed text left on the page under
`finish_reason: error` is the same divergence. It is not. Driven through the
stub that answers every completion with `finish_reason:error`, the page shows
`visible partial` under
`· error: after 3 retries: API error (502): provider ended the response with finish_reason=error`
— and the conversation holds `"role":"assistant","content":"visible partial"`.
Page and conversation agree, which is the law `runTurn` states for a permanent
failure mid-stream: the streamed text is real work the person watched arrive, so
it is kept and the turn is sealed. Unchanged by this branch, checked on both
binaries.


santoshkumarradha:
Context-audit coordination with draft #717: this stream replacement fix is complementary. #717 pairs tool results only with their actual assistant batch and keeps outcome/read-consumption identity per call occurrence, because replacement and later responses can reuse provider call IDs.

When combining the changes, please retain the reset of abandoned partial text/reasoning/tool assembly and add or preserve a combined regression where a replacement response reuses an earlier ready call's ID. The final transcript and completion/checker evidence should contain only the surviving call's arguments and its own result, with no abandoned partial answer or borrowed success. Keep this as context propagation testing; #717 adds no retry or ownership policy.

PR #717, opened 2026-09-09
Preserve context across tool rounds and worker handoffs

santoshkumarradha:
Provider tool-call IDs can repeat between rounds. We treated them as transcript-wide identities, which could label an earlier result with a later command, mark an unanswered call successful, or compact a fresh read before the model had used it. Worker handoffs also omitted existing conclusion/write evidence and could retain stale evidence after repair or rollback.

This change pairs results with their actual assistant batch and tracks runtime evidence by call occurrence. It preserves retrievable result sources, supplies the current bounded worker conclusion and actual check directory, carries already-declared checks across a full-request handoff, and includes small completed write inputs in the existing completion digest. Empty attempts and rollback no longer masquerade as current successful work. The main prompt now treats a completion-reader observation as something to check against actual work. Restored output, change explanations, and rewind use the same call association; legacy internal continuation text remains hidden.

No additional model calls, decision layers, model overrides, generation limits, or issue-specific behavior are introduced. Existing reading-only check authority remains unchanged; prose commands do not become executable declarations. Manual, performance documentation, and change entry are included.

Validation for final source `84ce69bba`:

- Uncached full session and manual suites pass (290.989s / 1.936s).
- Focused race regressions pass; quick checks, including build/vet/format, packed manuals and laws, pass on the identical production source.
- Native and static Linux ARM64 `make build` pass with clean source stamps.
- Focused combination with dev `8d8cd6c895` passes, including ground-shift resolution/recheck and preserved worker/rollback context.
- Regression tests cover repeated IDs, unanswered/orphan results, early read folding, evidence retrieval/inheritance, completed write inputs, declared-check handoff, repair/rollback, and real saved-history/rewind behavior.

The full run exposed an existing test observation race: a hand can finish after the closing request snapshot, and its report correctly arrives in the next wake. A controlled reproduction confirms this on baseline and candidate; the test now observes that wake. Two digest/file-count fixtures were corrected to use legal tool batches and the existing separate evidence directory.

A fresh Spark cohort is running as Fleet `20260909-205221-000387-af-context-fidelity-candidate-02`: five frozen GitHub issues × two seeds, plus four writing/data controls, exact `deepseek/deepseek-v4-flash-0731`, provider generation defaults, and at most two concurrent cells. All 145 consumed fixtures/graders are authenticated; the offline binary/image preflight passes. Previous Pi/mini results are the reference rather than simultaneous fresh competitors, so provider conditions can differ. Private serialized request logs are enabled for this candidate. The first live admissions confirm the exact model/defaults and retained request bodies. The frozen benchmark remains in progress; no Pareto claim is made. All CI checks on this PR are green.

The initial infrastructure attempt failed before Submit because its frozen protocol-12 client met the current protocol-13 binary (#691); all 14 cells made zero model calls. It is preserved separately. The active attempt uses an independently frozen driver with exactly those two protocol literals updated, and its real offline Hello/Watch/Close preflight passes. Original fixtures and grading are unchanged.

This is the context-only alternative to the held experiment in #688. The owner authorized merging the best-supported candidate. This narrow repair is selected over the broader #741 experiment, whose latest frozen run has writing and CPF quality regressions. #714 is tested for integration but excluded from the frozen candidate to keep its separate behavior out of this experiment. Restored caption metadata chronology remains a separate display-only limitation.

Combined-dev validation at `0a4ef1ff4` passes: quick gate (build, vet, formatting, packed manual, laws) and full uncached session/manual suites (290.507s / 1.692s). All PR checks are green. This merge adds context correctness fixes, not a claim that every benchmark issue is solved. PR #741 remains draft while mandatory progress-note interference and streamed reasoning assembly are investigated.


santoshkumarradha:
The frozen benchmark for this PR remains unchanged at source 84ce69bba (Fleet387). #741 (https://github.com/Agent-Field/aforge-v2/pull/741) includes its live context-fidelity fixes and additionally removes the routine completion-reader/threshold-handoff machinery that the data-control audit showed reopening a valid result. It supersedes this PR as the implementation candidate; keep the existing run as a separately labelled comparator. The newer candidate is still draft and has no established Pareto result.

PR #739, opened 2026-09-09
tasks: a section heading says what its fold hides, and stops stating a total the foot disagrees with

AbirAbbas:
## What was true

The tasks place has counted two different partitions of the same rows since conversations
became its structure. The sections file a conversation and everything under it under where
the CONVERSATION stands — its most urgent member's answer — while the foot counts what each
piece of work IS. The heading standing between them claimed to reconcile the two, and could
not: it stated a per-section TOTAL in the foot's own words.

On `make demo-home`, `alt+2`, at `dev@c85e10a1`:

```
tasks · 14 chats · 16 subtasks · $3.32

your call · 2 of 10 shown
› ▾ Sweeping the Frame Budget
    ▸  Measure the frame budget in every package that draws a…   holds 3 more · 55m · 1 file
  ▾ The Tab Bar's Counts
    ▸  Count the tabs the bar draws                              holds 3 more · 2h · 3 files
…
 1 your call · 1 running · 2 waiting · 9 finished today · 3 earlier
```

`10` and `1` for the same two words, one frame apart. `tasksSectionHead`'s own comment said
"Open the fold and the numbers agree by themselves"; they cannot, because the difference is
mostly the filing and only partly the fold. Three things drifted with it in #653:

- `internal/manual/chat/tasks.md` documented the opening heading as
  `tasks · 148 pieces of work since aug 11 · $34.10`. That branch is taken only when the
  page holds NO conversation at all — the form every ordinary machine draws,
  `tasks · 15 chats · 13 subtasks · $2.98`, was nowhere on the page.
- The manual explained the foot/heading difference as folds alone, which is the same false
  claim in prose.
- `railGroupWords` became `{needs you, running, queued, waiting, done}`, and the ALL-CAPS
  law above `app.roomKinWord` still read "AND THAT ONE WORD IS `parked` AND NOT `queued`" —
  a word the table no longer has, forbidding the word that is now its neighbour's. Four
  manual passages still taught the retired roster words, and the roster footer was quoted as
  `1 need you · 3 running` / `148 parked · 12 done` where the code builds
  `3 running · 1 needs you` / `148 waiting · 12 done`.

## What is true now

The foot keeps its basis. That is deliberate and it is the one thing this change refuses to
touch: a conversation stopped on a question with three workers out must not read
`4 your call`, which is the defect #653 fixed at the row level.

What goes is the claim that the two partitions are one number. A section heading now says
only what its folds are holding back — `your call · 8 folded away` — and says nothing when
[...]

PR #743, opened 2026-09-09
Fix the task-states acceptance: a landing on an engine host can be answered

### LINKED ISSUE #706: chat: a top-level landing that is your call draws its reason and no answers row
## What happened

2026-09-09, `dev@52ff7dde2` + `test/task-states-e2e`, on a Linux/arm64 box with
tmux 3.4 and a real `OPENROUTER_API_KEY`:

```
go test -tags e2e -run 'TestTUIE2E/a_refused_landing_is_incomplete' -count=1 -timeout 15m -v ./internal/e2e/
```

The landing card comes up as its own question and then stops:

```
  ? ◆ Review the pull request diff · your call · 42s · 1 file
    nobody could check it
─ main ─────────────────────────────────────────────
 ›
```

There is no third row. No `[a] accept`, no `[n] not right`, no `[s] tell it`, no
`[d] let aforge decide this one`. The tier is right and the reason is right; the
answers are absent, which is the shape `docs/design/task-states/DESIGN.md` was
written to close — "a card with no choices" — in its other half.

The subtest fails at `internal/e2e/tui_e2e_test.go:213`, waiting twenty seconds
for `[n] not right`, and then three presses of `n` walk six rows up the
conversation and type themselves into the box, because the key is correctly
refused on a card that is not offering that answer.

## Replication

**Deterministic (no model).** The fixture is in the repository and the graph is
seeded, so nothing here asks a model anything:

```
make build
go test -tags e2e -run 'TestTUIE2E/a_refused_landing_is_incomplete' -count=1 -timeout 15m -v ./internal/e2e/
```

`seedUndecidedRoot` (`internal/e2e/tmux_test.go`) writes ONE top-level node,
`"state": "unverified"`, `"merge": "inplace"`, and the surface replays it on
attach. What a developer sees today is the two-row card above and the twenty
second wait.

**The comparison that makes it a defect and not a fixture.** In the same binary,
in the same run, `seedUnchecked` (`internal/e2e/taskstates_e2e_test.go`) writes a
node of the same state and merge and its card draws all four columns:

```
  ? ◆ Port the parser · your call · 42s · 1 file
    nobody could check it
    [a] accept · [n] not right · [s] tell it · [d] let aforge decide this one
```

```
go test -tags e2e -run 'TestTaskStatesE2E/your_call' -count=1 -timeout 15m -v ./internal/e2e/   # passes
```

**What has been ruled out, by measurement rather than by reading.**

- Not the keyboard. The row is absent from the frame, not merely unanswered.
- Not the setup screen or the greeting. Both were real and both are fixed on
  `test/task-states-e2e` (`rig.skipSetup`, `statesAnswerKey`), and the failure
  survives them.
- Not the node naming no changed file. That was the one structural difference
  between the two fixtures; adding `"changed": ["diff-review.md"]` put ` · 1 file`
  on the head and left the answers row absent.
- Not the title's width. `app.settleRows` at width 150 draws
  `[a] accept · [n] not right · [d] let aforge decide this one` for both titles
  and for both changed lists when driven directly in `internal/tui3`, so the row
  the real binary is not drawing is a row the layout will draw.

So `app.settleChoices` is coming back empty for this node through the real door
and not through the unit one, which points at what `card.status.Ask` is filled
with on a REPLAYED root — `tasksettle.go`'s `canSettleYes`/`canSettleNo` answer
only for `TaskAskCheck`, `TaskAskHeld` and `TaskAskConflict`.

## Acceptance

End to end, through the door a person uses:

`go test -tags e2e -run 'TestTUIE2E/a_refused_landing_is_incomplete' ./internal/e2e/`
passes on `internal/e2e/tmux_test.go`'s own `seedUndecidedRoot` fixture — the
frame carries `[n] not right` beside `nobody could check it`, and `n` leaves
`you said it is not finished` and a second card reading ` · incomplete`.

A `internal/tui3` unit test is welcome beside it and is not the acceptance: the
same card already draws its chips in that package today.

### LINKED ISSUE #707: chat: the rail's row for a your-call landing names the work and stops, saying no tier word
## What happened

2026-09-09, `dev@57940a813` (immediately after #694), on a Linux/arm64 box with
tmux 3.4 and a real key:

```
go test -tags e2e -run 'TestTaskStatesE2E/the_rail_and_the_roster' -count=1 -timeout 30m -v ./internal/e2e/
```

The conversation's card and the roster page both answer the question. The COLUMN
does not:

```
RAIL · ? Port the parser
```

```
  ? ◆ Port the parser · your call · 42s · 1 file
    nobody could check it
    [a] accept · [n] not right · [s] tell it · [d] let aforge decide this one
```

`/history` draws `your call` on the same node in the same frame. The column
draws the tier cell and the title, and nothing after them — so the one question
every row of work is read to answer first is unanswered on the surface that is
there to answer it at a glance.

`docs/design/task-states/DESIGN.md`, "The rail and the roster":

> The rail row is `<tier glyph> <title> · <word or reason>`, cut from the right.

and, in `internal/tui3/tasktier.go`'s own header for `tierRow`:

```
//	? Port the parser · your call · conflicts with your branch: parser.go
```

The column is thirty cells at this width (`railCols`), and
`Port the parser · your call` is twenty-seven, so this is not the row giving
ground: `tierRowBody` sheds the reason's list first and then cuts the title down
to `tierTitleFloor`, and either of those would still have left `your call` on the
row.

## Replication

**Deterministic (no model).** On `test/task-states-e2e`:

```
make build
go test -tags e2e -run 'TestTaskStatesE2E/the_rail_and_the_roster' -count=1 -timeout 30m -v ./internal/e2e/
```

`seedUnchecked` writes one top-level `"state": "unverified"` node and the surface
replays it on attach. What a developer sees today is the `RAIL · ? Port the
parser` line above and a failure naming it.

## What this looks like from the outside

`tierRow` and `tierRowBody` — the functions `tasktier.go` introduces as "THE ONE
PLACE THIS SURFACE TURNS A READING INTO A CELL AND A ROW" — have no non-test
caller in `internal/tui3`:

```
$ grep -rn 'tierRow(\|tierRowBody(' internal/tui3/*.go | grep -v _test
internal/tui3/tasktier.go:158:func tierRow(...)
internal/tui3/tasktier.go:168:	return lead + tierRowBody(status, title, room)
internal/tui3/tasktier.go:173:func tierRowBody(...)
```

The column composes its rows in `app.railEntryRows` instead — a title row, and
then a second row from `app.railUnder` that carries `session.TaskStatus.RowWord`
and is drawn only when `app.railSaysMore` says so. On this landing neither row
carried the word.

## Acceptance

End to end, through the door a person uses:
`go test -tags e2e -run 'TestTaskStatesE2E/the_rail_and_the_roster' ./internal/e2e/`
passes — the column's own row for a your-call landing carries `your call`, and
where the column is wide enough, the reason after it.

santoshkumarradha:
The four things the real-terminal acceptance of the task-states wave (#698) found, closed.

Closes #706. Closes #707.

## #706 — a landing card with a reason and no answers

The root cause is not the fixture and not the surface's layout. **v3's ordinary launch
talks to an engine host over `internal/remote`**, and `internal/remote`'s agent had none of
the four doors that decide a landing — `ResolveUnverified`, `ResolveConflict`,
`HandUnverifiedToModel`, `TakeBackDecision`. `app.settleDoors()` asserts them on the agent
the window holds, the assertion failed, and the absence law then did exactly what it is
written to do: it removed the whole answers row, leaving

```
? ◆ Review the pull request diff · your call · 42s · 1 file
  nobody could check it
```

and nothing to press.

**The path length is why it read as a fixture bug.** A host is only reachable where its
socket path fits (`enginehost.socketLimit`, 104 bytes), and the temporary home a subtest
runs in is named after the subtest — so `TestTUIE2E/a_refused_landing_is_incomplete` got a
host and drew no chips, while `TestTaskStatesE2E/your_call_asks_in_three_columns…`, whose
name is thirty characters longer, fell back to the in-process engine and drew all four.
Same binary, same run, two answers. On macOS the temp root is long enough that neither ever
reaches a host, which is why the subtest passes on a laptop.

So the four verbs cross: `Task.Settle` carries all five acts, a `Welcome.TaskSettle` flag
says whether the engine behind them has them, and `internal/tui3` asks that flag before it
draws a chip — an engine too old to answer draws the reason and no chips, which is where
the absence law wants it. A reading window is refused by the existing allow-list.

## #707 — "the rail row names the work and stops"

**The rail row was never the row that was measured.** The column stands to the right of the
conversation, so no line of a capture that has one begins with a tier cell —
`statesRailRow` matched on the whole line, found nothing, and `statesRail` then pressed
ctrl+g "because there was no column", closing the very column it was measuring and reading
the tab strip that appears in its place. The column was saying
`your call · nobody could check it` one row below the name the whole time:

```
│ ? Port the parser         #1
│   your call · nobody could
│   check it
```

So the reader reads the column, and a row is read WITH ITS BLOCK — thirty cells will not
hold a name and a reason side by side. `tierRowBody` is split over a new `tierRowParts` so
the two lists of work can paint the title and the reading apart, and a `internal/tui3` unit
test pins the glyph, the name, the word and the reason on the column.

## The conflict shape

It was unreachable, and the engine was right. The repository the harness builds is on
`main`, which is on the protected list, so the landing kept its branch and never merged and
nothing could clash. The subtest now moves the checkout onto `work` first — and leaves the
person's clashing change **uncommitted**, because a commit while the work is out moves the
branch, and a branch that moved since the cut is one aforge will not write either. A
landing that keeps its branch off a plain unmoved checkout is now a failure rather than a
skip.

`docs/design/task-states/DESIGN.md` said `branch kept` appears "only when keeping was asked
for". It appears whenever the landing kept the branch — asked for, a protected or detached
checkout, or one that moved since the cut — and the card's own report says which.

## Stale needles

The setup door's title counts two steps (`setupStepsFor` builds at most two) and the suite
waited for three. `waiting on you` was checked and is **not** stale: `internal/tui3`'s
`notifyAskWord` still spells it and home draws `session.PresenceWaiting`, which is the same
string — so those two subtests' reds have another cause and were left alone.


santoshkumarradha:
**Overlap, for sequencing.** This branch touches `internal/remote/{wire.go,wire_task.go,server.go}`
and #737 (feat/questions-s2) touches `internal/remote/{client.go,server.go,wire.go}`. The regions
are different — #737 edits wire.go's method-const block and `server.invoke` around line 1769, this
adds a field to the `Welcome` struct and a `case MethodTaskSettle` beside `MethodTaskHold` — so
whichever lands second should rebase without a textual conflict. Nothing here touches
`internal/tui3/task.go`, `tasksplace.go` or `room.go`, so there is no overlap with #739.

PR #752, opened 2026-09-10
orchestrate: a bare model id is not a vendor-qualified row's price

### LINKED ISSUE #749: engine: a bare model id is metered at a vendor-qualified row's rate, so an adaptive run's tank drains at a ninth of the unpriced fallback
## What happened

On `dev@1e338386` (2026-09-09), the adaptive run's fuel table prices a **bare**
model id — the id a direct vendor's base actually serves — at the rate of a
**vendor-qualified** row that merely ends with the same name.

At the seam every planner call goes through (`orchestrateCost` in
`internal/session/orchestrate.go`, which falls back to the price table whenever
the provider reports no cost), one million prompt tokens on a bare
`deepseek-v4-flash` bills **$0.14**, the OpenRouter list price of
`deepseek/deepseek-v4-flash`, rather than the **$1.25** the documented
`unpriced` fallback charges for a model the table has never heard of:

```
--- FAIL: TestBareModelCostsTheUnpricedRate (0.00s)
    bare_evidence_test.go:12: a bare id metered at 0.14
```

The tank therefore drains at roughly a **ninth** of the rate the table's own
comment says an unknown model must drain at, so the 80% warning and the 100%
gate arrive nine times too late. The comment above `unpriced` states the law
this breaks: *"A model this table has never heard of is the case where the tank
matters most — nobody has checked what it costs — and metering it at nothing
would leave a run with a cap it can never reach."* A wrong vendor's rate is
worse than no rate, because it looks right.

## Replication

**Deterministic (no model).** From a clean checkout of `dev`, add this file and
run it. Both assertions fail today.

```go
// internal/orchestrate/bare_evidence_test.go
package orchestrate

import "testing"

func TestBareIsNotTheVendorsRow(t *testing.T) {
	if price, known := PriceOf("deepseek-v4-flash"); known {
		t.Fatalf("a bare id priced as a vendor row: %+v", price)
	}
}
```

```
$ go test ./internal/orchestrate/ -run TestBareIsNotTheVendorsRow
--- FAIL: TestBareIsNotTheVendorsRow (0.00s)
    bare_evidence_test.go:7: a bare id priced as a vendor row: {In:0.14 Out:0.28}
```

The same one file down, at the door the run actually meters through:

```go
// internal/session/bare_evidence_test.go
package session

import (
	"testing"

	"github.com/Agent-Field/agentfield/sdk/go/ai"
)

func TestBareModelCostsTheUnpricedRate(t *testing.T) {
	response := &ai.Response{Usage: &ai.Usage{PromptTokens: 1_000_000}}
	if got := orchestrateCost(response, "deepseek-v4-flash"); got != 1.25 {
		t.Fatalf("a bare id metered at %v", got)
	}
}
```

The package's own `TestMeter` asserts the wrong behaviour on purpose today —
`internal/orchestrate/fuel_test.go`, search `a bare model name matches its
vendor's row` — so this is a belief to be invalidated, not a regression.

**Field (real models).** Point aforge at a vendor's own base rather than the
router — `AFORGE_BASE_URL=https://api.deepseek.com`, that vendor's key in
`OPENAI_API_KEY`, `AFORGE_MODEL=deepseek-v4-flash` — and start an adaptive run
with a small cap. No vendor in the current survey returns a dollar figure on
`usage`, so **every** call falls to the table, and the whole run is metered at
another catalog's price for a model this build has never priced. A few cents and
a couple of minutes; the control run is the same goal with a router slug
(`deepseek/deepseek-v4-flash`), whose gauge must not move.

## Where

`PriceOf` in `internal/orchestrate/fuel.go` — the loop under the exact-map
lookup, search `strings.HasSuffix(id, "/"+name)`. It runs after `:`/`@` suffix
cutting and matches a bare `name` against any qualified key ending `/<name>`.
The table it matches against is `prices`, search `deepseek/deepseek-v4-flash`,
and the fallback it steps over is `unpriced`.

Consumers: `Meter` / `MeterCall` in the same file, reached from
`orchestrateCost` and `orchestrateExec.spend` in
`internal/session/orchestrate.go`.

## The fix

**A bare name is not a vendor's row.** An id with no `author/` in front of it
never matches a qualified row; it is a model the table has not priced, and it
gets `unpriced` like any other. A vendor's own serving of a model is not the
router's serving of it, and the two are not obliged to cost the same.

Everything else stands exactly as it is:

- a qualified id — `deepseek/deepseek-v4-flash`, with or without a `:free`
  variant or an `@2026-01` build suffix — prices from its row as before;
- an id matching no row still gets `unpriced` (`{1.25, 10.00}`), which is not
  zero, deliberately;
- no vendor rates are added from memory. Every vendor's model line has rolled
  over since the table was written, and a guessed row is the same defect with a
  different number.

Nothing a person reads changes: the manual describes the tank, the 80% mark, the
gate and the $100 default, and says nothing anywhere about how a model id is
matched to a price. This is a wrong number behind an unchanged sentence.

This is deliberately **not** a redesign of the id grammar — a bare id staying
unpriced is the honest answer until service-qualified identities exist.

## Acceptance

- **e2e:** at the run's own metering door, `orchestrateCost` with a response
  carrying usage and **no** `Cost` returns `1.25` for one million prompt tokens
  on a bare `deepseek-v4-flash` — the `unpriced` input rate — and still returns
  `0.14` for the qualified `deepseek/deepseek-v4-flash`. The two together are
  the fix and its control.
- **e2e:** a response that *does* carry a `Cost` is unchanged: the provider's
  own figure still outranks the table, bare id or not.
- **Unit:** table-driven over the whole of `prices` — every qualified key, and
  the same key with `:free` and `@2026-01` on the end, prices exactly as its row
  says; every one of those keys with its `author/` cut off is unknown and comes
  back `unpriced`; an id in no row comes back `unpriced` with a non-zero `Out`,
  because an unpriced model is not a free model.
- The change entry's `invalidates` names what people believed before — that a
  bare model name matched its vendor's row — and `internal/orchestrate`'s own
  `TestMeter`, which asserted it, asserts the new law instead. No manual page
  states the old behaviour, so none changes.

AbirAbbas:
## What was wrong

`PriceOf` (`internal/orchestrate/fuel.go`) cut a `:variant` or `@build` suffix
off a model id and then walked the price table matching a **bare** name against
any **vendor-qualified** key ending `/<name>`. Every key in that table is
qualified, so a bare `deepseek-v4-flash` — which is the id a vendor's own base
serves, as against the router's `deepseek/deepseek-v4-flash` — resolved to that
row's list price of $0.14 in and $0.28 out.

That is not a missing price, it is a confident wrong one, and it looks right.
The adaptive run's fuel tank meters and gates on this number: at the run's own
metering door a million prompt tokens billed **$0.14** where the documented
`unpriced` fallback charges **$1.25**, so the 80% warning and the 100% gate
arrived some nine times too late. The comment above `unpriced` states the law
this broke — a model the table has never heard of is the case where the tank
matters most, because nobody has checked what it costs.

## What is true now

**A bare name is not a vendor's row.** The suffix-matching loop is deleted. The
exact lookup above it already priced every qualified id, suffixes and all, so a
bare id now falls to `unpriced` like any other model nobody has priced.

The suffix cutting, the `unpriced` figures and every qualified row are
untouched, and **no rate was added to the table** — every vendor's model line
has rolled over since it was written, and a guessed row would be the same defect
with a different number. No shipped default needs one: the shipped chat model
and all five crew-tier models are vendor-qualified ids.

This is deliberately **not** a redesign of the id grammar. A bare id staying
unpriced is the honest answer until service-qualified identities exist.

Nothing a person reads changes. The manual describes the tank, the 80% mark, the
three answers at the gate and the $100 default, and says nothing anywhere about
how an id is matched to a price row — the fallback table is machinery a person
never sees. This is a wrong number behind an unchanged sentence.

## Validation contract

Tests derive from these, not from the implementation.

- **A bare id is not a vendor's row.** `PriceOf("deepseek-v4-flash")` reports
  unknown and returns `unpriced`; likewise the bare form of every key in the
  table, with and without `:free` / `@2026-01`. *Fails on `dev`, passes here.*
- **Every qualified id prices exactly as before.** Table-driven over `prices`
  itself, so a row added later is covered: each key, and each key with a
  `:free` or `@2026-01` suffix, returns its own figures and reports known.
- **An id in no row is still unpriced,** with a non-zero `Out` — an unpriced
  model is not a free model, and no new path to `$0.00` is created. The empty
  and whitespace ids behave as before.
- **The run's own metering door agrees.** `orchestrateCost` with usage and no
  reported `Cost` returns `1.25` for a bare `deepseek-v4-flash` and `0.14` for
  the qualified `deepseek/deepseek-v4-flash`; a response that *does* carry a
  `Cost` still returns that figure verbatim for both — the provider's own number
  still outranks the table.

Proven against `origin/dev` before the fix: the same three tests fail there.

```
--- FAIL: TestMeter
    fuel_test.go:27: a bare model name is not its vendor's row
--- FAIL: TestABareNameIsNotAVendorsRow/bare_deepseek-v4-flash
    fuel_test.go:45: PriceOf("deepseek-v4-flash") = {In:0.14 Out:0.28}, true; want {In:1.25 Out:10}, false
--- FAIL: TestOrchestrateCostKeepsABareNameOutOfAVendorsRow
    orchestrate_cost_test.go:19: orchestrateCost without a reported cost for "deepseek-v4-flash" = 0.14, want 1.25
```

`internal/orchestrate`'s `TestMeter` asserted the old behaviour on purpose
("a bare model name matches its vendor's row"); it asserts the new law instead.
That belief is what the change entry invalidates.

## Gates

All run in the worktree with `GOFLAGS=-buildvcs=false`.

| Gate | Result |
| --- | --- |
| `go build ./... && go vet ./...` | PASS, silent |
| `make test-quick` | PASS — 18 packages `ok` |
| `make test-laws` | PASS — 17 packages `ok` |
| `make test PKGS='./internal/orchestrate ./internal/session' TEST_FLAGS='-count=1'` | PASS — `ok internal/orchestrate 0.236s`, `ok internal/session 270.004s` |
| `make build` | PASS |

No line was added to `.github/known-red.txt`.

Closes #749


AbirAbbas:

Issue #754, opened 2026-09-10
build: an unregistered AFORGE_QUESTION_DEMO leaves internal/config red on a clean dev, so every pull request that touches it opens with a failure it did not cause


AbirAbbas:
## What happened

`TestRegistryCoversEveryUserFacingEnvironmentPin` fails on a clean checkout of
`dev` @ `1e3383869`, with no changes of any kind:

```
$ export GOFLAGS=-buildvcs=false
$ go test ./internal/config/ -run TestRegistryCoversEveryUserFacingEnvironmentPin -count=1
--- FAIL: TestRegistryCoversEveryUserFacingEnvironmentPin (0.03s)
    settings_test.go:111: AFORGE_QUESTION_DEMO (…/internal/tui3/questiondemo.go) is neither a settings row nor operator plumbing — register it in settings.go
FAIL
FAIL	github.com/Agent-Field/aforge-v2/internal/config	0.036s
```

It is not in `.github/known-red.txt`, so the ledger does not skip it and `make
test` does not either. Every pull request that touches `internal/config` — or
any package whose suite the `touched packages` gate pulls that one in behind —
now opens with a red gate it did not cause, which is exactly the signal
`known-red.txt` exists to protect.

## Replication

**Deterministic (no model).** Two commands from a clean tree, under a second, no
key:

```
git switch dev && git pull
GOFLAGS=-buildvcs=false go test ./internal/config/ -run TestRegistryCoversEveryUserFacingEnvironmentPin -count=1
```

The failure quoted above is what a developer sees today.

**Field (real models).** None needed; this is a structural law over the tree.

## Where

`questionDemoEnv` in `internal/tui3/questiondemo.go` declares
`AFORGE_QUESTION_DEMO`, which raises a fixture question on the page. The law that
refuses it is `TestRegistryCoversEveryUserFacingEnvironmentPin` in
`internal/config/settings_test.go` — search `is neither a settings row nor
operator plumbing`. It walks the repository for `AFORGE_*` pins and demands each
one be either a settings row or on the operator-plumbing list, with
`AFORGE_TEST_` the only prefix exempt.

The pin arrived with the questions page (#737).

## The fix

Whoever owns the questions page decides which of the two this is and says so in
`internal/config/settings.go`: operator plumbing, if `AFORGE_QUESTION_DEMO` is a
developer's door onto a fixture and nothing a person would ever set; or a
settings row, if it is not. A third option, if the demo has done its job: delete
[...]

Issue #738, opened 2026-09-09
chat: one ctrl+t in /model throws away a launched thinking level, and the picker walks a shorter ladder than the settings row


AbirAbbas:
## What happened

On `dev@df3153e5b`, a thinking level asked for at launch is thrown away by one keypress in
`/model`, and two of the five rungs the `thinking` settings row offers cannot be reached from
the picker at all.

```
$ aforge --reasoning xhigh
  … status line: model · deepseek/deepseek-v4-flash-latest:xhigh
/model                       # the row reads deepseek/deepseek-v4-flash-latest:xhigh
ctrl+t                       # one press
  … status line: model · deepseek/deepseek-v4-flash-latest      ← the level is gone
```

`internal/tui3/palette.go`'s `reasoningCycle` is `[]string{"", "low", "medium", "high"}` and
`nextReasoning` returns `""` for any word that is not in it. `xhigh` and `max` are not in it,
so the walk does not advance them — it clears them, silently, and the next turn goes out with
no reasoning object on the wire at all. That is the same class of defect #643 closed on the
engine side: a pinned level that does not reach the wire is a knob that does nothing for
whoever set it. Here the surface does the dropping.

The same dial is spelled two ways in the same corpus. `/settings`' `thinking` row and the
task page's own control both walk `auto → low → medium → high → xhigh → max → auto`
(`internal/config`'s `EffortChoices`, `internal/tui3/taskeffort.go`), and
`internal/manual/chat/task-controls.md` says so. `internal/manual/chat/commands.md` and
`internal/manual/chat/models-and-cost.md` say the picker walks
`off → low → medium → high → off` and that it "does not offer `xhigh` or `max`" — two
vocabularies, two spellings for absence (`off` and `auto`), and no gate holding either to the
code.

## Replication

**Deterministic (no model).** Put this in `internal/tui3/pickerladder_repro_test.go` on a clean
checkout and run `go test -count=1 -run TestReproALaunchedLevelSurvivesOneCtrlT ./internal/tui3/`:

```go
package tui3

import "testing"

func TestReproALaunchedLevelSurvivesOneCtrlT(t *testing.T) {
	agent := &fakeAgent{model: "deepseek/deepseek-v4",
		levels: map[string]string{"deepseek/deepseek-v4": "xhigh"}}
	a := newTestApp(agent)
	a.pick.startFor([]Model{{ID: "deepseek/deepseek-v4", Reasoning: true}}, a.model, chatModel)
[...]

Issue #724, opened 2026-09-09
build: the binary outgrew SIZE-BUDGET by nine hundred kilobytes, and `make check` now stops at its last step for everybody


AbirAbbas:
## What happened

On 2026-09-09, `dev@8d8cd6c8`, on a clean checkout, `make check` runs vet, the
formatter, the suite, the packed manual — and then stops at its last step:

```
bin/aforge is 55542025 bytes. The budget in SIZE-BUDGET is 54600000. Over by 942025.

Shrink what you added — pack a corpus (internal/packed), drop a
dependency, stop embedding what can be fetched — or raise the number
in SIZE-BUDGET IN THIS COMMIT, so the extra weight is a decision somebody
signed for rather than a drift nobody saw. PERF.md states the policy.
```

`make check` is the end-of-change ritual `CLAUDE.md` sends everybody to, so
this is red for every change on `dev` today regardless of what that change
touched. Nobody caused it and nobody can clear it from inside their own work.

The last merge that weighed in under the cap is `65f060d3` at 53,633,289 bytes.
Between it and `8d8cd6c8` the binary grew 1,908,736 bytes and crossed a budget
of 54,600,000 that nobody moved.

## Replication

**Deterministic (no model).**

Any clean checkout of `dev`, any of the four shipped platforms, Go 1.26.5:

```
make size
```

Today that prints the block above and exits 1. `make build` alone is green —
the ratchet gates `check` and not `build` deliberately — so a developer only
meets it at the end of a change that is otherwise finished.

The comparison that dates it, in two detached worktrees:

```
git worktree add --detach ../weigh-base 65f060d3 && (cd ../weigh-base && make build && stat -c%s bin/aforge)
git worktree add --detach ../weigh-head origin/dev && (cd ../weigh-head && make build && stat -c%s bin/aforge)
```

`make build` and not `go build`: the build fetches the pinned furrow into
`internal/furrowbin/cache/` and `//go:embed` links it in. A bare `go build`
yields about 51.7 MB on linux/amd64 and reads falsely green.

**Field (real models).** None. No key, no network beyond the furrow fetch, no
model. It is arithmetic on a file the build produced.

## Where

`SIZE-BUDGET`, one integer, byte-identical since #235. The comparison is the
`size` target in `Makefile` (search `Over by`), and the policy it points at is
the section headed `## The size ratchet` in `PERF.md`.

What actually grew, measured between the two binaries above (linux/amd64,
Go 1.26.5):

[...]

PR #757, opened 2026-09-10
config: one door builds a provider client, and a thinking level stops going on the wire as part of the model id

### LINKED ISSUE #750: config: six clients are built beside the profile's own factory, so a thinking level goes on the wire as part of the model id and a standing pass routes with no price ceiling
## What happened

`internal/config` has one place that turns a profile into provider settings —
`Config.providerConfig(model)`, reached by `Client`, `ClientFor`, `MediaClient`,
`VisionClient` and `DocumentClient`. It splits a thinking level off the model id,
carries the key and the base URL, and hands the adapter the three catalog seams
(`SupportsParameter`, `ReasoningProfile`, `ModelPrice`).

Six production sites do not go through it. They compose a `provider.Config`
themselves out of a copied key and a copied base URL:

```
$ git grep -n 'provider\.Config{' -- '*.go' | grep -v _test.go
cmd/aforge/chatv3_harness.go:123:	client, err := provider.NewClient(provider.Config{
cmd/aforge/chatv3_subharness.go:80:	client, err := provider.NewClient(provider.Config{
internal/config/config.go:826:	return provider.Config{
internal/session/agent.go:47:	client, err := provider.NewClient(provider.Config{
internal/session/memory_consolidate.go:233:			client, built = provider.NewClient(provider.Config{
internal/session/standing_run.go:1024:			client, built = provider.NewClient(provider.Config{
internal/session/tools_doc.go:98:	client, err := provider.NewClient(provider.Config{
```

Six of the seven are the copies. Two consequences are live today, not merely
untidy:

1. **A thinking level goes on the wire as part of the model id.** `moonshotai/kimi-k3:low`
   is how a tier row, `--plan-model` and `AFORGE_PLAN_MODEL` all say "that model,
   thinking a little". `providerConfig` splits it — bare slug in `Model`, level in
   `Effort` — and says in as many words that sent whole it is "a slug no provider
   publishes, which is a 404 on every planning call". The conversation's own client
   (`agent.go`), the document rung, the harness runner and the subharness runner all
   send it whole.
2. **A standing pass and a memory consolidation route with no catalog seams.** Neither
   carries `ModelPrice`, so neither sends the `provider.max_price` ceiling every other
   request on this machine sends, and `latencyPriceCeiling`'s own comment says what that
   costs: a latency ask with no ceiling was measured paying ~3.5x list across 44 bench
   cells. Neither carries `SupportsParameter`, so a knob no endpoint publishes is a 404
   with no endpoints left rather than a field that was never sent.

And the standing defect underneath both: the day this package learns a second base URL,
a second key, or a header policy, a harness, a subharness, a standing run, a memory
consolidation and a document read keep talking to the old one.

## Replication

**Deterministic (no model).** Two clients, one model id, one stub server — no key, no
network. Drop this in `internal/session` as `wire_probe_test.go` and run
`go test ./internal/session/ -run TestWireProbe -count=1`:

```go
func TestWireProbe(t *testing.T) {
	var body []byte
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		body, _ = io.ReadAll(r.Body)
		w.Header().Set("Content-Type", "text/event-stream")
		io.WriteString(w, "data: [DONE]\n\n")
	}))
	defer srv.Close()

	agent, err := session.New(session.Config{
		Workspace: t.TempDir(), Model: "stub/model:low",
		APIKey: "probe-key", BaseURL: srv.URL,
	})
	if err != nil { t.Fatal(err) }
	_, _ = agent.Send(context.Background(), "hello")
	if strings.Contains(string(body), `"model":"stub/model:low"`) {
		t.Fatalf("the level went on the wire as part of the id: %s", body)
	}
}
```

Today it fails, and the body it prints carries `"model":"stub/model:low"`. The same
model id through `config.Config{Model: "stub/model:low", BaseURL: srv.URL}.Client()`
sends `"model":"stub/model"` with the level in the request's reasoning field — the two
doors disagree about the same string.

**Field (real models).** Needs `OPENROUTER_API_KEY`, a few cents, under a minute. Pin a
role to a tier value that carries a level (`/settings`, or `AFORGE_PLAN_MODEL=<slug>:low`),
open `aforge`, and read a document or let a standing item check. The refusal names a slug
with a colon in it and no endpoint serving it. The control is the same slug without the
level, which answers.

## Where

`providerConfig` and its five callers in `internal/config/config.go`; the six composite
literals the grep above lists — search `provider.NewClient(provider.Config{`. In
`cmd/aforge`: `v3RunHarness` (`chatv3_harness.go`) and `v3Subharnesses`
(`chatv3_subharness.go`). In `internal/session`: `New` (`agent.go`), `newDocClient`
(`tools_doc.go`), `NewStandingSentinel` (`standing_run.go`), and the consolidation seam
(`memory_consolidate.go`).

## The fix

**A provider client is assembled in exactly one package, and that package is
`internal/config`.** A site that needs its own client says which model, how long it may
take, and what its own surface owns — never the key and never the base URL.

`internal/config` grows one small exported door for callers that already hold a profile,
and one for `internal/session`, whose own `Config` carries the account. Every one of the
six sites builds through it, so all six inherit the level split and the catalog seams,
and all six move together the day a second service arrives.

Nothing else changes: same base, same headers, same key precedence, same model for a
plain model id and an empty catalog.

## Acceptance

- **e2e:** through the real door at each of the six sites, against an `httptest` stub —
  `session.New`, `newDocClient`, `NewStandingSentinel`, the consolidation seam,
  `v3RunHarness`, `v3Subharnesses`. Each asserts the request reached the stub's base with
  the configured bearer, and that a model id carrying a level arrives as the bare slug
  with the level in the reasoning field. Each fails on `dev` today.
- **e2e:** the control — a plain model id and no catalog seams produce byte-for-byte the
  request they produce today: same base, same headers, same model, no ceiling invented.
- **Unit:** a structural law, run by `make test-laws`: no non-test file outside
  `internal/config` composes a `provider.Config` that sets `APIKey` or `BaseURL`.

AbirAbbas:
## What was wrong

`internal/config` has one place that turns a profile into provider settings, and six
production sites did not use it. Each composed a `provider.Config` out of a copied key
and a copied base URL: the harness runner, the subharness runner, the conversation's own
client, the standing sentinel, the memory consolidation and the document read.

Two things were live defects, not untidiness:

- **A thinking level went on the wire as part of the model id.** `slug:low` is how a tier
  row, `--plan-model` and `AFORGE_PLAN_MODEL` all say "that model, thinking a little".
  The profile's door splits it — bare slug in `Model`, level in `Effort` — and its own
  comment says that sent whole it is a slug no provider publishes. Four of the six sent
  it whole.
- **A standing pass and a memory consolidation routed with no catalog seams.** No
  `ModelPrice`, so no `provider.max_price` on a latency-sorted ask — the exact shape
  `latencyPriceCeiling` was written for after a cost autopsy measured ~3.5x list across
  44 bench cells. No `SupportsParameter`, so a knob no endpoint publishes becomes a 404
  with no endpoints left rather than a field that was never sent.

And underneath both: the day this package learns a second base, a second key or a header
policy, those six keep talking to the old one.

## What is now true

`internal/config` exposes three named doors and no unexported assembly:

- `ClientConfigFor(apiKey, baseURL, model)` — the one place a key and a base URL become
  a `provider.Config`, and where the level split lives. It is what `internal/session`
  reaches, because a session's own `Config` carries the account and no profile.
- `Config.ClientConfig(model)` — that, plus what a profile knows: the timeout and the
  three catalog seams. The five clients this package builds all go through it.
- `WithoutSeatPin(...)` — drops the seat's level for a raw request path, so
  `DocumentClient`'s reason for clearing it is written once rather than once per caller.
  It is exported because the session's document read needs the same shape and may not
  spell the adapter's effort vocabulary itself (its own `effortguard_test.go` law).

The four `internal/session` sites go through one private `Config.clientConfig`, which
adds the timeout and the seams the session owns. The two `cmd/aforge` doors use
`settings.ClientConfig(model)` and set only their own `harnessTimeout`. `internal/session`
never spells an `APIKey` or a `BaseURL` field again.

One consequence worth naming: a session launched on `slug:low` now keeps the bare slug as
its own model, because the turn passes that model on every request and would otherwise
put the suffix back after the client door split it off. The level still reaches the wire,
as the client's configured effort. Any surface that reads the session's model back — the
status row, the status sheet — therefore shows `slug` where it used to show `slug:low`.
That is the honest reading of "the level is not part of the id", and #738 / #745 are the
lanes that own how a launched level is displayed and walked.

## Validation contract

Each item is a test through a real door against an `httptest` stub. **C1–C7 were run
against a clean detached checkout of `origin/dev` and every one of them fails there**;
C8 passes there, which is what makes it the control.

- **C1** `TestAConversationSendsTheBareSlugAndCarriesTheLevelBeside` — `session.New`
  reaches the stub's base with its bearer and sends the bare slug with the level beside
  it. On dev: `model = "stub/conversation-client-door:low"`.
- **C2** `TestADocumentReadSendsTheBareSlugThroughTheSessionsAccount` — same through
  `newDocClient`, no reasoning object invented, and the empty-key refusal still reads
  exactly `this session has no API key`. On dev: `model = "stub/document-client-door:low"`.
- **C3** `TestAStandingPassCarriesTheSessionsPublishedPriceCeiling` — the sentinel's
  request carries the ceiling derived from the session's price seam. On dev: no
  `provider.max_price` at all.
- **C4** `TestAMemoryConsolidationCarriesTheSessionsPublishedPriceCeiling` — the same
  through the consolidation seam. On dev: no ceiling.
- **C5** `TestAHarnessRunSendsTheBareSlugThroughTheProfilesAccount` — on dev:
  `model = "stub/harness-client-door:low"`.
- **C6** `TestASubharnessRunSendsTheBareSlugThroughTheProfilesAccount` — on dev:
  `model = "stub/subharness-client-door:low"`.
- **C7** `TestOnlyConfigTurnsAnAccountIntoProviderSettings` — a `go/ast` law over the
  tree, on `make test-laws`: no non-test file outside `internal/config` composes a
  `provider.Config` that sets `APIKey` or `BaseURL`. On dev it names all twelve lines
  across the six files.
- **C8** `TestTheSessionClientDoorsLeaveAnOrdinaryRequestAlone` and
  `TestTheHarnessClientDoorsLeaveAnOrdinaryRequestAlone` — the control: a plain model id
  with no catalog seams produces the request it always did, at all six doors. Green on
  dev and green here.

## Gates

Run in the worktree with `GOFLAGS=-buildvcs=false`, on the rebased branch.

| Gate | Result |
| --- | --- |
| `go build ./... && go vet ./...` | clean, no output |
| `gofmt -l` over every changed file | clean, no output |
| `make test-quick` | pass |
| `make test-laws` | pass |
| `make test PKGS='./internal/config ./internal/session ./cmd/aforge' TEST_FLAGS='-count=1'` | `internal/session` ok, `cmd/aforge` ok; `internal/config` fails only on the pre-existing red below |
| `make build` | pass |

**Two reds are pre-existing on a clean `dev` and are not this change.** Both were
reproduced on an untouched detached checkout before this branch existed:

- `TestRegistryCoversEveryUserFacingEnvironmentPin` — `AFORGE_QUESTION_DEMO` is declared
  in `internal/tui3/questiondemo.go` and registered nowhere. Filed as #754. It is not in
  `.github/known-red.txt`, so every pull request touching `internal/config` opens with
  it. Both files are outside this change's scope and nothing here was added to the
  known-red ledger.
- The `SIZE-BUDGET` ratchet — #724. `dev` builds 56,168,713 bytes against a budget of
  54,600,000. This branch builds **smaller** than the base it sits on.

Closes #750


santoshkumarradha:
There is one explicit-opt-in hole in the standing and memory paths. Both still resolve their role with `roles.Resolve(...)`, whose contract returns `ResolveCall(...).Model` and deliberately drops the `Call.Effort`. Passing that already-bare `model` to `parent.clientConfig(model, ...)` means a role pin or tier value such as `roles.standing=vendor/model:low` / `tiers.<tier>=vendor/model:low` emits neither the suffix nor a `reasoning` field. The session-default case happens to work because `Ladder` deliberately does not split that fallback, so the current wire tests do not expose the role/tier case.

Please resolve these two private-client seats with `roles.ResolveCall`, give `call.String()` to `clientConfig` (and keep `call.Model` for the request/model bookkeeping), with one real-wire regression covering a suffixed role or tier setting. That preserves the operator opt-in while keeping ordinary requests free of app-imposed reasoning fields. It also makes the PR's “thinking level leaves the model id” claim true for all six named doors.

AbirAbbas:
`touched packages` is red here on `TestRegistryCoversEveryUserFacingEnvironmentPin`, which is not this change: `AFORGE_QUESTION_DEMO` has been unregistered on `dev` since #733 and the test is in no known-red ledger, so every pull request touching `internal/config` opens with it. Reproduced on an untouched detached checkout of `dev` before this branch existed.

Filed as #754 and fixed in #762. Once that merges I will rebase this and the gate should be green — `internal/session` and `cmd/aforge` are already ok in the same run, and `light gate` passed.

The `SIZE-BUDGET` ratchet (#724) is also red on `dev`; this branch builds smaller than the base it sits on.

AbirAbbas:

PR #758, opened 2026-09-10
Questions: consent, the task proposal, the stop card and the close-a-tab card move onto the block

santoshkumarradha:
The migration wave's fourth and fifth blocks: the two cards a person raises with
their own hand. Both keep every law they had, and both stop drawing themselves.

## What moves

`internal/tui3/stop.go` and `internal/tui3/tabclose.go` drew their own cards —
two layout loops, two cursors, two key sets, two sets of pointer targets. They
draw nothing now. The question block (`internal/tui3/question.go`) draws both as
`session.AskConfirmation`, and what is left in each file is the LANE'S: which
work `x` was pressed at and what stopping it does not take away; which tab
`ctrl+w` was pressed on and what that conversation had running.

```
?  Stop this job?
     The process is ended; its log is kept.
     1  stop it
     2  keep going
   [enter] take the pick · [esc] keep going · [←→] pick
```

```
?  Close this tab? the tree walk is working · 2 tasks running
     nothing here is deleted
     1  keep running  it keeps going here; find it under Chats, and ctrl+shift+t brings the tab back
     2  stop work     the reply, tasks and jobs stop; nothing is deleted
     3  cancel        nothing changes
   [enter] take the pick · [esc] cancel · [←→] pick
```

## What it costs

`k` and `s` are not keys on the close-tab card. They answered it outright — one
keystroke, no cursor, work ended — which is the bypass that card was built to
not have. Every answer carries its own digit now, a digit MOVES THE CURSOR onto
the answer it names, and `enter` is still what decides.

## What the block needed

Four things, all of them about a question the person raised rather than one that
arrived:

- it sorts to the FRONT of the queue (`questionRaisedHere`), because `esc` on a
  card raised over a five-minute-old permission was folding the permission;
- neither the draw stamp nor the 250ms settle guard applies to it;
- a half-typed sentence does not outrank it — `enter` belongs to the card;
- it leaves no receipt, because the act is the receipt.

## Two defects found on a real screen

- `stopTarget.question()` returned the question and the promise as one sentence,
  which the block then drew twice (head row, reason row). They are separate now.
- A job's page takes the frame whole, so the card `x` raised over it was a
  question nobody could see, holding the keyboard. The page steps aside for it.

## Not in this PR

The standing card, the harness feed card, `pickrow.go` and the takeover confirm.
Each is blocked on a file another lane owns; the lane report says which and why.


santoshkumarradha:
At current head `30d09ac81`, `internal/session/question.go:1450` treats every text-only task answer as `Approved: true`. The new `TestEveryTypedSentenceIsACorrectionAndNotAHiddenAnswer` explicitly confirms that typing `no`, `stop`, or `don't` and pressing Enter starts the task with those words appended to its brief. This changes the previous refusal behavior into execution authorization.

Please preserve complete refusal answers (including case/punctuation variants) as declines; longer corrections such as `no, use the flag` can still approve the corrected brief. An explicit refusal should never authorize work merely because it arrived through the text box. This does not affect our unattended benchmark driver, which uses `ResolveConsent`, but it is material for the interactive task door.

PR #759, opened 2026-09-10
catalog: a cache belongs to the base it came from, so a DeepSeek key stops being served OpenRouter's rows

### LINKED ISSUE #748: models: the model catalog cache does not know which base it came from, so yesterday's OpenRouter rows are served to today's DeepSeek key
## What happened

2026-09-09, `dev@1e3383869`, a freshly built `bin/aforge`. The model catalog is cached at
`<state root>/model-catalog.json` and **the file says nothing about which base URL it was
fetched from**. So a machine that talked to OpenRouter yesterday serves OpenRouter's rows
to a completely different vendor today, and every capability the belt derives from the
catalog — draws, speaks, sings, films, sees — is derived from another service's facts.

Pointed at DeepSeek, with an OpenRouter catalog already on disk:

```
$ AFORGE_HOME=$H AFORGE_BASE_URL=https://api.deepseek.com \
  AFORGE_MODELS=bytedance-seed/seedream-5-0-pro aforge models
panel:
  bytedance-seed/seedream-5-0-pro  sees · draws
```

DeepSeek has no image endpoint and has never heard of that id. No network call was made:
the cache was inside its TTL, so the row was served straight off disk.

The same is true with no cache at all. The compiled-in `hardcodedFallbacks()` are eleven
OpenRouter ids — Seedream, Krea, Fish Audio, Kokoro, Lyria, Seedance — and they are served
for *any* base whose fetch fails:

```
$ AFORGE_HOME=$(mktemp -d) AFORGE_BASE_URL=https://127.0.0.1:1/v1 \
  AFORGE_MODELS=bytedance/seedance-2.5 aforge models
panel:
  bytedance/seedance-2.5  sees · films
```

That is the opposite of the standing law that a capability which cannot work is *absent*,
not broken: the verb goes on the belt, the model calls it, and the call cannot succeed.

## Replication

**Deterministic (no model, no network, no key.)** Three states, one binary:

```sh
make build

# 1. A FRESH cache from another base is served whole.
H=$(mktemp -d)
cat > "$H/model-catalog.json" <<'JSON'
{"fetched_at":"2999-01-01T00:00:00Z",
 "models":[{"id":"bytedance-seed/seedream-5-0-pro",
            "input_modalities":["text","image"],"output_modalities":["image"]}]}
JSON
AFORGE_HOME=$H AFORGE_BASE_URL=https://api.deepseek.com \
  AFORGE_MODELS=bytedance-seed/seedream-5-0-pro bin/aforge models
# today:  bytedance-seed/seedream-5-0-pro  sees · draws
# wanted: no modality word — this base has never published that id

# 2. A STALE cache from another base is served when the fetch fails.
#    Same file with "fetched_at" two days old and AFORGE_BASE_URL=https://127.0.0.1:1/v1.
# today:  bytedance-seed/seedream-5-0-pro  draws

# 3. NO cache, failed fetch: the compiled-in OpenRouter rows are served anyway.
AFORGE_HOME=$(mktemp -d) AFORGE_BASE_URL=https://127.0.0.1:1/v1 \
  AFORGE_MODELS=bytedance/seedance-2.5 bin/aforge models
# today:  bytedance/seedance-2.5  sees · films
```

All three print a capability sentence for a service that does not have the capability.
`aforge models` is only the cheapest window on it; the same rows arm the media hands, the
`/model` picker and the belt.

**Field (real models).** `OPENAI_API_KEY=<a DeepSeek key> AFORGE_BASE_URL=https://api.deepseek.com
AFORGE_MODEL=deepseek-v4-flash aforge do "draw me a red circle" --dir "$(mktemp -d)" --json
--yes-spend --timeout 5m` on a machine that has used OpenRouter before. Under a minute and
a fraction of a cent. `generate_image` is on the belt because the catalog says a drawing
model exists; the call then goes to an endpoint DeepSeek does not serve.

## Where

`internal/catalog/catalog.go`. Search `cacheName` — the cache file name is the constant
`"model-catalog.json"` with nothing else in it, and `cachePath` joins it to the profile
directory. `load` reads it before it fetches (search `readCache(path)`), serves it again
when the fetch fails (search `The network is gone and the cache is old`), and otherwise
returns `hardcodedFallbacks`. Nothing on any of those three roads compares
`Options.BaseURL` with the base the rows came from, because the rows do not carry one.

`internal/config/config.go`, search `DefaultBaseURL`, is where the default base is spelled;
`AFORGE_BASE_URL` is the door that changes it.

## The fix

**A cache belongs to the base it was fetched from, and is never served to another one.**
The rows written to disk carry the normalised base they came from, and a load whose base
does not match that key treats the file as absent — the same as a cold machine.

**The compiled-in fallback rows are OpenRouter's, so they are served only for the default
base.** For any other base with no usable cache and a failed fetch the catalog is *empty*.
That is the correct answer under "a capability that cannot work is absent, not broken":
with no drawing model advertised, `generate_image` is not on the belt at all, and the
picker has nothing to offer rather than eleven names the service has never heard of.

**Nobody is made to refetch by the upgrade.** A `model-catalog.json` written by today's
build has no base in it; it is read as the default base's cache, which is what it almost
always is. Existing installs keep their picker and pay no cold fetch.

Nothing a person reads changes on the default base — same picker, same prices, same
startup cost, no extra network round trip on the fresh-cache path.

## Acceptance

- **e2e:** the deterministic recipe above, states 1–3, through `bin/aforge models` with
  `AFORGE_BASE_URL=https://api.deepseek.com`: the panel line for a row that exists only in
  the other base's cache prints **no** modality word. Exit code 0.
- **e2e:** the control — the identical recipe with `AFORGE_BASE_URL` unset (or the default
  OpenRouter base) still prints `sees · draws`, from the same cache file written by the
  previous build. The upgrade costs nobody a refetch.
- **Unit:** in `internal/catalog`, all four cache states (none; fresh from base A; stale
  from base A; base A's cache plus a failing fetch from base B) — a catalog loaded for
  base B contains no row that came from base A. And a non-default base with no cache and a
  failed fetch is empty, so `Config.ResolveImageModel` and its three siblings return `""`.
- **Unit:** the default base is named in exactly one place and interpolated everywhere
  else, so `config.DefaultBaseURL` and the catalog's notion of "the base the built-in rows
  describe" cannot drift apart.
- No manual page states anything about the catalog cache file or promises the generation
  verbs unconditionally, so no page changes; the change entry's `invalidates` says that a
  catalog cache is no longer shared across base URLs.

AbirAbbas:
## What was wrong

The model catalog was cached at `<state root>/model-catalog.json` with **no base URL in the
key and no base recorded inside the file**. `load` served that one file before it fetched,
served it again when the fetch failed, and otherwise served `hardcodedFallbacks()` — eleven
OpenRouter ids (Seedream, Krea, Fish Audio, Kokoro, Lyria, Seedance) advertising image,
speech, music and video output — for **any** base.

So a machine that talked to OpenRouter yesterday served OpenRouter's rows to whatever
`AFORGE_BASE_URL` named today, with no network call at all. Verified at a real door on
`dev@1e3383869`:

```
$ AFORGE_HOME=$H AFORGE_BASE_URL=https://api.deepseek.com \
  AFORGE_MODELS=bytedance-seed/seedream-5-0-pro bin/aforge models
panel:
  bytedance-seed/seedream-5-0-pro  sees · draws
```

The belt is built from the catalog, so this put `generate_image`, `speak`, `generate_music`
and `generate_video` in front of a vendor with no such endpoints — the opposite of the
standing law that **a capability that cannot work is absent, not broken**.

## What is now true

- The **normalised base** (trimmed, trailing slashes dropped, scheme and host lowercased; an
  empty base is the default base) is stamped into every cache written, checked on every cache
  read, and mixed into the file name for anything but the default base
  (`model-catalog-<16 hex of sha256>.json`). A cache from one base is never served to another.
- The **compiled-in rows are the default base's and nobody else's.** Any other base with no
  usable cache and a failed fetch gets an **empty** catalog, so the four generation verbs are
  simply not on the belt.
- **Nobody pays for the upgrade.** The default base keeps the file name
  `model-catalog.json`, and a legacy file with no `base` key is read as the default base's
  cache — so an ordinary install's picker and prices survive with no cold refetch. The trade
  is written down in the code: an install already running a custom base has a legacy file
  under the default name, and it can be read once as the default's, bounded by the TTL.
- `ResolveMusicModel` was the one generation slot of four that **invented** a name
  (`google/lyria-3-clip-preview`) when the catalog advertised nothing. It now answers nothing,
  like image, speech and video.
- `config.DefaultBaseURL` reads `catalog.DefaultBaseURL`: the built-in rows are that service's
  ids, so the package that serves them owns the constant and the literal is spelled once.

Nothing a person on the default base sees changes — same picker, same prices, same startup,
no new network round trip. `internal/manual/chat/making-pictures-audio-and-video.md` already
says "These tools are only there when a model behind them is … a kind with no model available
simply has no tool, rather than a tool that refuses", so **no manual page changes.**

## Validation contract

Each new test was run against an untouched `origin/dev` worktree before the fix, and the ones
that must fail there do.

| Behaviour | Test | On dev |
| --- | --- | --- |
| In four cache states — none; fresh from base A; stale from base A; base A's cache plus a failing base B fetch — a catalog for base B carries no base A row. A file whose recorded base disagrees is refused. | `TestACatalogCachedForOneBaseIsNeverServedToAnother` | FAILS (3 of 5 subtests) |
| The default base is unchanged: fresh cache served with zero fetches, stale cache served with its own `FetchedAt`, otherwise all eleven built-ins with image, speech, music and video present. | `TestTheDefaultBaseKeepsItsCacheAndOfflineFallbackBehavior` | passes (as it must) |
| A non-default base with no cache and a failed fetch is empty — no model, and nothing for any of the four generation modalities. | `TestTheBuiltInRowsAreTheDefaultBasesAndNobodyElses` | FAILS — "has 11 models, want none" |
| At the belt seam, that empty catalog leaves every generation slot unnamed. | `TestAnEmptyCatalogLeavesEveryGenerationModelUnresolved` | FAILS — resolved Seedream, Fish Audio, Lyria, Seedance |
| A legacy `model-catalog.json` in exactly today's shape is served for the default (and empty) base with no fetch, and is not served to a custom base. | `TestALegacyCacheFileIsTheDefaultBasesCache` | FAILS on the negative half |
| A cache records its normalised base; equivalent spellings share one file; the default base's file name is unchanged. | `TestACacheRecordsItsNormalizedBase` | FAILS |
| Startup does not grow: the default-base fresh-cache path performs zero HTTP requests. | asserted inside the default-base test | passes |

## The launch ratchet moved down, and that is the second commit

`deadCatalogEndpoint` points `AFORGE_BASE_URL` at a refusing httptest server, so until this
change the launch was handed the eleven compiled-in rows **through the leak this PR closes** —
and arming the five media hands is what took it to sixteen blocking catalog reads. With the
custom base correctly empty the same launch asks **eleven**, logging `media: the image
fallback names bytedance-seed/seedream-5-0-pro, which cannot draw — passing over it` four more
times. Serving the five capability rows from the fixture instead was tried first and lands on
fifteen, not sixteen, because an explicit vision row ends that search earlier than the
compiled-in list does; a sixth row would only be forcing the counter. So the constant moved,
and `PERF.md` moved with it in the same commit, saying plainly what the smaller figure covers
and what it no longer covers.

## Gates

Run in the worktree with `GOFLAGS=-buildvcs=false`, on the branch rebased onto `dev@49e80d5fc`:

- `go build ./...` — clean. `go vet ./...` — clean.
- `make test-quick` — all packages ok.
- `make test-laws` — all packages ok.
- `make test PKGS='./internal/catalog ./internal/config ./internal/command ./cmd/aforge' TEST_FLAGS='-count=1'` —
  `internal/catalog` ok, `internal/command` ok, `cmd/aforge` ok (all four launch laws pass:
  1 subharness read, 11 launch reads, 0 unpacks twice). `internal/config` fails on
  **`TestRegistryCoversEveryUserFacingEnvironmentPin` only**, which fails identically on an
  untouched `origin/dev@1e3383869` and is already filed as #754.
- `make build` — clean.
- `make size` — over budget by 1,581,001 bytes, and **this branch is byte-identical to its
  rebase base** (56,181,001 on both), so the overage is entirely #724.

Closes #748


AbirAbbas:
CI note: `touched packages` is red on **`TestRegistryCoversEveryUserFacingEnvironmentPin` only**, and it is not this change.

```
--- FAIL: TestRegistryCoversEveryUserFacingEnvironmentPin (0.03s)
    settings_test.go:111: AFORGE_QUESTION_DEMO (internal/tui3/questiondemo.go) is neither a settings row nor operator plumbing — register it in settings.go
```

It fails identically on an untouched `origin/dev` worktree at both `1e3383869` and `49e80d5fc`, it is not in `.github/known-red.txt`, and it is already filed as #754. Every other package in the gate is green — `cmd/aforge` ok (all four launch laws pass), `internal/catalog` ok — and `light gate` passes. Left for #754's owner rather than fixed here, since #757 is open over the same file.

AbirAbbas:

PR #741, opened 2026-09-09
Simplify completion and preserve actual execution context

santoshkumarradha:
Ordinary answers could trigger another model to judge completion, be reopened because no tracked file changed, or move into a new task after a write/round/time threshold. Separately, twelve quiet tool rounds triggered a demand for a progress note and held further tools. Captured writing and CPF traces then ended on that requested progress note without producing the deliverable.

Restore the ordinary main-model → tools → results → final-response loop. Remove automatic routing/completion readers, threshold-driven handoffs, keyword delegation refusals and mandatory progress-note enforcement, including their unused registries and counters. Keep explicit tasks/forks/task division, actual results, declared checks, budgets, cancellation, retries, recovery and existing context reduction. Do not delete external output merely because it appeared during a turn. Historical stopped-task and journal records remain readable.

Context fixes #717 and reasoning assembly #746 are now on dev and integrated here. CRITICAL comments and regressions protect batch-local tool identity, completed reasoning blocks and tool execution independent of status prose. The model receives the original request and actual tool results; no new classifier, completion protocol, model override or task-specific rule is added. Manual, performance contract and change entry are updated.

Deterministic regressions cover a direct answer with zero side calls, 45 writes without a forced handoff, and 18 distinct quiet tool rounds followed by a write/final in exactly20 main calls and zero side calls. The quiet-work regression fails on the previous source at round6 and passes after removing the note rule. Explicit task/fork wakes, steering, cancellation and context/replay tests remain.

Current candidate source: `570207416a9e61eb9fc437b6252e5cc2ae3c28d5`, integrated with dev `5c4ee5362`. Quick checks (build/vet/format/packed manual/laws) pass; full uncached affected suites pass (session273.165s, manual2.039s, roles0.282s, e2e0.411s), native/static builds pass, and all CI checks are green. The separate static build and zero-paid-call preflight authenticate a new immutable14-case Spark cohort: the same five GitHub issues × two seeds plus four writing/data controls, exact `deepseek/deepseek-v4-flash-0731`, provider generation defaults, authenticated fixtures and unchanged graders.

**Keep this draft pending quality evidence.** The previous frozen `8ef946a5d` experiment lost quality on writing and CPF despite passing deterministic/CI checks. Its benchmark388 and context-only benchmark387 remain unchanged; failures remain in the comparison. The next candidate gets its own source, binary and output directory. No Pareto claim is made; Pi/Mini comparisons use matching completed cases and missing costs remain lower bounds.

[...]

PR #760, opened 2026-09-10
A thinking model is not a silent one, and no turn ends with nothing said

santoshkumarradha:
A chat turn on 2026-09-09 thought for 108 seconds at the fastest rate on the wire
that afternoon and then stopped: no answer, no error, no note, an idle status
line, and nobody able to say what had ended it. Five defects, one autopsy.

## The ceiling measured the wrong silence

`internal/lane/control`'s ceiling ran from the last **visible** token. A run of
reasoning produces none — hidden deltas move `delta` and the phase and never
`progress` — so every model that thinks for longer than a person's patience
reached the ceiling at exactly the ceiling, while it was streaming at full rate.
The measured turn wrote 6,174 reasoning tokens in 108 seconds and was reported
as a stall ten seconds in. The drift clock in `assess` already read the wire for
precisely this reason and said so in its own comment; the ceiling overrode it.

The ceiling now bounds a **still wire** — the time since the endpoint last wrote
anything at all, readable or not — with the one exception the rate-collapse
clock already owned: a visible rate that has collapsed against the lane's own
gap belief is a person waiting whatever the wire is doing. `firesAt`,
`Deadline` and `verdict` all read the same origin, so the search and the verdict
still cannot drift apart. Nothing is loosened: a heartbeat is not writing and
never moves `delta`, so a router holding the line open by saying nothing in a
well-formed way is bounded exactly as before, and the moment the deltas stop the
ceiling runs from there.

No new figure. The role's ceiling was always documented as a statement about
people; it now judges only the silence people are actually in.

## Infinity reached decisions and rows

`hazard.cost()` answered `+Inf` with no alternative or with nobody waiting, and
`Survival.Remaining` answered `+Inf` far enough into the tail. An infinity is a
number, so it crossed `wait > cost + margin` by construction, reached `Act`, and
reached the model-call log — where JSON cannot spell it, so `internal/calllog`
took the row's own figure off it and wrote a sentence about a float instead.

W(s) and A are `control.Seconds` now: a figure, nothing, or past what a belief
can price. `Seconds.Over` is the only comparison the package makes and it needs
both sides before it can say yes. `finite.go` keeps its promise that a value
never costs a row, drops the sentence — under the emptiness law a row with no
number carries nothing — and complains once, on stderr, because nothing in the
request path can produce one of these any more.

## `refused` was stamped where nothing was refused

The word was written wherever any gate anywhere said no, which made it the
commonest word in the log and the least informative: a race that put **three
arms on the wire** and was then declined a fourth carried
`hedged arms=3 refused=budget`. It now names what the question was **left
without** — no answer in, no rescue running, and the question really did end
with the one arm it started with. What is spent is unchanged.

## Nobody knew who cancelled the turn

`context.CancelFunc` carries no account of itself, so every door — the stop key,
another window taking the conversation over, a tab closing, a hosted session
retired, the session quitting — produced the same eight characters.
`internal/session/stopcause.go` gives each one a cause; `loop.go` reads it back
at one boundary and writes down the ending; the person is given **one sentence**
in their own words, except for their own stop, which is the one door owed
neither; the call row reads `context canceled (turn ended: taken over)`; and a
request cut out from under a **live** turn is re-asked through the existing
ladder rather than reported.

The 109-second arm was cancelled through `Agent.Interrupt` from the takeover
doorstep — `takeover.go`'s `drainTakeover` on the 250 ms presence tick, which
deliberately answers a request mid-reply, into `tui3`'s `closeFront` →
`endAgent` → `Interrupt`. That road now says `taken over` on the row and in the
journal, and puts one sentence in front of the person.

## The rest of the fixed figures

Audited: `lane.VisiblePatience`, `control.fallbackCeiling`, streamguard's three
silence bounds and its wall, `routeRaceWindow`, `DefaultBudget`, `stopGrace`,
every `WithTimeout` on the request path. One moved — `internal/reflex` read
`VisiblePatience` directly where it declares `RoleRecall`, and reads that role's
own ceiling now. The rest bound side errands whose failure degrades rather than
silences, or bound a still wire, and each is named in the change entry.

https://claude.ai/code/session_01PEEpmnn8oq3cyTtDYcHNdd


santoshkumarradha:
Performance/context integration note from #741 and the now-merged #755:

Our frozen benchmark found a primary routing 404 followed by an accepted rescue stream that hit `CutOverrun`. Previously the race reported the stale 404, bypassing the session’s existing cut recovery. #755 now preserves the typed rescue cut; it deliberately adds no provider-level retry after exposed output.

Please retain that all-arms-failed selection when integrating this branch’s `hedge.go` reporting changes, and keep `TestALaterAcceptedStreamCutOutranksThePrimaryRoutingRefusal` in the combined validation. The new generation-cancellation retry in `completeWithRetryReasoning` also overlaps the same cleanup boundary: cover a typed rescue cut alongside an internal cancellation and assert partial text/reasoning/tool fragments are cleared once, existing cut/attempt budgets remain bounded, and parent cancellation never starts another request. This is especially relevant when composing with #741’s simpler completion loop; neither change should resurrect a completion reader or mandatory progress-note call.

The current scored cohort is frozen and excludes this PR; improvements from these changes must be measured in a new cohort rather than attributed to the running results.

PR #762, opened 2026-09-10
config: AFORGE_QUESTION_DEMO is operator plumbing, so internal/config is green on a clean dev again

### LINKED ISSUE #754: build: an unregistered AFORGE_QUESTION_DEMO leaves internal/config red on a clean dev, so every pull request that touches it opens with a failure it did not cause

AbirAbbas:
## What was wrong

`TestRegistryCoversEveryUserFacingEnvironmentPin` fails on a clean checkout of `dev`,
with no changes of any kind:

```
--- FAIL: TestRegistryCoversEveryUserFacingEnvironmentPin (0.03s)
    settings_test.go:111: AFORGE_QUESTION_DEMO (…/internal/tui3/questiondemo.go) is neither a settings row nor operator plumbing — register it in settings.go
```

The pin arrived with the questions page (#733) and was registered nowhere. It is not in
`.github/known-red.txt`, so nothing skips it: every pull request that touches
`internal/config` — or whose `touched packages` gate pulls that suite in — opens with a
red it did not cause. #757 is one of those right now.

## What is now true

`AFORGE_QUESTION_DEMO` is on `OperatorEnvPins`, beside `AFORGE_WIRE_LOG`, with the reason
written out. Plumbing rather than a settings row on the fixture's own account of itself
(`internal/tui3/questiondemo.go`): there is no key, no slash command and no row that
mentions it, it writes nothing anywhere, and it goes the day the `ask` tool and the cards
give the page real questions to draw. Those are the terms the wire log is listed under —
a developer's door, not a preference. A row would be offering to persist a fixture over
somebody's conversation every launch.

Nothing else changed: one entry and its comment, no behaviour, no new pin, and nothing
added to the known-red ledger.

## Validation contract

- The registry law refuses an unregistered `AFORGE_*` pin and accepts a registered one.
  `go test ./internal/config/ -run TestRegistryCoversEveryUserFacingEnvironmentPin
  -count=1` **fails on `origin/dev`** with the line quoted above and **passes here**.
- The pin stays out of the editable sheet: it is on the allowlist, so it is listed
  read-only in the environment footer and never becomes a row. `internal/tui3`'s full
  suite is green, which is where the sheet is drawn.

## Gates

Run with `GOFLAGS=-buildvcs=false`.

| Gate | Result |
| --- | --- |
| `go build ./... && go vet ./...` | clean, no output |
| `gofmt -l internal/config/settings.go` | clean, no output |
| `make test-quick` | pass |
| `make test-laws` | pass |
| `make test PKGS='./internal/config ./internal/tui3' TEST_FLAGS='-count=1 -timeout 15m'` | `internal/config` ok 81.4s, `internal/tui3` ok 418.2s |
| `make build` | pass |

The `SIZE-BUDGET` ratchet is still red on `dev` (#724) and is untouched by this.

Closes #754


AbirAbbas:

PR #763, opened 2026-09-10
Reduce old observed tool results without requiring file changes

santoshkumarradha:
A long research turn can retain hundreds of kilobytes of shell and read results because the current fold requires a later file change and refuses all savings unless they reach its entire lower target. Reduce complete tool-result batches once they have been presented to the model and leave the existing recent window, regardless of the kind of task.

The change removes file-write consumption state, keeps the original ask and call arguments, and reuses head/tail views with recoverable originals. Useful partial reductions must satisfy the existing cache-reclaim share; already-reduced, recent, unseen and unfileable results remain unchanged. Speculative tool execution and generation settings are unchanged. The manual and PERF describe the changed retention law and its limits.

Validation: real Submit/replay and partial-reclamation regressions failed before the change and pass afterward, including repeated tool IDs, complete batches, artifact recovery and non-text content. The draft includes exactly dev888b57e3 (#760), preserving #755 typed rescue cuts and #760 retry/cancel behavior. Frozen head1925ed7c1 passed fresh uncached session (288.732s), manual (1.718s), provider (57.744s) and controller (0.456s) suites, make test-quick and make build. GitHub light gate, touched packages and aggregate check are green. No later dev revision was included.

This is an isolated experiment. **Do not merge before a controlled benchmark demonstrates an improvement against the measured best version.** No paid benchmark has been launched and no speed, cost or quality improvement is claimed from these regression tests.


santoshkumarradha:
Related context/recovery finding: draft #808 preserves `read` during a task's finishing turn. In the frozen random holdout, the actual final request contained archive stubs for all three reads of the target source, while the finishing tool set had removed `read`; the attempted reread was rejected. The original user request remained intact, so this is evidence retrieval being blocked, not loss of the user's instructions.

Please preserve the archive-retrieval contract when adapting context reduction or progress handling: replacing a result with a file reference depends on the worker still being able to retrieve that file. #808 adds a combined deterministic regression and a minimal capability fix, with no task-specific rules or changed progress budgets. Spark validation is queued; no performance benefit has yet been measured and neither draft should be merged on that claim.

santoshkumarradha:
Closed per owner request.

PR #766, opened 2026-09-10
A conversation that moves mid-thought asks its question again where it lands

santoshkumarradha:
Follow-up to #760.

## What was believed

takeover.go's second law says a request to move a conversation is answered at once, mid-reply or not, because "the window that asked resumes it from the checkpoint with the partial reply already in the journal". That is true of a turn that had **said** something. It is false of a turn that had only been **thinking** — reasoning never enters the partial, so there is nothing to keep. #760's measured case was exactly that one: 108 seconds of streamed reasoning, a takeover, and the conversation arriving in the new window as a question with nothing under it.

#760 made that ending visible — a failed call naming the door in the journal, one sentence to the person — but the sentence ended `— ask again to pick it up`, which hands the repair back to the person who typed the question once already.

## What is true now

The window that takes the conversation over asks the question again itself, through the ordinary turn door, and the reply streams where the person is sitting. They type nothing.

**The decision is the journal's shape, not a flag.** Two windows are two processes and the only thing they share is the file, so the file says it: the last message is the person's own words, and the only thing after it is this machine stopping the turn that was answering them (`readJournal` reads it, `replayedSession.stopped` carries it, `Agent.stoppedTurn` holds it and spends it once). That shape cannot be produced by anything else —

- an assistant message, a tool call or a result after the question means an answer of some size arrived, so the tail is not the question any more and today's behaviour stands (tasks land paused and resume from their checkpoint; no side effect runs twice);
- a provider's refusal is a thing that happened on the way to an answer and leaves the shape alone;
- a person's own stop writes no row at all, so a turn somebody stopped themselves can never look like this.

`resumesUnanswered` is a switch with one name on it. A takeover is the one door where somebody is demonstrably sitting in front of the conversation a moment later; every other door leads somewhere else, so those keep their #760 sentence.

**No second copy of the person's words.** `userMessage.resumed` marks a message that is already in the transcript and in the file; `startTurnLocked` skips the record and does everything else exactly as it does for a typed message — steer, presence, title, checkpoint, the retry ladder, the spend rail. The stopped attempt's reasoning and partial are simply gone, because nothing kept them.

**One line, from one constant.** `session.ResumedWord` — `the reply stopped when this conversation moved — asking again` — in `cutShortNotice`'s register.

`journalError` gains `door`, so the reader deciding this never parses prose.

## The mirror hazard, closed with it

A takeover request is good for ten minutes and the beat answers one in a quarter of a second — so a request still on the disk when a **later** turn opens is one nobody is waiting for. Answering it took a reply away from somebody who typed their question after it was written.

- `drainTakeover` no longer lets a request older than the running turn's start end that turn. It is **held**, not dropped: the request stays live, so a window that really is waiting is answered on the first beat after the turn ends.
- `takeoverTick` withdraws its own request when the claim succeeds. A lock can free for its own reasons — the other window quit, or moved to another conversation — and the session opened on it would otherwise find this window's own question on its beat and let go of a conversation nobody asked it to.

## The engine road is unchanged, and deliberately

`movedAway` is `takeOver` minus the interrupt: on the engine road the turn keeps running in the engine for the window that took it, nothing pauses and nothing is lost, so there is nothing to resume. Verified, untouched.

## Manual

`chat/models-and-cost.md`, `chat/screen.md` and `chat/home.md` said the reply was lost and to ask again; all three now say it is asked again for you, and say the two cases where it is not. Three new probes in `internal/manual/chat_test.go`.

## Tests

- `internal/session/resume_test.go` — end to end through two agents on one file: a thinking turn taken over, reopened, asked again once, with the person's words in the journal exactly once; a person's own stop is never asked again; a turn that had already written part of its reply is not; a table of ten whole journals over the shape itself; the one-door law and the sentence's register.
- `internal/session/takeover_test.go` — a request older than the running turn is held and answered after it.
- `internal/tui3/takeover_test.go` — the arriving window adopts the resumed turn, says the line once, and leaves no question behind in the folder.

```
go build ./...                                          ok
go vet ./...                                            ok
make test-laws                                          ok (18 packages)
make test-quick                                         ok
make test PKGS=./internal/session TEST_FLAGS=-count=1   ok  277.9s
make test PKGS='./internal/tui3 ./internal/manual ./internal/e2e' TEST_FLAGS=-count=1
                                                        ok  445.2s / 1.7s / 1.2s
```

Nothing added to `.github/known-red.txt`.

https://claude.ai/code/session_01PEEpmnn8oq3cyTtDYcHNdd


santoshkumarradha:
Compatibility finding with #763 (also possible with today's turn fold), checked at `3d825c2e`:

`readJournal` reopens the unanswered-question shape for every `user && !Note && !Steer` message ([sessionfile.go:1683](https://github.com/Agent-Field/aforge-v2/blob/3d825c2e2bf7f9f9ae557a0daa67403aad290153/internal/session/sessionfile.go#L1683)). A turn fold appends `[folded …]` as a user message without registering it as a note, then `appendCompaction` writes that final window message as an ordinary message ([turnfold.go:188–203 in #763](https://github.com/Agent-Field/aforge-v2/blob/1925ed7c12e597e9d82fc47c618afe1c6823bc7f/internal/session/turnfold.go#L188)). Consequently, clearing `asked` at the compaction marker is undone by the fold message below it.

Concrete sequence: completed tool calls/results → fold/window ending in `[folded …]` → next response is thinking-only → takeover. Replay marks that synthetic tail as an unanswered question, and `lastPersonMessageLocked` accepts it because it checks only user role/nonempty text ([resume.go:149](https://github.com/Agent-Field/aforge-v2/blob/3d825c2e2bf7f9f9ae557a0daa67403aad290153/internal/session/resume.go#L149)). This starts a new paid turn on the fold marker despite the stated rule that work which already called tools must not be automatically asked again.

Please exclude the existing `isCompactionNote` protocol markers from both question recognition and the final resume guard (including legacy unmarked journals), or provide equivalent durable provenance handling. Add a regression that writes a real fold window, appends a takeover stop, reopens, and asserts `ResumeStoppedTurn` returns false with zero model/tool executions and unchanged recoverable result pointers. No changes to #763's running benchmark are needed for this review.

Issue #393, opened 2026-09-02
host: a background reading whose trip panics keeps its latch forever, so the page freezes with nothing said


santoshkumarradha:
## Problem

The surface over a connection (`--host`, `--at`, and the local socket door) keeps four readings of the far machine warm on goroutines of its own: the places behind home, what has been spent, what is remembered, and the standing items. Each takes a latch before its round trip and releases it inside the trip, after the answer comes back.

A trip that panics never releases the latch. Every later reading sees a fetch still out and declines to start another, so the page freezes on whatever it held — on the first reading after launch, nothing — for the life of the process. Nothing is said: the guard records the fault in the log and the goroutine ends quietly. By the emptiness law an empty home, spend or memory page is indistinguishable from a machine with nothing on it, so a person reads a broken page as a quiet one.

The witness is `go test ./cmd/aforge/`, which logs six recovered nil dereferences per run:

```
fault scope="chatv3/host-ledger" panic=runtime error: invalid memory address or nil pointer dereference
fault scope="chatv3/host-memory" panic=…
fault scope="chatv3/host-world-prime" panic=…
```

Two door tests build the surface with no client at all. The wire doors are closures made from the client, and a closure made from a nil is a perfectly good function right up until it is called, on the goroutine. The tests pass because the guard swallows the panic — which is the second half of the report: a suite that logs six faults on a green run is a suite in which the seventh fault is invisible.

## Can a person reach it today

Not by the nil. Every product door gets its client from a handshake, returns on error, and reads the agent and the welcome synchronously before building the surface, so a nil would stop the launch before any duty exists. A dropped link does not nil the client: it redials itself, and a call in the gap or after a close is an ordinary error, which the readings already handle by keeping what they held.

The frozen latch is real and general: it needs only a panic inside a trip, from any cause, and no reading survives one. It has no known trigger in the product today. Severity: low now, and a shape that #338's family is about — a background duty failing with nobody told.

## The law

A background reading never runs against a connection that cannot serve it: it is either given a live one or it does not start. A reading that dies lets go of its latch, so the next beat asks again exactly as after an ordinary error. And the surface hears about it once.

## Fix shape

One seam, no nil check per site:

1. The four readings are armed through one door that knows whether there is a connection. Armed with none, a reading refuses every trip; the page reads what it holds and no goroutine is sent down a nil.
[...]

PR #770, opened 2026-09-10
Experiment: let useful tool calls proceed without mandatory progress notes

santoshkumarradha:
A useful tool-only turn currently gets two narration demands, then its tools are held and the turn can stop solely for omitting a progress note. This experiment removes that mandatory narration path and its false assertion that provider reasoning is not retained. Optional updates remain visible; genuine repetition, invalid-call correction, completion, delegation, cancellation and provider retry policies remain.

The real Submit regression fails on pinned dev c9aff5b757 after 12 of 18 useful calls and passes with all 18 after this change. Historical task endings keep their labels and stopped classification. The single-tenant process-rule registry is removed rather than left behind a flag.

Validation on the current PR head 033906887:

- Full uncached session: passed, 290.253s; full manual: passed, 1.748s.
- `make test-quick`: passed, including packed manual, build, vet, formatting and laws across 63 files in 18 packages.
- `make build`: passed; local binary stamped 033906887, dirty=false.
- Focused genuine repetition/error/invalid controls and optional-note retention passed.

The previous CI failure was reproduced on its exact integration commit ad337144c, which combined faf9c5ac7 with newer dev 2309b20fe. The combined task headings exceeded the manual section-list budget. Commit 033906887 shortens one heading, preserving retrieval vocabulary and passing the exact integration's section-bound and search checks. No runtime code, test expectations or limits changed in that follow-up; standalone faf9c5ac7 also passed the focused section test.

Benchmark #393 is running 14 matched pairs / 28 cells on Spark. Its immutable pins are control c9aff5b757d14fbeacb02d01d9e2c5e10513a2d3 and candidate faf9c5ac79c33440cd6b711d2ee8cde0f3527843, with the same DeepSeek V4 Flash model, provider-default generation/reasoning settings and matched seeds. The documentation-only 033906887 commit was made after launch; it has not replaced any running binary or frozen artifact. Results must distinguish whether the removed enforcement actually activated in a control run.

Draft experiment only. No performance or quality improvement is claimed, and this PR must not merge before the matched benchmark supports it.


santoshkumarradha:
Related context/recovery finding: draft #808 preserves `read` during a task's finishing turn. In the frozen random holdout, the actual final request contained archive stubs for all three reads of the target source, while the finishing tool set had removed `read`; the attempted reread was rejected. The original user request remained intact, so this is evidence retrieval being blocked, not loss of the user's instructions.

Please preserve the archive-retrieval contract when adapting context reduction or progress handling: replacing a result with a file reference depends on the worker still being able to retrieve that file. #808 adds a combined deterministic regression and a minimal capability fix, with no task-specific rules or changed progress budgets. Spark validation is queued; no performance benefit has yet been measured and neither draft should be merged on that claim.

PR #773, opened 2026-09-10
home: the consent row is the gate's own sentence, and a one-off that fired is in `since you left`

santoshkumarradha:
Two defects the tmux TUI suite read off a real screen and logged in-test (#756).
Both are on the switcher's side of home; internal/session is untouched.

## 1. `consentws wants to needs your ok to run bash`

internal/session's `consent.go` writes ONE line for a permission gate — `needs
your ok to run ` and the tool's name — and its own comment says what that line
is for: **"the one line another window may answer this from"**. It is already a
whole predicate about the conversation whose name home draws beside it.

`switcherConversationNote` prefixed `wants to ` onto it. That prefix dates from
when the engine handed over a bare action, and what a person actually reads on
home today is the sentence above.

**The row is the engine's sentence now, and this side adds nothing to it.** Not a
condition on the shape of the line — one sentence, written in one place,
repeated exactly. A lane that ever writes a bare action owes the whole sentence
instead, which is what `consent.go` already promises.

The unit test that covered this — `TestTheSwitcherRepeatsAConsentQuestionInItsOwnWords`
— fed a sentence nobody writes (`send the report on your behalf?`), a bare
action that the prefix made a clause of. That is exactly why both halves were
green while the screen was wrong. It feeds the engine's real line now and
asserts the note **is** it.

## 2. `since you left` was silent about the commonest firing there is

internal/standing's `tick.go` stamps `LastFired` and then retires a one-off **in
the same pass**. `app.standItems` drops every retired item, correctly — a thing
that is over is not keeping an eye on anything — and `addLedger` walked the
bands. So `remind me in 1 minute to drink water`, fired with the terminal shut,
left the screen a person came back to saying nothing had happened.

The same reading carries the retired firings out now (`standItems` returns both
halves from one walk of the store — the second half is a test on the very
documents the first was throwing away), and the ledger walks **what stands and
what went**. The line is `standing.LastLookLine`'s, the way a live watch's is:
`fired 3 minutes ago — it told you`.

**The dead item is still on no list.** The band is what is true now; the block is
what happened. Both are asserted.

**And the first look still marks nothing.** The retired half is bounded by the
look stamp where the walk is taken (`standFiredSince`), not only where the block
is built — otherwise a machine nobody has closed home on hands the ledger every
reminder it ever fired, for the block to throw away one line later.

## The two logged findings are assertions

- `the_firing_reaches_the_person` had a three-armed `switch` that logged
  `FINDING:` and passed. It demands the `since you left` block and the word
  `fired ` now.
- `answer_from_home_across_two_windows` demands the gate's sentence whole and
  refuses the prefix, through a new `consentRowLine` needle in
  `tuiwords_test.go` owned by `internal/session` — so the day that sentence
  moves, the untagged gate names it in under a second.

## The manual

`internal/manual/chat/home.md` quoted `wants to send on your behalf` as the
consent row in four places, and said the ledger holds three kinds of line with
no mention of the one that is over by the time you read it. Both are true again;
`keeping-an-eye.md`'s "a reminder fired but nothing showed up" list says home's
own block names the firing too.

## Verified

Here: `./internal/tui3/` (432s), `./internal/session/` (305s), `./internal/e2e/`
(untagged word gate), `./internal/manual/...`, `./internal/standing/`,
`make test-laws`. The new ledger test was confirmed red against the unfixed
walk before the fix went in.

On the Spark: build, those three packages, `make build`, the two tagged TUI
subtests against a real model, and `make test-laws` — results in a comment
below.


santoshkumarradha:
## Verified on the Spark, against a real model

Branch tip `129100541`, in a detached worktree off `origin/fix/home-switcher-findings`.

| | |
| --- | --- |
| `go build ./...` | ok |
| `go test ./internal/tui3/` | **ok, 420.2s** |
| `go test ./internal/session/` | ok, 195.1s |
| `go test ./internal/e2e/` (untagged word gate) | ok |
| `make build` | ok |
| `make test-laws` | ok |

### The two subtests, on a real terminal

```
--- PASS: TestTUIE2E/answer_from_home_across_two_windows (21.23s)
--- PASS: TestTUIE2E/the_firing_reaches_the_person (224.36s)
```

**Defect 1, the row as it reads now.** Window B's home, over window A's stopped
conversation:

```
the chips home offered: › needs your ok to run bash
```

The gate's own sentence, whole, with nothing in front of it. Before this change the
same screen read `consentws wants to needs your ok to run bash`.

**Defect 2, the `since you left` block after a one-off fires with no window open.**
`remind me in 1 minute to drink water` stood, the terminal was shut, `aforge tick`
fired it, and the window that came back drew:

```
fired just now · it was the time you asked for — it told you            standing
```

That line is `standing.LastLookLine`'s, the same sentence a live watch's firing gets.
The subtest's old three-armed `switch` logged `FINDING:` here and passed either way;
it asserts now.

### One note on the first run

The first pass of `the_firing_reaches_the_person` failed on `tui_e2e_test.go:699` —
`the status line never grew a 'keeping an eye on N' segment` — with **both** of the
assertions this change is about passing in the same run. That check reads the standing
seam directly through `app.keepingCount`, which never goes through `standItems`,
`readStandBands` or `switcher.go`, so no code on this branch can reach it; it is the
loaded-box papercut #756 named when it widened that wait to a minute. The Spark was
running three other lanes at the time. On a re-run, on the same commit:

```
tui_e2e_test.go:701: the `keeping an eye on N` segment arrived 0s after the item stood
```

An `internal/tui3` run in the same first pass reported `signal: terminated` at 418s
with no `--- FAIL` line anywhere — the binary was killed from outside. Re-run clean at
420.2s, and CI's touched-packages job runs the same package green.

PR #775, opened 2026-09-10
docs: run final and full acceptance tests only on Spark

santoshkumarradha:
Final end-to-end and full acceptance suites must run on Spark, so concurrent sessions do not load the owner’s laptop. Record this standing rule in the shared CLAUDE.md / AGENTS.md instructions, including full affected-package suites, exact-revision remote evidence, and no local fallback when Spark is unavailable.

The same policy is already present in the owner’s global Claude and Codex instructions. This change makes the repository instructions consistent. Documentation only; checked the diff and AGENTS.md symlink. No test suite was run on the laptop.


santoshkumarradha:
Closing as superseded by the merged test-policy wording (post-#1015).

This PR proposed a hard rule: full / end-to-end acceptance suites run NEVER ON THE LAPTOP, no local fallback when Spark is down. The text now merged on dev deliberately keeps a softer position — `make check` is "the full-tree … ritual for Spark, staging, or an intentional full laptop run" — i.e. an intentional laptop run remains permitted. That is the accepted position, so the stricter NEVER is intentionally not carried.

Nothing else in this docs-only change is missing from current dev. Closing without merge; reopen if the laptop-fallback should be removed after all.

PR #777, opened 2026-09-10
bench: add an independent random GitHub holdout

santoshkumarradha:
The five repeatedly used GitHub issues are now a development set. Add a preregistered random holdout from cached SWE-rebench-v2 fixtures, excluding all five existing repositories, to test whether the measured harness changes generalize.

Freeze a random ranking, five distinct primary repositories and ordered reserves before solution outcomes. The Spark-only calibration script requires an assertion-failing original and passing upstream solution under the same offline tests; records infrastructure exclusions; never replaces scored failures or rewrites acceptance. Reference solutions remain outside agent workspaces.

Planned scoring: the existing Aforge baseline and optional-progress candidate, Pi 0.84.2, and mini-SWE 2.4.6; two seeds; the same exact DeepSeek Flash model and provider-default generation/reasoning settings. Forty scored trials only after calibration and actual-client preflight. This is a cached-fixture convenience sample, not all GitHub. Non-coding controls remain separate.

Validation: Python syntax parsed, fixture hashes recorded, diff checked. Calibration queued on Spark as `20260910-034120-000396`; no paid trials or local test suites launched. No product behavior change and no Pareto claim.

Shared grading adapter now supports nested test directories and requires calibrated reference test identities/outcomes. Offline regression checks are queued on Spark as `20260910-043358-000397`; scoring still has not launched.


santoshkumarradha:
The first four-arm Hypothesis result exposed a benchmark specification defect: all four arms timed out and fail collection on `emails(domains=...)`, but the supplied issue3567 explicitly proposes `allow_special_use` and never requests the `domains` argument. The retained PR3572 also incorporates the separate PR3582 ("Add domains() strategy as optional arg for the emails() strategy"); `linked-issues.json` includes only3567.

Base-fails/gold-passes calibration checked mechanics but did not establish that the public task specification covers the upstream test scope. This result must not be presented as a clean solve-quality comparison. Both seeds are flagged; all frozen inputs, raw failures, timeouts, patches, and costs remain intact. No test weakening, scored retries, or post-result replacement.

Before another campaign, the general preparation needs a specification-to-acceptance scope audit, including PRs that combine changes beyond a single linked issue. I am checking the other fixtures for the same issue. This is a benchmark context defect, not evidence that Aforge lost context it was given. The current campaign remains diagnostic; no runtime merge is justified by this pair.

PR #776, opened 2026-09-10
tasks: a refused accept says so, and a re-landed your call is answerable again

santoshkumarradha:
Closes #767.

The owner hit this live on 2026-09-09. A divided task landed `your call · nobody could check it` under `task.settle = auto`. The model called `tasks 1 resolve accept`. The tool replied *"Its branch has come home and anything waiting on it can start"* — but the branch had not come home: the merge was refused because the person's own **untracked** copies of the very files the task wrote were sitting in their folder. The model then told them "Task 1 is done and accepted", and the screen went bright and handle-less: a rail row reading `your call · conflicts with your branch:` with a dangling colon, a room saying `this task has finished — say it to main` with no answers, a conversation card frozen in the first landing's shape, and children 2–6 wearing the false receipt `you looked at this yourself and took it as done`.

Five of the issue's six defects are fixed at their causes. Defect 6 (the checker's window) is reported, not fixed.

### 1 · the tool reply says what actually happened

`internal/session/tools_tasks.go` — the model's verb goes through `Agent.resolveUnverifiedBy(…, TaskAskOwnerModel)` and the reply is built by `resolvedReply`, which READS THE NODE BACK after the door returns and answers from its state and its merge. A refused accept now replies `task 1 is still your call — your folder already has files the task wrote: … · its branch task/… was kept. Nothing merged and nothing failed`. Refute and reaudit answer the same way.

### 2 · a re-landed node is an open question again

Three independent causes, all of them found and closed:

- **Nothing ever published a landing question.** `Agent.landingQuestion` existed, but only `WatchQuestions`' attach-time replay ever put one in front of a surface. `internal/session/task_landing_question.go` (new) publishes on every landing and withdraws when the node settles, and `emitTaskUpdate` calls it.
- **`questionDrawnHere` did not claim `QuestionLanding`**, so the block would not have drawn it anyway.
- **The surface threw the re-settle away.** `app.taskUpdate`'s de-dup guard drops a same-state notice unless it carries recognised news, and a new merge word, a new file list and a new road were none of them news. `taskReasks` in `internal/tui3/task.go` is that recognition.

So the card, the room and the chip all come from **one object** now. `internal/tui3/tasksettle.go` shrank from 1047 lines to ~230: its own rows, its own five letters, its own receipts and its three copies of the routing are gone, and what is left is the card lookup, the redraw, and the two things nothing else provides (`[s] tell it`, and the legend's ranked-prefix hint built from the question's own answers).

### 3 · the rail sheds the colon with the list

[...]

Issue #767, opened 2026-09-10
tasks: an accept whose merge is refused replies "its branch has come home", and the re-landed your-call card has no answers anywhere


santoshkumarradha:
## What happened

2026-09-09, conversation `91c8fc31e5845cc1` ("ai devs"), engine build 550cdb72c, `task.settle=auto`. A divided task (1, five children 2–6) landed `your call · nobody could check it` (the checker ran ~29s per call and was abandoned twice). Under auto the model called `tasks 1 resolve accept`. The tool replied:

> accepted task 1 as done on your reading of it, with nobody else's check behind it. Its branch has come home and anything waiting on it can start.

The branch had NOT come home. `tasks.json` for the node: `state: unverified, merge: conflicted`, report `conflicts with your branch: leads-contact-sheet.md, research/method.md, research/sections/*.md …`. The person's `work/` tree is on `main` with `?? leads-contact-sheet.md` and `?? research/` UNTRACKED — copies the model itself had written into the person's folder during its turn — so the landing was refused on the "your own uncommitted work in …" road (groundcarry.go), not a content conflict. The model, told the branch was home, then told the person "Task 1 is done and accepted".

What the person then saw:
- rail: `? Emails and socials for… · your call · conflicts with your branch:` — the file list shed, the colon left dangling;
- the task's room: header `? your call · 25m 27s · 41 tool calls`, foot `this task has finished — say it to main`, NO answers row anywhere; typing `done good` into the room composer did nothing;
- the conversation: no card with `[a] resolve it · [n] drop it · [s] tell it` for the re-landed node — the landing card was created under auto (no chips, "aforge is deciding") and the re-settle to a person-owned conflict never gave it any;
- children 2–6 accepted by the model under auto carry the receipt `you looked at this yourself and took it as done`, which is false: the model did.

Status line and presence both said `your call`. A bright terminal state with no handle.

## Defects, each its own fix

1. `internal/session/tools_tasks.go:603` — the accept reply is unconditional. It must read the node's state after `ResolveUnverified` returns and say what happened: came home (done) / refused with the reason sentence and the node still `your call`. The model reasons from this sentence.
2. `internal/tui3/taskdone.go` — a re-landed node (a fresh notice for an id that already has a card) must get a card whose `asks`/chips follow the NEW notice's `Decider` (person → chips), not the frozen first landing's. The room (`roomSettleRows`) must draw the answers row for any person-owned your-call node it shows.
[...]

PR #778, opened 2026-09-10
tui3: the column and the box say the same thing about a landing as the card does

santoshkumarradha:
Follow-up to #776, which merged while the tmux run that found these was still going. Two screens went on drawing the shape #767 was filed about.

**The composer's placeholder.** #776 stopped the room's *foot* saying `this task has finished — say it to main` over a standing question, and the box went on saying it in its placeholder — which is worse, because the placeholder is the row a person is looking at when they decide where their words go. `[s] tell it` is one of the three answers on the block above that box, and what it does is point that box at that task. Measured on a real screen in the new `-tags e2e` subtest:

```
 To: Emails and socials for the fifteen leads
 › this task has finished — say it to main
```

**The column's reason.** `app.taskStatus` built its `TaskFacts` without `Shifted` and `GroundHeld`, so the rail and the roster read the third road over all three. On the same screen the header said `your folder already has files the task wrote: leads-contact-sheet.md, research/method.md` and the column beside it said `your call · conflicts with your branch` — about a folder with no branch of theirs in it. All three of those your-call questions name files and take the same two answers; the sentence between them is the only thing that says which, and it is the half a person can act on.

Two unit tests, both of which fail without the change:

- `TestTheBoxOfANodeThatNeedsALookStillPointsAtIt` — and a landing nobody is waiting on keeps the sentence.
- `TestTheColumnNamesTheRoadTheCardNames` — and an ordinary branch clash is still a branch clash.

`TestRefusedAcceptE2E/the_same_answers_stand_inside_the_nodes_room` is what caught the first one; it goes green with this.


santoshkumarradha:
Verified on the Spark at `bbab8c7b4`:

- `go test ./internal/tui3/ ./internal/e2e/ ./internal/manual/...` — **units=0** (tui3 419.9s)
- `go test -tags e2e -run "TestRefusedAcceptE2E|TestTaskStatesE2E"` — **e2e=0**, 591.7s. `TestRefusedAcceptE2E` both subtests PASS (the room one is the one that failed before this change). `TestTaskStatesE2E` all PASS bar two environmental SKIPs it always takes — `a_branch_that_clashes_with_yours` (no task branch was cut in four minutes, so there was nothing to clash with) and `the_auto_settle_floor_hands_it_back`.
- `make test-laws` — **laws=0**, all 18 packages.

All three PR checks green.

PR #789, opened 2026-09-10
questions: every answer has a pointer the arrows walk and enter takes, and the task page shows the question it needs you for

santoshkumarradha:
## What was wrong (owner's screens, 2026-09-10)

1. `enter` on a needs-you row (home) opened the task's record page — report, files, transcript — with **no question and no `a`/`n`/`s`**. The landing question was on the block above the conversation's box all along (`5 questions · alt+a`), but the record page is a place: the block is neither drawn nor keyed under it, and `alt+a` from there did nothing.
2. "none of my arrow keys make it obvious how to have question" — only the stop card had a cursor. Every other question: digits, plus `enter` on the model's pick when there was one. `↑`/`↓` did nothing.
3. On a card the consequence beside the label and the note under it were both dim: one grey paragraph with no hierarchy.

## What is true now

- **A pointer on every question with answers** (`questionPointerStart`): `▸` on a card, the selected band on a line. `↑↓`, `←→`, `tab`/`shift+tab` walk it; `enter` takes it; a digit still answers at once. Starts on the asker's pick, else the first answer; on a confirmation on the answer that loses nothing (stop.go's law kept). Under a place (home's move card) `↑↓` stay the place's.
- **The asker's pick says `suggested`** in its consequence column on every card; `▸` means one thing now. Key rows: `[enter] take it` on every question with answers, `[↑↓] choose` (card) / `[←→] choose` (line), both given up early.
- **Hierarchy on a card:** label in the question hue, note in the prose ink, consequence dim.
- **The task record page draws the landing question** (head, reason, `[a] accept · [n] not right · [s] tell it`) above its foot for a node of THIS conversation, walks `←→`, and routes `a`/`n`/`s`/`enter` through `questionOptionKey`; `s tell it` puts the sheet down before opening the room. Foot: `←→ choose · enter take it · …`. The chip (`alt+a`) closes the tasks place too.
- Manual: card/line paragraphs, the key table, a new section *The task page shows no accept or not right keys — where the landing question is when I press enter on home*; `take the pick` respelled `take it` everywhere (e2e word table updated).

## Tests

`TestArrowsWalkThePointerOnACardAndEnterTakesIt`, `TestArrowsWalkThePointerOnALineAndEnterTakesIt`, `TestTheTaskRecordPageDrawsTheLandingQuestionAndTakesItsKeys`; existing pick/checklist tests adjusted to the new words. Full `internal/tui3` on the Spark — results in a comment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #791, opened 2026-09-10
questions: the room's answers walk, open and click, and read in three tiers

santoshkumarradha:
**Stacks on #789** (`fix/question-pointer`) and will be rebased onto `dev` once that merges.

The owner opened a question out on 2026-09-10 and reported two things about the page (`internal/tui3/questionroom.go`, the page `o` opens): "once its opened the layout org and everything seems really bad and no arrow or click or bugs please clean all those", and "make sure there is some textual hierarchy in design in options like the same line and next line in options look same and a bit weird".

Both were true. The foot said `[←→] choose` and neither arrow moved anything on the page; no key opened or folded a section but `o`; nothing on it answered to a mouse at all; and an open answer drew its body, its consequence, the asker's reason and what would change its mind as four identical grey rows at one indent — one of which read `would switch if If you are targeting…`, because the model's sentence was capitalised and the prefix is a clause.

### The pointer

- `↑`/`↓` walk the answers. The foot names **that** pair now (`[↑↓] choose`): this form's answers stand in a column of sections, and the pair it was naming walked nothing.
- `→` opens the answer under the pointer, `←` folds it. They state what they want rather than toggling, so holding one down is safe. Where the question carries a shape of its own — a dial, a hole with choices — the side arrows are still that shape's.
- `enter` takes the answer the pointer is on, which is #789's law arriving on the page. It costs nothing on a fresh room: the room opens standing on the asker's pick, so `enter` straight away still takes the recommendation.
- **A click on an answer's row** moves onto it and opens it; a **second click on that same row** is `enter`. One press to read, one to decide — a page where the first click answered would answer with evidence somebody had not read. Only an answer's own row answers to the mouse, and it is the only row that lights under the pointer (hover.go's law); its body, its diagrams and your own notes are prose.

### The hierarchy

Three tiers, and every row says which it is:

- the **label** — bold in the question hue, exactly as the card already spells an answer's word; it was the body's own ink, so a heading and its paragraph were one column of identical grey
- the **body** — the prose ink
- the **asides** — all dim, each behind its own word: `then ·` what taking it would leave true, `why this one ·` the asker's reason, `would switch if` what would change its mind
- one blank row closes an open answer, so the next label is not merely a different indent

`would switch if If you are targeting…` is gone: a leading `if` the reader is about to read twice is dropped and the opening capital lowered, with a word whose second letter is also a capital (`SQLite`, `HRV`) left alone.

### The evidence

An answer's own blocks stay under that answer at its body's indent, with a blank row above each. The question's own `Attach` is now **one dim-titled section** — `what it showed you` — above the answers, indented under its title, rather than a run of unowned pictures starting at the gutter after them.

### And one safety fix found on the way

Two tests in this package were already red on `fix/question-pointer`, and one of them was red for a reason: `enter` now takes the answer the pointer is on, and the pointer started on the first answer of a **consent gate** — so on the gate for `rm -rf *`, `enter` meant `allow once`. Measured, not reasoned.

`questionPointerStart` now opens on the answer that loses nothing wherever nobody but a person may answer — a confirmation, an irreversible act, or a permission that is not cheap to take back. That condition is `questionHandsOnly`, which is the predicate `d`/`D` were already reading, lifted out so the two cannot disagree. Every other kind still opens on its first answer: on a landing row or a proposal the safe answer is the one that does nothing, and a pointer parked there would make `enter` mean "no" on every card. Both stale tests are rewritten to the law that is now true.

### Also

- `internal/manual/chat/questions.md` — two new sections under the room's page (`Moving around the page — the arrows, and clicking an answer`, `What each line under an answer means`), the page's own list grew the evidence section, and the key table says what the arrows and `o` do inside the page.
- `docs/design/questions/DESIGN.md` — the **room** bullet and the **HUE** law.

### Checks

- `make test-laws` green
- `go test ./internal/manual/ ./internal/e2e/ -count=1` green
- full `./internal/tui3/` on the Spark — result posted as a comment
- driven as a person in tmux against a real model — screens posted as a comment


santoshkumarradha:
### The full `internal/tui3` suite on the Spark — **green**

```
$ GOMAXPROCS=4 GOFLAGS=-p=2 go test ./internal/tui3/ -count=1 -timeout 15m
ok  	github.com/Agent-Field/aforge-v2/internal/tui3	422.250s
```

Run on `spark` in a detached worktree at head **`2dd197d0096271ecc11170190de27518795f1dc2`**. Also green here: `make test-laws`, `go test ./internal/manual/ ./internal/e2e/ -count=1`.

Worth recording, because it is the reason three commits landed instead of one: **four tests in this package were already red on `fix/question-pointer` (#789)** before this branch touched anything. All four are green now.

| test | why it was red on #789 | what fixed it |
| --- | --- | --- |
| `TestEnterTakesThePickOnlyWhereThereIsOne` | stated the pre-pointer law; and underneath it `enter` on the gate for `rm -rf *` meant `allow once` | the pointer opens on the answer that loses nothing where nobody recommended anything; test rewritten to the law that is now true |
| `TestTheFootOffersNoAnswerUntilThereIsOne` | same, on the foot: a question with answers always has something to take | test rewritten to the shape that really drops the key — no answers at all |
| `TestAnEmptyBoxAnswersNothingAndTheWayOutStillDoes` | `enter` over an empty box on an `InputText` question (a connect key) took the pointer's answer, which reads as a decline | `questionEnter` takes nothing from an empty box when the answer is a sentence |
| `TestACardDrawsARowPerAnswerAndMarksTheAskersPick` + two in `bundle_test.go` | red on the *first* cut of the safety fix, not on #789 — opening on the safe answer everywhere moved a proposal's pointer off `start it` | the condition sits **below** the pick, so a proposal is untouched |

---

### Driven as a person, in tmux, on the Spark

`bin/aforge chat --no-host` in a real 100×36 terminal, real keys and real SGR mouse presses, with the page's own fixture (`AFORGE_QUESTION_DEMO=reading`, `internal/tui3/questiondemo.go`) supplying a question that has bodies, `+`/`−` lines, a consequence, a pick with a reason and what would change its mind, and a diagram under an answer. Screens captured per key with `tmux capture-pane`.

**On opening** — three tiers, each aside behind its own word, the foot naming the pair that actually walks the page:

```
‹ back
? which store should the ledger sit on?
  aforge · a schema change is next and it is cheaper before there are rows
  the turn waits on it
  ▾ 1 postgres                                                               my pick · fairly sure
      Rows already carry a foreign key into it and the migration is one file.
      + one place to back up
      - another service to run locally
      then · the ledger and the rest of the project share one connection
      why this one · because it is the only store the reporting job already reads
      would switch if the ledger ever has to run on a machine with no server on it

      what it would look like
        app --> pg --> report
  ▸ 2 sqlite beside the project
  ▸ 3 a file per day
─────────────────────────────────────────────────────────────────────────────────────────
nothing chosen yet · just this once
[enter] take it · [esc] later · [x] compare · [?] ask back · [d] you decide · [↑↓] choose
```

**`↓` then `→`** — the band moved and the second answer opened (the band is in the escapes; here it is, off the captured `.ansi`, before and after `↓`):

```
before:  ESC[48;5;235m  ESC[38;5;243m▾ESC[39m ESC[1mESC[38;5;140m1 postgresESC[0m
after:                 ESC[38;5;243m▾ESC[39m ESC[1mESC[38;5;140m1 postgresESC[0m
         ESC[48;5;235m  ESC[38;5;243m▸ESC[39m ESC[1mESC[38;5;140m2 sqlite beside the projectESC[0m
```

```
  ▾ 2 sqlite beside the project
      One file in the repository's own folder.
      + nothing to run
      - the reporting job reads a copy
      then · the ledger travels with the checkout
  ▸ 3 a file per day
```

**One click on the third answer's row** — it moved there and opened, and did **not** answer:

```
  ▾ 3 a file per day
      Append-only, one file a day, nothing to migrate ever.
      + nothing to migrate
```

**A second click on that same row** — it answered, and because the fixture's question is not one the session raised, the engine's refusal is drawn where the foot was, which is the honest end of that path:

```
session: that answer names nothing — pick one of the answers the question offered, or say what you …
```

Two things I could not show here and why: the model I drove first (`deepseek/deepseek-v4-flash`) could not call `ask` with a block per option at all — three `Invalid arguments: options takes a list` and then `stuck: ask has failed the same way 3 times` — so the question came from the page's own fixture instead of from a turn; and the fixture carries no `Question.Attach`, so the `what it showed you` section is pinned by `TestTheQuestionsOwnEvidenceIsOneTitledSectionAboveTheAnswers` rather than by a screen.

santoshkumarradha:
Rebased onto dev after #789 merged (the lane branched before #789's last three commits; the block's `c`/`?` prompt and its test were restored during the rebase). Spark: `internal/tui3` full suite **ok 422.2s** at eff116d78. Gate green.

PR #790, opened 2026-09-10
Remove the v2 remnants: the root compositor, the model picker, the blocks engine

AbirAbbas:
## What this removes

Everything the v2 chat surface left behind that nothing has drawn with since it
went on 2026-08-31 (entry 111). **65 files, 17,906 lines deleted; 96 files
changed, 177 insertions, 18,713 deletions.** No behaviour changes.

| What | Lines | Why it is dead |
| --- | --- | --- |
| `internal/tui2/modelui` — the chip and the picker, 6 files + 6 tests | 4,773 | the only symbol any package outside `internal/tui2` names from `modelui` is `ModelWord`, in `word.go`, called once from `internal/tui3/spendplace.go`. `word.go` and `word_test.go` stay. |
| `internal/tui2` root — the shell, compositor, layout solver, panes, placeholder, capability probe, OSC writer, 19 files | 5,576 | its only non-test importer was `modelui/picker.go`, deleted above |
| `internal/tui2/blocks` — the whole rendering engine | 6,864 | `internal/tui3` and `cmd` never imported it. After the two rows above its only non-test importer was `tokens/styler.go`, and only for two `var _` assertions and the four methods written to satisfy them |
| `internal/tui2/tokens/breakpoints.go` + tests | 442 | after the rows above, no symbol declared in it is named from outside `internal/tui2`. `ContextToken`'s one external caller was `modelui/chip.go:162`; the footer registry, `FitFooter`, `SplitDiffAtWidth` and `HUDRowCap` had none |
| `internal/tui2/tokens/identity.go` + test | 251 | `IdentityFor`, `IdentityNext`, `AssignIdentities` have no production caller anywhere in the module — a dead file inside a live package |

Also removed from `tokens/styler.go`: `Paint`, `PaintGlyph`, `PaintIdentity`,
`Token(state, hue)` and the two `blocks` interface assertions. Nothing outside
`internal/tui2` called any of them. `PaintToken`, `PaintOn`, `PaintRowOn`, `Fg`
and `Glyph` — what v3 paints through — are untouched.

**What stays, and why:** `internal/tui2/tokens` is the icon and colour
vocabulary v3 draws with and carries its own law; `internal/tui2/prose` renders
v3's markdown; `internal/tui2/reltime` is v3's and `internal/subharness`'s; and
`modelui/word.go` holds `ModelWord`. `internal/tui`, `internal/command`,
`internal/head` and `internal/resident` are **not touched** by this pull request.

[...]

PR #792, opened 2026-09-10
Remove the v1 window: 111 lines move out of internal/command, 36,037 go

AbirAbbas:
> **Stacked on #790.** Base is `chore/remove-v2-remnants`, so this diff shows
> only the v1 window. Merge #790 first, or hold this one independently — that is
> why it is a separate pull request.

## The shape of it: 111 lines move, 36,037 go

`internal/tui` (32,669 lines) and `internal/command` (3,368) have been
unreachable from every shipped `main` for weeks, held up by **two thin edges**:

- `internal/command`'s only non-test importer is `cmd/aforge/chat.go`, which names
  exactly four package symbols: `command.NewMediaModels`, `command.LoadPrefs`,
  `command.AttachmentStoreRoot`, `command.NewSessionID`.
- `internal/tui`'s only non-test importer is `internal/command`, and that edge
  carries **no behaviour** — ten data types and two `var _` interface assertions.

So this extracts **111 lines** into a new `cmd/aforge/chatprefs.go`, rewires
`chat.go` in three places, and deletes both packages. **108 files changed, 161
insertions, 37,236 deletions.**

The blank identifier is why this looked alive for so long:
`var _ tui.Commander = (*Commander)(nil)` is not a caller, it is a promise about
a shape — and read as a declaration it keeps every type it names looking
reachable. Any name-based dead-code pass has to special-case `_`.

## Four symbols from the moved ranges were deliberately left behind

Moving them would have moved dead code into a live package, against CLAUDE.md's
"a capability that cannot work is absent, not broken":

| Symbol | Its only callers |
| --- | --- |
| `fallbackModels` | `Commander.Models` (`command.go:400`) |
| `MediaModels.Set` | `Commander.SetModel` (`command.go:692`) — dropping it also drops the `internal/catalog` import the new file would carry |
| `SavePrefs` | `Commander.SetModel` (`:688`), `Commander.SaveSplitPct` (`:771`) |
| `scratchRootFor` | `command.New` (`:248`), `NewVisitor` (`:1245`), and its own test |

The three moved functions take the spellings `cmd/aforge` already calls them by —
`loadChatPrefs`, `attachmentStoreRoot`, `newSessionID` — so **no call site
changes** and no name forwards to itself.

## `chat.go` does not collide with #651

Open pull request #651 changes `cmd/aforge/chat.go` at 1027-1065, inside
`buildBrain`'s body. This change's hunks, verbatim from
`git diff origin/dev -- cmd/aforge/chat.go`:

```
@@ -19,7 +19,6 @@ import (
@@ -214,7 +213,7 @@ func buildBrain(w *chatWindow, session string, opts brainOptions) (*chatBrain, e
@@ -3305,16 +3304,6 @@ func withDocumentAttachmentBrief(brief string, paths []string) string {
```

Three hunks, none between 1000 and 1100.

## The tests: one law re-pointed, fourteen deleted with their subject

The rule was **the code under test decides.** A test whose subject survives is
re-pointed; a test whose subject is `*command.Commander` cannot be, because the
subject is what is being deleted.

**Re-pointed:** `TestEveryChatPreferenceIsFrontedByASettingsRow` — a reflection
law over `Prefs`' fields against the rows `config.NewSettings` builds. Both
halves are live; `Prefs` just moved. Its two exemption comments stop citing
`internal/tui2/modelui`, `internal/tui2/chat` and `internal/command`'s
`CurrentModel`; the reasons themselves are unchanged and still true.

**Kept whole:** `chat_test.go`'s other **28** tests — the delivery gate,
learned-facts parsing, the consolidator, reflex envelopes, resident briefs, the
unasked gate, budget-stopped partials. That is live `aforge do` behaviour that
was only ever wearing a v1 driver, and deleting the file would have been the one
real regression in this wave. Only `TestNodeTraceWindowStartsOnALineBoundary`
goes with the trace reader. (`command.Seq` at `chat_test.go:467` is a field on a
local variable named `command`, not the package — it stays.)

**Moved, not deleted:** `voiceRoundTripFunc`, a plain `http.RoundTripper`
adapter that lived in `voice_models_test.go` and is used by
`context_budget_test.go`.

**Deleted with their subject, and where the behaviour is covered now:**

| Test | Covered now by |
| --- | --- |
| `TestVoiceModelDiscoveryFiltersTranscriptionOutput` | `internal/tui3.TestEachModelSlotAsksItsOwnQuestion` |
| `TestVoiceModelCatalogUsesStaleCacheOfflineAndHasDefaultFallback` | `internal/catalog.TestStaleCacheWinsOverOfflineAndEmptyUsesDefaults` |
| `TestVoiceModelChoicePersistsInChatConfig` | `internal/tui3.TestAMediaSlotPickWritesTheKeyTheResolverReads` |
| `TestModelPaletteCatalogFiltersEveryCapabilitySlot` | `internal/tui3.TestEachModelSlotAsksItsOwnQuestion` |
| `TestStandingSlashListsOnlyActiveCharters` | `internal/store.TestCharterSpecDissolvesIntoCanonicalLifecycle` |
| `TestCancellingAWorkerLeavesThePartialForTheRestartToRead` | `internal/store.TestCancellationKeepsWhatTheWorkerHadWritten` |
| `TestChatNotebookSearchEvidenceAndRetractMoment` | partly — `internal/store/fact_retrieval_test.go` covers the search and `internal/store.TestFactQuarantineExcludesRetrievalAndCanBeRehabilitated` / `TestUserRetractedBeliefStaysDownWhenTheDistillerRederivesIt` cover the retraction. **The `· let go — …` moment posted into the thread was the commander's and is not covered.** |
| `TestChatNewSessionAdvancesAttachedWatermark` | partly — `internal/store.TestSeenWatermarkIsJournaledAndSurvivesRebuild` covers the watermark; the commander's `NewSession` wiring is not covered |
| `TestVoiceSpendUsesTheDurableUsageLedger` | **NO EQUIVALENT FOUND** |
| `TestBoostPreferenceFollowsWorkPersistsOverrideAndCanBeCleared` | **NO EQUIVALENT FOUND** |
| `TestPlanPreferenceFollowsWorkPersistsOverrideAndCanBeCleared` | **NO EQUIVALENT FOUND** |
| `TestMediaModelPreferencesPersistAndUpdateFutureLeafSnapshot` | **NO EQUIVALENT FOUND** |
| `TestBudgetSlashForms` | **NO EQUIVALENT FOUND.** `internal/tui3.TestEveryDoorLandsOnTheSameEditor` covers `/budget`, `/budget 50` and `/budget none`, but not `/budget default 35` or `/budget unlimited today` |
| `TestTheCommanderRestartJournalsTheDurableRestartCommand` | **NO EQUIVALENT FOUND** |
| `TestChatCommanderBuildsALiveRegistry` | its subject was the commander's settings registry |

Six behaviours have no current-surface test. They are not regressions — the code
that implemented them is gone, and v3 has its own doors — but if v3's
follow-work/override model preferences, its voice spend ledger or its `/budget
default` and `/budget unlimited today` forms are meant to be pinned, they need
tests of their own. That is a decision worth making knowingly rather than
discovering later.

## The worker trace now has no reader

`Commander.NodeTraceSince` and `NodeTraceTail` were the **only** production
callers of `exec.TracePath` in the module. `internal/tui3` and
`internal/session` contain no reference to it at all — v3 never read a worker's
`.aforge/trace/` file this way.

After this change `internal/exec/executor.go` still **writes** those files and
nothing in production reads them; `exec.TracePath` survives with only its own
test. Not a regression, but a fact the owner should hold: either v3 grows a
reader, or the trace is write-only telemetry whose writer earns its own audit.
`command.go:917-940` documented a repair to exactly this reader, so it is worth
saying out loud that the empty-feed defect it describes cannot recur — there is
no feed.

## go.mod

`internal/tui/charmbubbles/` was the `replace` target for
`github.com/charmbracelet/bubbles`, and its only importers were
`internal/tui/model.go` and `internal/tui/settings.go`. The `replace` and the
`require` both go; `go mod tidy` then dropped `charm.land/lipgloss/v2` (a direct
dependency of the v1 window alone), demoted `colorprofile` and `termenv` to
indirect, and wrote down two transitive requirements the graph already carried
sums for. `grep -rn 'charmbracelet/bubbles'` over the tree, `go.mod` and
`go.sum` finds nothing.

**A `go.mod` change makes CI's `touched packages` job run the WHOLE tree**
(`.github/workflows/ci.yml`), so the acceptance here is the full suite, not a
subset.

## The three laws, and the one deliberately left alone

`internal/iconlaw`'s `surfaces` drops `"../tui"` — and its comment said "A
package joins this list the day its sweep lands, and **never leaves it**", so
that sentence now says the true thing: a package leaves when the package stops
existing, which is not the same as a live surface quietly dropping out of the
sweep. `internal/guard`'s `lockedPackages` drops the path (its `WalkDir`
propagates errors, so a missing directory would have broken the walk);
`lockScanFloor = 60` is untouched and the run still reports the same 40 loose
sections. `internal/lane`'s forbidden-import list drops the path.

`.github/known-red.txt` and `internal/ci`'s `knownRedEntries = 1` are
**untouched**: the single entry is `TestEveryLockInTheGuardedTreeUnlocksFromADefer`
and it is still red for its own reason.
`internal/registry/registry_test.go` parses `internal/head/toolbelt.go` **by
path**, and `internal/head` is not touched here, so that law is untouched too.

## What this does NOT touch

`internal/head` (its conversational `Head` cohort needs a `go/types` dead set,
and `internal/registry`'s law has to move off parsing `toolbelt.go` by path
first) and `internal/resident`. Both are follow-ups.

## Gates — run in this worktree, results verbatim

| Gate | Result |
| --- | --- |
| `go build ./...` | pass, exit 0, no output |
| `go vet ./...` | pass, exit 0, no output |
| `make fmt-check` | pass, exit 0, no output |
| `go build -tags=aforge_packed_manual ./...` | pass, exit 0, no output |
| `grep -rn 'aforge-v2/internal/command' --include=*.go .` | no output (exit 1) |
| `grep -rn 'aforge-v2/internal/tui"' --include=*.go .` | no output (exit 1) |
| `grep -rn 'charmbracelet/bubbles' --include=*.go . go.mod go.sum` | no output (exit 1) |
| `make test-laws` | pass — `laws: 63 files in 18 packages`, 18 × `ok` |
| `go test ./internal/manual/` | pass — `ok … 1.506s` |
| `go test -run 'Manual' ./internal/tui3/ ./internal/session/` | pass — `ok` both |
| `go generate ./internal/manual && go test -tags=aforge_packed_manual ./internal/manual` | pass — `ok … 1.501s`, no tracked file dirtied (the packed corpora are ignored build products) |
| `go test -count=1 -timeout 20m ./cmd/aforge/` | pass — `ok … 102.027s` |
| `go test -count=1 -timeout 15m ./internal/tui3/` | pass — `ok … 420.480s` |
| `go test -count=1 -timeout 15m ./internal/session/` | pass — `ok … 231.217s` |
| `go test -count=1 ./internal/store/` | pass — `ok … 53.753s` |
| `go test -count=1 ./internal/exec/` | pass — `ok … 70.013s` |
| `go test -count=1 ./internal/iconlaw/ ./internal/registry/ ./internal/lane/...` | pass — `ok` ×4 |
| `go test -count=1 ./internal/e2e/` (untagged) | pass — `ok … 0.224s` |
| `go test ./internal/guard/ -run TestEveryLock` | **fails for its known reason** — the single entry in `.github/known-red.txt`, skipped by `make test` and `scripts/laws.sh`. Same 40 loose lock sites as before, no missing-directory or scan-floor error. |
| `go test -count=1 ./internal/config/` | **fails on `TestRegistryCoversEveryUserFacingEnvironmentPin`** — `AFORGE_QUESTION_DEMO (internal/tui3/questiondemo.go) is neither a settings row nor operator plumbing`. **Reproduced on clean `origin/dev`**; the pin arrived with #780. Not this branch's, and not in the known-red ledger — see the note on #790. |
| whole-tree `make test PKGS=./...` | running at the time of opening; the result is posted as a comment, since the `go.mod` change makes CI run it |


AbirAbbas:
### The whole-tree suite: 85 packages `ok`, two reds, both dev's

A `go.mod` change makes CI's `touched packages` job test the whole tree, so the
whole tree is the acceptance here. Run as CI runs it —
`make test TEST_FLAGS='-count=1 -p 1' PKGS='./...'`, which is
`go test -timeout 15m -count=1 -p 1 -skip '^(TestEveryLockInTheGuardedTreeUnlocksFromADefer)$' ./...`:

**85 packages `ok`.** The heavy ones, verbatim:

```
ok  github.com/Agent-Field/aforge-v2/cmd/aforge         96.325s
ok  github.com/Agent-Field/aforge-v2/internal/tui3     422.445s
ok  github.com/Agent-Field/aforge-v2/internal/session  253.137s
ok  github.com/Agent-Field/aforge-v2/internal/resident 213.685s
ok  github.com/Agent-Field/aforge-v2/internal/store     70.047s
ok  github.com/Agent-Field/aforge-v2/internal/remote    25.785s
ok  github.com/Agent-Field/aforge-v2/internal/workspace  4.040s
```

**Two failures, and both reproduce on clean `origin/dev`:**

| Red | Verdict |
| --- | --- |
| `internal/config` `TestRegistryCoversEveryUserFacingEnvironmentPin` — *AFORGE_QUESTION_DEMO (internal/tui3/questiondemo.go) is neither a settings row nor operator plumbing* | **pre-existing on clean `origin/dev`**, reproduced in a detached worktree at `6836c4c22`. The pin arrived with #780. It is red on a clean tree and **not** in `.github/known-red.txt`, and `make test-laws` does not select it. Worth its own issue. |
| `internal/revision` `TestAnEmptyFirstVerdictIsAskedAgainWithRoomAndItsAnswerStands` — *the retry did not get more room: [0 0]* | **pre-existing on clean `origin/dev`**, reproduced standalone |

Neither package reaches any `internal/tui2` package, neither imported
`internal/tui` or `internal/command`, and this branch changes no file in either.

`make test-packed-manual`: pass. The ledger's single entry
(`TestEveryLockInTheGuardedTreeUnlocksFromADefer`) is skipped by the same target
CI uses, and it fails for its own reason with the same 40 loose lock sites as
before — no missing-directory error from the removed walk, and `lockScanFloor`
did not move.

### CI does not run on this pull request, and that is a finding

`.github/workflows/ci.yml` triggers on `pull_request: branches: [dev]`. This
pull request's base is `chore/remove-v2-remnants`, so **the PR gate never fires
on it** — `gh pr checks 792` reports "no checks reported". A stacked pull request
in this repository gets no CI at all until its base merges and GitHub retargets
it to `dev`.

So I dispatched the same workflow against the branch by hand
(`gh workflow run ci.yml --ref chore/remove-v1-window`): **light gate success,
check success** — that is `go build ./...`, `go vet ./...`, `make fmt-check`, the
packed corpora regenerating clean, the three manual gates, and `make test-laws`,
all on GitHub's runner. Its `touched packages` job reported *"No Go file and no
module file changed; nothing to run"*, because a `workflow_dispatch` has no base
to diff against and falls back to `HEAD~1`, which here is the change entry
alone. That job's green is therefore not evidence; the whole-tree run above is.

### `make size`: 0.9 MB off dev, and still over a budget dev is already over

| | bytes |
| --- | --- |
| clean `origin/dev`, this machine, same toolchain | 56,500,489 |
| `chore/remove-v2-remnants` (#790) | 56,471,817 |
| this branch | **55,591,177** |
| `SIZE-BUDGET` | 54,600,000 |

Together the two pull requests take **909,312 bytes** off dev. The budget was set
on `linux/arm64` and this machine is `x86_64`, so dev is already 1.9 MB over it
here; `make size` is red before and after, and it is not on the pull-request
gate. The number is left alone deliberately — it is a ceiling, and lowering it
is a policy call for the owner.

Worth noting against the 37,236 deleted lines: the linker had already dropped
almost all of them. 0.9 MB is what a genuinely-referenced vendored `bubbles`
fork and `lipgloss/v2` were costing.

AbirAbbas:
Rebased onto `dev` after #790 was squash-merged as 0117888b1.

This branch was stacked on #790, so it still carried #790's eight commits on top of the pre-merge base. Replayed only the nine commits that belong to #792:

```
git rebase --onto origin/dev bfc964b78 chore/remove-v1-window
```

where `bfc964b78` was #790's tip. Because the replay base is #790's tip and `dev` already contains #790's content through the squash, both files `git merge-tree` had flagged against the old base replayed clean. I checked each one keeps **both** intents rather than one side:

- **`CLAUDE.md`** — the two sides edit disjoint regions, so nothing was dropped. `dev`'s new text from #798/#799 is intact: the `make pr-ready` merge ritual, the "proof must not waste the box" bullet, the `test-touched` / `pr-ready` rows in the targets block, and the two Learned-\* sections. This branch's edits are also intact: the `internal/tui` row is gone from the surface table, the `internal/tui2` row names the three surviving packages, and the icon law now lists `internal/tui3`, `internal/head`, `internal/resident`.
- **`internal/lane/structure_test.go`** — `dev`'s forbidden-import list already lacked the `internal/tui2` root entry, since that was #790's change and is now in `dev`. So the only remaining edit here is this branch's own: drop `internal/tui`. `net/http`, `internal/provider`, `internal/tui3` and `internal/session` are untouched.

Nothing was resolved by taking one side wholesale. The diff against the base is now 109 files, +192 / −37,236 — #790's half no longer double-counts. `go.mod` is still `go mod tidy`-clean.

## Gates

Green on the rebased tip: `go build ./...`, `go vet ./...`, `make fmt-check`, `make test-laws` (18 packages), `make test-packed-manual`, `go test ./internal/e2e/` (untagged), and `go test ./internal/lane/ ./internal/iconlaw/ ./cmd/aforge/`. `internal/guard` fails only `TestEveryLockInTheGuardedTreeUnlocksFromADefer`, the single entry in `.github/known-red.txt`, which `make test` and CI both skip by name.

## The red on `check` is not from this branch

Because this branch edits `go.mod`, the `touched packages` job derives **the whole tree** rather than a package subset, so it runs packages a normal pull request never reaches. Six tests fail, in three packages this branch does not touch a single file of:

| Package | Test | Reproduces on the base commit? |
| --- | --- | --- |
| `internal/config` | `TestRegistryCoversEveryUserFacingEnvironmentPin` | yes — `AFORGE_QUESTION_DEMO` in `internal/tui3/questiondemo.go` is not registered in `settings.go` |
| `internal/revision` | `TestAnEmptyFirstVerdictIsAskedAgainWithRoomAndItsAnswerStands` | yes — `budget_test.go:101: the retry did not get more room: [0 0]` |
| `internal/tui3` | `TestEnterConnectsOpenRouterInTheBrowserAndHandsTheKeyToThisProcess`, `TestAJobsLogReaderTakesOneLastReadingAfterTheJobEnds`, `TestAPathInTheBoxIsNeverChipped`, `TestAnAbsolutePathDoesNotHoldTheCommandListOpen` | yes — all four, identically |

I checked out the base commit (436aa8426, this pull request's own base, with #792 absent) in a separate worktree and ran each of them there: all six fail with byte-identical messages. The four `internal/tui3` ones fail in isolation too, not only under whole-tree load, so they are deterministic rather than flakes. `internal/tui3` also compiles neither of the dependencies `go.mod` stopped requiring directly, so this branch cannot reach it.

So these are pre-existing reds on `dev`, surfaced here only by the whole-tree derivation. They are invisible on an ordinary pull request, and on a push to `dev` the same job diffs only the merge commit — which is why nothing caught them. Adding them to `.github/known-red.txt` is not the move, since that ledger only shrinks; they want their own fix, and fixing them inside this removal would bundle three unrelated packages into it.

PR #794, opened 2026-09-10
A model whose budget is spent on refusals moves to the next one, exactly as a cut stream does

santoshkumarradha:
## What was wrong

A chat turn on `deepseek/deepseek-v4.1-flash` took one 502 and three 429s between 13:00:17 and 13:01:32 on 2026-09-10 and ended. The journal reads:

```
{"class":"transport","reason":"the endpoint refused","action":"give_up","attempt":4}
```

then `EventError` with `after 3 retries: …`. In that same minute `z-ai/glm-5.3-flash` was answering every task-node call in the same session. **The fallback chain was never asked.**

The model hop was reachable only from the stream-cut branch (`provider.CutFrom(err)`). The refusal branch walked `failureLimits().TransportAttempts` with `backoffWait` and then returned. So a turn had two answers to "this model's budget is spent" — one that moved and one that stopped — depending on which kind of failure happened to arrive.

## What this does

**One exit from exhaustion, decided in one place.** The decision moved into the policy object that already answers "what was that failure worth" — `internal/taxonomy`:

- `ActionHop` joins `ActionRetry` and `ActionGiveUp`.
- `transportPolicy.Decide` walks **one budget with two kinds of spending**: an outright failure spends a rung off `Limits.TransportAttempts` with its doubling wait; a cut stream spends a shorter allowance (`SilentCutAttempts`, `BlindCutAttempts`, `DegenerateCutAttempts`) with no wait at all, because the request was served and the *reply* came apart.
- The difference between moving and stopping is `Evidence.FallbackAvailable` — the caller's fact. The policy never names a model.

The turn loop's two roads are folded into one. `cutBudget` and its three constants are gone; the loop reads a verdict and acts on it. A cut and a refusal now differ only in the evidence they arrive carrying.

`movesForFailure` (a task node's tier purchase) already routes through the boundary and is deliberately untouched — that is a *tier* decision, not a chain hop.

## The surface can draw one honest row

`EventRetrying` grows `Event.Retry *RetryNews`:

| field | what it is |
| --- | --- |
| `Model` | the model whose attempt just failed |
| `Attempt`, `Attempts` | how far into *that model's* patience — the policy's number, not a constant a surface holds |
| `Reason` | why, in the person's words (`the model would not take the request`) — never the journal's `the endpoint refused` |
| `Next` | the model being moved to; **empty on an ordinary retry**, which is the distinction a surface could not draw before |

It is filled on **every** `EventRetrying` the loop sends, cut notices included. `Text` is still the whole sentence, so a surface that reads only `Text` is unchanged. It rides the remote link under its own json key, and a peer built before it exists sees none of it.

## The exits that must not move, and do not

- `provider.IsConnectionUnavailable` — returns as it always did.
- `RefusalFrom(err).OurRequest()` — a 4xx that named no upstream is our own bytes; no retry, no move.
- context overflow and non-retryable shapes.
- `--one-model` — now refused in the loop as well as at the door that empties the chain, so the promise holds against any completer.
- A strict lane pin is untouched: it is a person's answer about a *machine*, and each model carries its own.

## Tests

- `internal/taxonomy`: a spent ladder gives up alone and moves with a model in hand; a move buys no tier, ends no turn and waits for nothing; each cut shape spends its own allowance and ends the same three ways.
- `internal/session`: four refusals then an answer on the fallback — the move is announced with `Retry.Next`, the four journal rows are all on the model that failed and numbered 1..4, the last reads `hop`, and the sealed turn names the model that answered. A walked-out chain names every model tried. `--one-model` never moves. Our own bytes return at once.
- `internal/remote`: the payload crosses the link, and an older peer's retry arrives with none.
- The manual page that said `never moves your model` is gone, replaced by what the turn now does with the three sentences it actually spells.

## Evidence

`make test-laws` green locally (18 packages). `go test ./internal/e2e/` untagged green. Full `./internal/taxonomy/ ./internal/session/ ./internal/provider/ ./internal/remote/ ./internal/manual/` on the Spark — result in a comment below.

https://claude.ai/code/session_01WiyuVEaLLjV99zaGmTnqGy


santoshkumarradha:
## Spark evidence

Rebased onto `dev` after #793 (the refusal door) merged, so the in-stream 429 now arrives at `wireEvidence` carrying its served machine and status exactly as an HTTP one does — one reader, `provider.RefusalFrom`, unchanged.

**Tested SHA `9275b1aaee89d561b45aebf3c1871868ea9197de`** (worktree off `origin/fix/refusal-hop`, go1.26.5 linux/arm64, `-count=1 -timeout 30m`):

```
ok  	github.com/Agent-Field/aforge-v2/internal/taxonomy	0.003s
ok  	github.com/Agent-Field/aforge-v2/internal/session	191.371s
ok  	github.com/Agent-Field/aforge-v2/internal/provider	54.973s
ok  	github.com/Agent-Field/aforge-v2/internal/remote	25.770s
ok  	github.com/Agent-Field/aforge-v2/internal/manual	1.486s
ok  	github.com/Agent-Field/aforge-v2/internal/ci	0.660s
ok  	github.com/Agent-Field/aforge-v2/internal/e2e	0.263s
```

`go build ./...` green on the same tree. Locally: `make test-laws` green (64 files, 18 packages), `go test ./internal/e2e/` untagged green, `make changelog-check` green (416 entries). Nothing added to `.github/known-red.txt`.

PR #802, opened 2026-09-10
A model picked while a task is being checked is the next run's, not a rewrite of what ran

santoshkumarradha:
`RetargetTask` decided what a pick could move from the node's STATE alone. A node is `TaskRunning` across three lives — its own worker, the check reading what that worker left, and every repair round — so the worker's reading is over the moment `runTaskChild` returns, which on a checked node is minutes before anything settles.

A pick made in that window rewrote the frozen `spec.model`, so the row, the checkpoint and the landed card all named a model that never ran a token of the work, while the bill named the one that did.

It reads `node.lifeNow() == TaskPhaseChecking` now — the same word the steering door already uses under *THE CHECK HAS NO READER* — and a pick with no turn left to reach lands in `nextModel`, where a settled node's pick already lands. The room draws `next model <id>` and the sidebar heads itself `Next run setup`, so no surface learns new vocabulary.

Measured on a real run: a node admitted on the install's worker tier with all nineteen of its calls billed there, and a card claiming the low tier — the model that had only ever run the hidden `router`, `taskname` and `caption` roles.

`make pr-ready` exited 0. New tests: `TestAModelPickedWhileTheCheckReadsBecomesTheNextRunsAndLeavesTheRowAlone` and `TestAModelPickedWhileTheWorkerReadsStillMovesTheWork` in `internal/session/taskretarget_test.go`.


santoshkumarradha:
Not merging this alone: tagged e2e / `TestTUIE2E` is not green on this branch by itself. Combined with #804 and #806 as https://github.com/Agent-Field/aforge-v2/pull/807 and proving that stack on Spark.

santoshkumarradha:

Issue #803, opened 2026-09-10
A 40ms audit window makes two true landings race, and the unattended subtest fails under load


santoshkumarradha:
## What happened

`internal/session`'s `TestAWindowOneStallClosedIsDecidedByThePosture/left_running_with_a_ceiling`
fails on a loaded box. Measured 2026-09-10 on `dev@64dc75f28` (macOS, 10 cores, eight
busy-loops running beside it). It is not caused by the change that found it — the same
binary built from `64dc75f28` and from a branch on top of it fail at the same rate in
alternating rounds:

```
--- FAIL: TestAWindowOneStallClosedIsDecidedByThePosture (1.49s)
    --- FAIL: TestAWindowOneStallClosedIsDecidedByThePosture/left_running_with_a_ceiling (0.62s)
        unattendeddoor_test.go:1097: the landing is missing "without answering and was abandoned":
            Wrote greet.go with the greeting.
            taken as it stands: nobody could check it — asked twice, and neither call answered — nobody could check it in 40ms, and the run is unattended
```

The test asks for a landing that carries `checkerStalled` — "one call ran <bound> without
answering and was abandoned" — because the run it builds hangs both checking calls inside a
40ms window, and the first is meant to be abandoned at its own bound with the second asked
after it. Under load the landing carries `checkerAskedTwice` instead: both calls were made,
neither answered, no call is reported as having held its bound. Everything else the subtest
asserts is present, including `checkerWindowClosed` and `checkerRanOut`.

This is the shape AGENTS.md says is a bug report rather than a known shape — it fails only
when other work is running beside it — so it is filed rather than rerun in isolation. It is
NOT in `.github/known-red.txt` and must not be added to it.

## Replication

**Deterministic (no model).** The test's own completer is a stub; only the clock is real.

```sh
for i in $(seq 1 8); do (while :; do :; done) & done      # load the box
go test -count=15 -run 'TestAWindowOneStallClosedIsDecidedByThePosture' ./internal/session
kill %1 %2 %3 %4 %5 %6 %7 %8
```

On an unloaded laptop this is green at `-count=15`. Loaded, it failed 1 in 10 runs in two of
six alternating rounds (three rounds each of `dev@64dc75f28` and a branch on top of it,
prebuilt with `go test -c` so compilation was not part of the measurement). The failure is
the block quoted above.

**Field (real models).** None needed — no key, no network, no cost. The window under test is
[...]

PR #804, opened 2026-09-10
A handover no longer lists finished work as what is left

santoshkumarradha:
A conversation hit its checkpoint ceiling, split the rest into a sketch, wrote a
2,641-token handoff and launched the first task. The worker re-read files the
parent had already read and re-derived conclusions the handoff had already
stated.

The tempting diagnosis is that it was short of context. It was not. It had the
handoff, the evidence and the quotes. What it had was **a document that said two
different things about the same work**: the machine-composed line at the head of
the brief, `WHAT IS LEFT, AS PARTS:`, named reading two files as remaining work,
while the calls-already-run section further down named the same two reads, with
their pointers, as done. The worker obeyed the louder and earlier half.

Three sentences, in the three places that were silent. Nothing structural — no
new field on the brief, no new section, no result bytes anywhere near an
admission handle.

### 1. The second reader is told that finished work is not a part

`checkpointSketchAsk` already excluded work handed out. It now excludes work
already done, in the same voice, appended after that clause:

> Work that is already done is not a part either: what the account above shows
> finished is not what remains, and drawing it sends somebody to do it a second
> time.

The measured wording of everything before it is untouched — no digits, none of
the narrowing words `TestTheSketchAskIsTheMeasuredWording` forbids.

### 2. The handoff's writer is told where the line between its first two things falls

`checkpointHandoffWriteAsk` named four things and left this to inference. It now
says it, between the four things and the closing instruction about the person's
own words:

> Nothing that is already done goes under what is left to do: what has already
> been read, run or found out is what they already know, and putting it under
> what is left sends them to do it again.

### 3. The worker is told which half of its own brief wins

`admissionEvidenceRule` described the calls-already-run section without ranking
it against anything, so a contradiction had no resolution. Appended:

> AND THIS SECTION IS WHAT HAS ALREADY HAPPENED, which is what settles a
> disagreement with the work above: where the parts or the brief read as though
> one of these calls were still to be made, it has been made already — read it
> through its pointer instead of running it again, and run it again only where
> the line says it failed or where what you read disagrees with the brief.

**The body-absence law is untouched.** No result bytes or excerpts were added to
a handle; the bytes stay where they are and the worker fetches what it needs.
Entry room inside the 5,000-byte `admissionBudget` follows the rule's length by
itself through `admissionOverhead` — 3,707 bytes, from 4,030, and the budget
test still passes.

### And the folder the same brief could not write to

The same run's brief named `<session>/work/task-creation-flow.md` as its
deliverable, which is what `deliverablesDir` answers with for a conversation
that owns its workspace. That folder is outside the ground, so `taskCopy` rightly
left it alone, `taskGroundGuard` refused the first write there, and the worker
invented an address inside its own copy and reported the deviation. `newTaskCopy`
now takes that folder as a third argument and binds it to the same path under the
copy, and the section that explains the mapping says so outright — which is that
self-correction made the contract instead of a wasted round. The folder is
dropped when the session has none, when it is borrowed, or when the ground or the
copy already holds it.

**Person-quoted paths are still verbatim.** `composeBrief` binds only the
model-authored sections and has never touched the quoted request; that is
unchanged, and a test asserts a person's own `put it in /s/work/flow.md please`
comes through untouched.

### Tests

- `TestBothHandoverAsksSayFinishedWorkIsNotWhatIsLeft` — both clauses word for
  word, and the handoff clause's position before the greeting instruction.
- `TestTheEvidenceSectionOutranksPartsThatNameACallAlreadyRun` — composes a brief
  whose parts name a read the evidence also names as done, asserts the precedence
  phrases render below the parts, and asserts no result body came with them.
- `TestTheConversationsOwnFolderIsWrittenAsThePathInsideTheCopy` — the binding,
  suffix preservation, lookalike siblings left alone, and every constructor drop.
- `TestTheBriefStatesTheConversationsOwnFolderWhereverItIsNamed` — the exception
  sentence, the bound sections, the person's sentence left verbatim, and no
  section for a borrowed session.
- `TestTheCopyTakesTheConversationsFolderFromTheTree` — owned binds, borrowed and
  the zero `Place` do not.

### Verification

- `go build ./...` — green.
- `make test PKGS=./internal/session TEST_FLAGS='-count=1'` — every test green.
  Two guard trips, neither a red in this change: the first run flagged the
  changelog file created while it ran, and the second flagged a branch and
  worktree the shared repository gained from a live session in the main checkout
  (the branch/worktree list is repository-wide, not per-worktree). Both left
  untouched.
- `make test-laws` — green, 18 packages.
- `make test-quick` — green.
- `TestAWindowOneStallClosedIsDecidedByThePosture/left_running_with_a_ceiling`
  fails under load. It is pre-existing: the same test prebuilt from this
  branch's parent fails at the same rate in alternating rounds. Filed as #803
  with a stranger-runnable replication. Nothing added to `.github/known-red.txt`.

### Manual

`internal/manual/chat/tasks.md` said a sketch naming work you have already done
was **not** covered; it is now covered as far as the ask can cover it, and the
page says which half of the brief wins and what genuinely remains uncovered (a
part that lands in the seconds after the account was taken).
`internal/manual/chat/how-tasks-run.md` said a path not under the project stands
exactly as written, with no exception; the conversation's own folder is now that
exception, and the page states it with the before-and-after address.


santoshkumarradha:
Landed with the other two ready waves on https://github.com/Agent-Field/aforge-v2/pull/807 rather than merging this into `dev` on its own. Spark tagged e2e is running on that combined tree.

santoshkumarradha:

PR #806, opened 2026-09-10
A room draws its own task's clock: news says which piece of work it is about

santoshkumarradha:
## What was wrong

Both news desks in `internal/tui3` — the phase desk and the lane desk — were keyed by **model**, which is an address and not an identity. Three things followed:

- two tasks running on one model overwrote each other's phase, and each room drew whichever of them had spoken last;
- a node's phase could never be drawn at all, because the only reading (`livePhase`) hard-dropped every role that was not `RoleTalk`, and there was no other reading;
- every rate on the frame was gated on `a.state != stateWorking` — the **conversation's** liveness — so a room showed no throughput whenever the conversation that launched the work was idle, which is the whole of a task's run: you hand the work out and your turn ends.

What a person read inside a task room, with the node writing: no provider, no tok/s, and a stale `running ask · 4m 55s` left over from the conversation.

`internal/session`'s `tellPhase` stamped model, role and `Session: a.newsKey()` — but `newsKey()` is the ROOT conversation, shared by every node in the family, so nothing on the news could tell two nodes apart.

## What is true now

**A news item belongs to a subject, and a window draws its own subject's news.**

- `provider.PhaseNews` and `session.LaneNews` both carry `Subject`. Empty means *the conversation*, which is what every existing producer and every older wire peer sends — so absence behaves exactly as it did before.
- `session.NewsSubject(conversation, node)` mints the one spelling (`<conversation>#<node>`), exported because both sides of the seam have to agree; the surface asks for it rather than rebuilding it. The conversation half is in the name because node ids restart per conversation and a guest room looks at another conversation's node 7.
- `provider.WithNode` is the context stamp beside `WithSession`, set once in the turn loop.
- One key function, `tui3.newsDeskKey(subject, model)`, keys both desks: the subject when there is one, the model when there is not. The lane **rings** stay keyed by model+lane, because those measure a machine and a subject in that key would split one machine's history.
- The window resolves its own subject. The conversation asks for the talk subject and **keeps** the `RoleTalk` requirement — the subject stops a node's news reaching that row, the role still stops the naming errands and memory reflexes that share the conversation's subject. A room asks for its node's subject and accepts a visible leaf role, because the subject already proves whose news it is. **There is no heuristic about a room being open**; two tasks on one model would show each other's rate under one.
- `windowPhase()` / `windowWorking()` are the one place "whose clock is on this frame" is answered. A room's own liveness is its subject's news being fresh, which the held and streaming beats make a real answer, and `roomPhase` has already refused anything past `phaseWindow`.
- The room's identity cluster now carries the node's own attribution — `⠋ Ship the parser fix · task glm-5.2 · via friendli` — on a four-rung width ladder, and the rate stands at the right edge as it does everywhere else. The ledger, meter and job counts on the same row are still the conversation's: those measure a session.
- The subject crosses the wire on both frames and through all four converters, and an older peer that sends none reads as the conversation.

## Tests

New, in `internal/tui3/newssubject_test.go`:

- `TestTwoSubjectsOnOneModelKeepTheirOwnNews` — two subjects on one model keep their own phase and their own served news, and a subject nobody said is still filed under its model.
- `TestARoomDrawsItsNodesRateWhileTheConversationIsIdle` — the room shows `38 tok/s` and `via friendli` with the conversation idle, says neither figure twice, and draws nothing once the news ages out.
- `TestARoomsPhaseIsItsNodesAndNotTheConversations` — `running go test` inside the room while the conversation's own `running ask` sits on the desk; and a room over a silent node draws nothing rather than borrowing that clock.
- `TestClosingTheRoomRestoresTheConversationsOwnLine` — `esc` gives the conversation its own rate back whole, and the node's phase never reached that row even while the room was shut.

`TestANodesOwnPhaseCrossesWithItsRoleOnIt` in `internal/remote` now asserts the subject crosses, survives the read back, and that an absent one reads as the conversation.

## Verification

`go build ./...`, `make test-quick`, `make test-laws`, `go test ./internal/e2e/` (untagged) and the four new tests are green.

`make test PKGS='./internal/tui3 ./internal/session ./internal/provider ./internal/remote' TEST_FLAGS='-count=1'` was first run against this branch's original base (`64dc75f28`): `provider` and `remote` green, `session`'s tests all green, and `tui3` red on four tests — `TestEnterConnectsOpenRouterInTheBrowserAndHandsTheKeyToThisProcess`, `TestAJobsLogReaderTakesOneLastReadingAfterTheJobEnds`, `TestAPathInTheBoxIsNeverChipped`, `TestAnAbsolutePathDoesNotHoldTheCommandListOpen`. Those were #798's harness clock and were fixed on `dev` by #801; this branch is rebased onto that, and all four pass here now. Nothing was added to `.github/known-red.txt`.

One thing worth knowing for anyone running `internal/session` on a laptop: its post-run checkout guard fired twice for me, naming worktrees under `~/.aforge/v3/projects/…/trees/N` that a **live aforge session** on this machine created and removed during the run. The guard filters out a neighbour's work by ignoring anything that is not on a `task/` branch — but the product's own task trees are exactly that, so a running aforge is indistinguishable from the suite writing into the tree. It fired once during a 16-second `make test-laws`, which cannot register a worktree at all.


santoshkumarradha:
Landed with the other two ready waves on https://github.com/Agent-Field/aforge-v2/pull/807 rather than merging this into `dev` on its own. Spark tagged e2e is running on that combined tree.

santoshkumarradha:

PR #808, opened 2026-09-10
Keep archived evidence readable while tasks finish

santoshkumarradha:
Context reduction replaces old tool results with file references, but the task finishing phase removed `read` while retaining `edit` and `write`. A worker could therefore be asked to save an accurate deliverable without any way to recover the evidence it had already read.

Keep the existing `read` tool during finishing and tell the worker to retrieve existing evidence as needed. Reading does not count as saving or reset progress rules. The existing stop decisions, permissions, execution budgets and exclusions for asynchronous generation remain in place. The manual describes the changed capability.

The regression test combines actual tool-result archival, successful retrieval, the finishing-tool transition and retrieval again. It uses a general text fixture and no model API, repository-specific instructions or benchmark paths.

Validation passed on Spark: Fleet420 reproduces the exact archived-read failure on the baseline, and Fleet424 passes the same test on the candidate, the full uncached session/manual suites, packed manual tests, and `make build`. All PR CI checks also pass. The ongoing matched benchmark403 is unchanged and does not include this patch. This draft claims a concrete correctness fix, not a measured quality/cost/speed improvement; keep it unmerged until validation and evaluation support it.

Related experiments: #770 removes mandatory progress notes; this fixes a different path that still exists with those notes removed. #763 changes context reduction and should preserve this retrieval contract.


santoshkumarradha:
Closed per owner request.

PR #831, opened 2026-09-10
Interpret completion claims in their original context

santoshkumarradha:
A completion report saying “no entries were deleted” could trigger an automatic repair because a quoted label still appeared in a file. In a captured benchmark, “all pass, none weakened or deleted” was treated as a claim that the quoted test command had been deleted, launching roughly six minutes of repair.

Remove the keyword-based claim classifiers and their automatic pass/fail decisions. The existing checker receives the original report once, acceptance, assembled files and check receipts, and interprets statements in that context. Explicit declarations in changed notes still arrive with source paths. Remove the duplicate report checklist and the post-check guess about which claims were examined based on shared words.

No extra classifier, model call stage, retry, generation limit or task-specific exception is introduced. Executable checks, restored-tree isolation, child-file assembly and dev's audit-window/second-attempt handling remain. The baseline snapshot preserves its existing 8,000-file bound under its own name. The manual and change entry describe the behavior.

Validation on Spark:
- Fleet433 reproduces the old false refutation with a general release-checklist fixture.
- Fleet435 validates current head `c8fe3da387b2abe7171bedd216b7cf62fefccf26`, integrated with dev `c0b20415129f580794d4ecd810baf4ec8de96e5b`: focused regressions; full uncached session (215.098s) and manual (1.309s) suites; build/vet, formatting, packed manual, structural laws, changelog and manual probes; `make build`.
- Build SHA256: `3d5ed752d5942849b6a2474bf9002d0697a0f8eb68eb79132aba9674d18e53ea`.

The regressions cover negated deletion, declarations interpreted in source context, negative checker findings, a report arriving exactly once and assembled child files. Prior433/434 source and proof are retained separately. No full or live tests ran on the laptop.

**Keep draft pending performance evidence.** A genuine contradiction now reaches the existing checker instead of the lexical fast path; that tradeoff must be measured along with avoided false repairs. Existing benchmark campaigns were interrupted by SIGTERM and remain frozen, with interrupted cells recorded separately. No new quality/cost/speed claim is made and this change is not in those runs.


santoshkumarradha:
Closed per owner request.

Issue #860, opened 2026-09-11
Two tui3 reds on dev: the greeting's first frame draws no spend, and a decayed belief still draws a tail


santoshkumarradha:
## What is wrong

Two tests in `internal/tui3` are red on a clean `dev` and nothing in
`.github/known-red.txt` names them. Both appeared between `f15f70b60` and
`155bc755d`.

```
--- FAIL: TestTheGreetingsFirstFrameCarriesTheSpend
    home_test.go:1818: the greeting read today 0 and the fortnight 3.25,
                       want 1.25 and 3.25 before the first beat
--- FAIL: TestADecayedBeliefDrawsNoTail
    lanewalk_test.go:237: a decayed belief has tail 212.71993655648765 (vague=false)
```

They are two different subjects and are filed together only because they arrived
together: the first is the home greeting's spend row reading zero for **today**
while the fortnight figure is right, and the second is a lane belief that has
decayed past the point of being worth drawing still producing a tail figure
instead of nothing (the emptiness law).

## Replication

A stranger can run this; it needs no key, no model and no fixture of mine.

```sh
git fetch origin dev
git worktree add --detach /tmp/af-devcheck origin/dev
go -C /tmp/af-devcheck test ./internal/tui3/ -count=1 -timeout 10m \
  -run 'TestTheGreetingsFirstFrameCarriesTheSpend|TestADecayedBeliefDrawsNoTail' -v
```

Both fail in under a tenth of a second at `origin/dev` = `155bc755d`.

## Acceptance

Each test passes on a clean `dev` with `-count=1`, for the reason the test
states rather than by moving the assertion — the greeting's first frame carrying
**today's** spend as well as the fortnight's, and a decayed belief drawing
nothing where it has nothing to say. Neither goes on `.github/known-red.txt`:
that ledger only shrinks.

## How it was met

Found by the recovery wave's R3 lane (#858) on a Spark `make pr-ready` run, whose
touched-package set was wider than the pull request's diff because that worktree's
`origin/dev` was older. R3's own pull request is green on all three CI jobs; these
two are `internal/tui3`'s and predate it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FSshDkszBQ9RpR4dPsC5fQ

santoshkumarradha:
Both fixed in #862 — https://github.com/Agent-Field/aforge-v2/pull/862

[...]

PR #863, opened 2026-09-11
The greeting's spend fixture dates today's row today

AbirAbbas:
## dev is red every night between 00:00 and 01:00

`TestTheGreetingsFirstFrameCarriesTheSpend` arrived with #842. It writes the spend row it means to be **today's** at `now.Add(-time.Hour)`:

```go
{At: now.Add(-time.Hour), USD: 1.25, ...},   // "today"
{At: now.AddDate(0, 0, -3), USD: 2.00, ...}, // three days ago
```

An hour before now is **yesterday** for the first hour of any day. So the greeting reads today as `0`, the assertion wants `1.25`, and the test fails — then passes again after 01:00, which is exactly the shape that gets written off as a flake.

Confirmed on a clean `origin/dev` at 00:32 local:

```
--- FAIL: TestTheGreetingsFirstFrameCarriesTheSpend
    home_test.go:1818: the greeting read today 0 and the fortnight 3.25,
                       want 1.25 and 3.25 before the first beat
```

## The fix

Date the row `now`. The current instant is the only time that is always inside today. One line, plus a comment saying why, so the next person does not reintroduce the hour.

`make test PKGS=./internal/tui3 TEST_FLAGS='-count=1 -timeout 15m'` — green.

## One pre-existing red this PR does not touch

`internal/tui3` also fails on **`TestADecayedBeliefDrawsNoTail`** (`lanewalk_test.go:237`), which fails identically on a clean `origin/dev@84bf71d61` with no change under it:

```
lanewalk_test.go:237: a decayed belief has tail 212.71993655648728 (vague=false)
```

It sets `old.At = time.Now().Add(-24 * time.Hour)` — the real wall clock — and then reads the view at `timeNow()`, the fixed test clock. Once real time drifts past that fixed date the belief is no longer 24 hours old relative to it, and the decay never fires. **Same family, different lane**, so it is left for its owner rather than bundled here; this PR is deliberately one commit.

## Why this is worth its own PR

It is offered separately from #861 so the nightly red stops whatever order things land in. #861 carries the same commit.

This is the third fixture on this work to fail with no change under it — #759's legacy cache aged past its TTL overnight, this one crosses midnight, and the decayed-belief test above mixes two clocks. **A test whose result depends on the wall clock reports the calendar.** A small structural law refusing wall-clock freshness in a fixture would catch all three; happy to write it if wanted.


AbirAbbas:
Superseded by #862, which landed on dev while this was open and fixes the same fixture more completely (it pins the whole homeLab clock to midday, so every fixture in the file is safe from midnight) along with TestADecayedBeliefDrawsNoTail (#860). Nothing to carry over.

PR #868, opened 2026-09-11
Every call can be watched: one seam for a call's whole life, fed from the readings it already has

santoshkumarradha:
A task room can draw a worker's call — thinking, tokens climbing, seconds since
it went out — and nothing else. A division, a sizing pass, a mark being read are
each one call through `callRole`, and every one of them draws a blank line while
it runs: measured between 12 and 219 seconds on 2026-09-10, which is
indistinguishable from a process that has stopped.

## What this adds

```go
func provider.WithCallProgress(ctx context.Context, watcher CallWatcher) context.Context
```

`CallProgress` carries the model, the machine that answered, which concurrent
request it is, when it went out, when the first token landed, the visible and
hidden counts, the phase (`started` · `paced` · `thinking` · `writing` ·
`ended`) and, on the last report, how it ended (`answered` · `cut` · `refused` ·
`cancelled`).

## Where it is fed from

Nowhere new. Every streamed reading in this process is already folded into the
hazard controller's watch (`armwatch.go`), where the tokens, the first token,
the heartbeats and the silences are counted for the waiting policy; and every
attempt's two ends are already written down by the model-call log's only writer
(`calllog.go`'s `Client.record`), under a law kept by a `defer` no path can
return past. This forwards what those two places know. There is no second
decoder and no second count of anything, and a `go/ast` law test fails the build
the day one appears.

A race reports each arm under its own `Attempt`; the arm that loses reports
`cancelled` from the race's own cancel site, which is the only place in the
process that knows the difference between exhaust and failure.

## The pacing park is a phase, not a second channel

`dispatch.go`'s three open-coded pacing sites collapse into one `park(bool)`
closure that flips the older `WithPacingNotice` bool and the `CallPaced` phase
together, so the two cannot disagree. `WithPacingNotice` survives only as the
door `internal/session` still calls; retiring it is one line there.

## Proof

- `TestAWatchedCallReportsGoingOutWritingAndAnswering` — scripted router, three
  router comment lines, a first word at 200 ms, forty tokens at forty a second:
  the watcher sees the request go out with no first token, the first token at the
  scripted moment and not at a heartbeat, a count that never goes backwards, the
  forty deltas coalesced, and `answered` carrying the final count and the machine.
- `TestAnAnswerThatStopsHalfwayIsReportedCut` — the connection goes away under
  the answer (new `lanestub.Profile.TearAfter`, the one ending the existing knobs
  could not stage) and it reads as `cut`, not as an answer and not as a cancel.
- `TestEachArmOfARaceIsWatchedAndTheLoserIsNotAFailure` — two arms, two
  `Attempt`s, one `answered` and one `cancelled`, and never both on one arm.
- `TestTheBeatHoldsAFastStreamToTenReportsASecond` — a hundred deltas over a
  scripted second draw ten reports.
- `TestTheCallProgressSeamHasOneFeeder` — the `go/ast` law above.

No person-facing string, key, command or tool changed, so the chat corpus is
untouched. `PERF.md` names no cap this changes.


santoshkumarradha:
Review answered at `6ddc0e1b6`. All six blocking findings and both non-blocking notes are addressed; the biggest one changed the design rather than a comment.

**#3 — `CallEnded` was per attempt.** The report is now opened **once per question**, at the door the question returns through (`completeWithMessagesStreaming`), and keeps each request's state apart under its own `Attempt`. An attempt's ending is LATCHED and never spoken; the ending is said once. A cancelled arm cannot become the question's ending while anything else could, so a race's loser no longer reads as the call being abandoned. Two new scenarios pin it: `TestACallThatWalksToAnotherMachineEndsOnce` (started twice, ended once, and the ending is the machine that answered) and `TestALatchedEndingIsNotAnEndingYet`.

**#4 — the law did not guard what it claimed, and this PR walked around it.** It now walks the SELECTOR CHAIN: `….progress.<anything>` outside `armwatch.go` fails the build. Naming the methods does not work — `opened` and `paced` are both spelled on other types here — and the bypass from `lostRace` is gone entirely: a cancel is classified from the log's own closing word like every other ending, so there is one spelling and no special case.

**#1 — the beat.** 33 ms, and it is `internal/tui3`'s `frameInterval`, with `remoteFrameInterval` named as the coarser link stride it is deliberately not. The derivation is in the doc comment and in PERF.md's bounds table beside `lane.SpokenWithin`.

**#2** — `paced` returns when the report is closed, staged in `TestALatchedEndingIsNotAnEndingYet` (the dispatcher really does take the park back after the closing row). **#5** — change entry added, with **#877** filed as the condition on leaving `WithPacingNotice` standing for one wave. **#6** — the door's doc now says *every streamed call that runs under a waiting controller* and names the gate.

Non-blocking: `TearAfter` counts deltas across both shapes of answer, so `Answer` + `TearAfter` no longer silently does nothing; the race scenario orders itself on `StallUntil` rather than on two wall-clock figures.

Proof on Spark (no suite run on the laptop), `09e3c0b62`: `internal/provider` 35.2s, `internal/lane` 6.9s, `lane/control` 0.2s, `lane/lanestub` 0.6s, `internal/guard` 0.1s — all ok, plus `make test-quick` green there. `internal/provider` has 42 loose lock sites on `origin/dev` and 42 here, so nothing was added to the known-red lock ledger.

No exported name changed; `CallEnded`'s contract did (once per question, not once per attempt), which is what a room drawing one line wants.

santoshkumarradha:
Blocker closed at `87250378d`.

You were right that it is reachable, and the fix is the latch rather than three guards: `finished()` now **spends** the report — whether or not anything was ever opened — and `opened`, `note`, `serving`, `paced` and `landed` all return on it. Spending it even when nothing opened matters for the same reason: a question that reached no machine at all must still close the seam, or a straggler from a cancelled arm opens one behind it that nothing will ever close.

`TestAStragglerFromACancelledArmCannotSpeakAfterTheEnding` stages exactly the path you described — note, serving, opened, paced, landed and a second finished, all from a loser's arm after the question has settled — and asserts the report count does not move, the last word stands, and there is exactly one ending. Removing the `spent` check from `note` alone fails it (`a straggler spoke 1 times after the ending`), so it is a real guard and not a restatement.

Both non-blocking notes done: `finished()` says out loud that it draws its row from attempt 0 when nothing ever landed — the honest picture of a question that stopped without anybody coming back — and `say()`'s comment now names **the lock** as what keeps the order, with the law scoped to where the seam may be fed from.

PR #876, opened 2026-09-11
No hidden work on the turn: one file memo, one off-path reading, one deferred write, and one door for a hand asking a model

santoshkumarradha:
Eight measured places where a person's turn paid for work it never asked for. The trap was eight local patches; these are **three shapes**, each one named type with its law in its doc comment, one test, and every caller using it.

## The three mechanisms

**`internal/filememo` — a file read at most once per change.**
`Memo[T]` spends one stat to decide whether it may answer. Used by the profile's `config.json` (`APIKeyConfigured` on every Enter, `FirstPrompt` on every turn, eight readers in `internal/config`) and by a project's `AGENTS.md` / `CLAUDE.md` (re-read whenever the prompt's clock goes stale — the came-back-to-the-terminal turn, and once per attached folder). `Stamped` folds a caller's own write counter into the freshness key, so `internal/config` can never serve back a write it just made, whatever a filesystem's timestamp resolution is.

**`internal/offpath` — the two shapes of work that must not happen on a person's path.**
`Take`/`Reading[T]` is a fact gathered beside the work and settled with a bound; `Deferred`/`Write` is a write a read path owes, coalesced, with `Settle` as the exit door. Three users today: the per-batch `git status`, the fix shelf's counters, the meta.json stamp. **The peer lane's S1 (per-tool-call status on `speed/task-start`) is the fourth** — `offpath.Take` is exactly its shape and the package is named and exported for that.

**`internal/session/toolask.go` — the one door a hand on the belt asks a model through.**
`view_image` and `read`'s three senses each made their own call, and each got the same four things wrong separately.

## What each finding cost, and what it costs now

| | before | now |
|---|---|---|
| `read` on media / `view_image` | untagged in the call log (`tag:null`, 2,830 unattributable finishes in ten days), silent for 7–20 s, unbounded in length, bounded at ten minutes | tagged `tool:<name>` + node, `looking at error.png` on screen for the wait, capped at what a tool result shows, bounded by `lane.RoleTool.GiveUp()` |
| `git status` per batch | inline, 10–30 ms small / seconds large | `offpath.Take`, joined at `lane.Hysteresis` |
| `compactToolHistory` | rebuilt the frozen prefix per request | **1.34 ms → 13 µs** (100 rounds), **11.7 ms → 156 µs** (1600) |
| `config.json` | read + parsed per call | one stat per call |
| `AGENTS.md`/`CLAUDE.md` | re-read per stale clock | one stat per render |
| `fixShelf.consult` | 4 whole-file cycles per fail→fix→succeed, on the READ path | zero writes on the path |
| `stampUserLocked` | LoadMeta+SaveMeta **with the agent lock held**, per user message | snapshot on the path, disk behind it |
| first write into a referred folder | `git worktree add` / recursive copy inside the tool call, silent | started at `ReferPlace`, `preparing · a working copy of <folder>` |

## The `read`-on-an-image decision

**The sense stays; the defect was never that `read` senses.** Decision 20's whole argument is that a model must not have to pick between two verbs for one intention, and `read`'s description has told it `Images, audio and video come back described, never as bytes` since that decision landed — so removing the sense would be removing a working capability, which is not what "a capability that cannot work is absent" says. What was actually wrong: the call was anonymous, silent, unbounded in length and carrying a second ten-minute budget beside the one the dispatcher already derives. All four are fixed at the door. `prompts/system.md` now says `read` **asks another model** on a picture and that it is a real, bounded, shown wait that is free the second time; the manual gained "How long does reading a picture or a recording take, and what do I see while it happens?".

## New role row

`lane.RoleTool` — the properties are derived, not described: somebody **is** waiting (the tool row is open in front of them → `Interactive`), its text is never drawn (`Visible: false`), it is text and is watched as text (`Streams: true`), one shot (`Horizon: 1`), as impatient as talk (`Patience: 1` → 90 s give-up, 10 s ceiling). `viewLookWindow` is deleted.

## Laws now guarding this

- `TestNoHandAsksAModelOutsideTheOneDoor` — `go/ast` over `tools_*.go`, fails the build on a second send site. On the `test-laws` gate.
- `TestAReadOnlyBatchRunsNoWorktreeReading` / `TestAShellBatchTakesOneReadingAndDoesNotWaitItOut` / `TestAFastWorktreeReadingStillReportsMovement`
- `TestConsultingTheFixStoreWritesNothingOnThePath`
- `TestAFrozenPrefixIsReducedOncePerConversationAndNotPerRequest` + `BenchmarkCompactToolHistory` (cold/warm)
- `TestTheProfileConfigIsParsedOncePerChange`, `TestAWriteDoesNotRewriteWhatTheMemoIsHolding`
- `TestAToolsAskIsBoundedByItsRoleAndNotByANumber`, `TestAnAskedModelMayNotWriteMoreThanTheBeltShows`
- `internal/offpath` and `internal/filememo` each carry their law as their own test file.

## Deleted

`viewLookWindow`; `senseOneShot`'s own completion, accounting and error handling; `viewImage`'s own context timeout, role, accounting and no-answer handling; the inline `worktreeDirt` call in `treeMoved`; `saveLocked` from `consult`; the synchronous `updateMeta` in `stampUserLocked`; the per-request view composition in `compactToolHistory`'s two walks; `readProfileConfig`'s and `readInstructionFileWithin`'s own reads; and the comment in `budget.go` saying this package deliberately holds no cache.

## Seams into other lanes' files

- **`loop.go:1379`** — one token, `compactToolHistory(` → `a.compactToolHistory(`. The memo has to live on the agent; there is no other per-session home for it.
- **`internal/lane/roles.go`** — additive: one `Role` constant and one table row.
- **`agent.go`'s `Agent.Close`** *(not made — L1's file)*: it should call `a.SettleMeta()`. Without it, the window in which a process could exit holding an owed meta.json stamp is the length of one small write; with it, zero.


santoshkumarradha:
**Green on all three jobs**, and two reds were found on the Spark and fixed before they reached here.

### Corrections made during the lane's own review (all in this PR)

- **The output ceiling was drafted and deleted.** The first draft sent `ai.WithMaxTokens(resultCaps().MaxBytes / bytesPerToken)`. `internal/lane/frontier.go`'s `capable` removes a machine from the serving set when its published `MaxOut` is below the request's `max_tokens` — and the belt's cap is ~12,500 tokens against measured answers of 588–1,256, so the ceiling would never bind and would only strike vision machines off the set of the tool this wave exists to speed up. It is now a **law** (`TestAHandSendsNoOutputCeiling`, `go/ast`) with the reasoning in `toolAskSendsNoOutputCeiling`, and the manual says the same thing to a person.
- **`RoleTool.Streams` was drafted false and is true.** `Streams` is not "was this request streamed" — every errand here is sent unstreamed — it is "does this role produce a token stream at all", and the token controller's one exclusion is media (`waiting_law_test.go` enforces exactly that). What is false of a hand's ask is `Visible`.
- **`worktreeDirtJoin` was drafted as `lane.SpokenWithin / 20`** — a ratio, which is the elegance bar's own example of what not to write. It is `lane.Hysteresis`, the quantity this build already defines as the smallest difference in waiting anybody notices, with no free constant.
- **The ahead-of-time working copy was a bare goroutine and is now the same `offpath.Write` as every other deferred write here** — one pass that cuts every referred folder with no copy yet, coalesced across folders (three folders dropped at once cost one goroutine, not three contending on `a.treeCut`), drained by `Agent.SettleWrites`. It was outliving two tests' tempdirs; fixing it removed the last bare goroutine in the change.

### Spark evidence

`/tmp/af-rec-l6 @ d4ff9438d` — `go test ./internal/session/ -count=1` → **ok 177.950s**.
At the identical code plus those fixes (`0183c9ad1`): `config` ok 5.8s · `provider` ok 35.7s · `offpath` ok 0.04s · `filememo` ok 0.04s · `manual/...` ok 2.3s · `tui3` ok 296.6s · `lane` ok 7.0s and 6.9s (twice on the branch), ok 6.9s on `dev`.

`internal/lane`'s `TestS2TheDefaultGoesSlowAndTheRouterMoves` failed **once**, while `tui3` and `session` were running beside it on the same box — `routed p90 4.199s against the pin's 4.465s, 6.0% better, gate 30%`. It is a statistical gate under load, passes three times otherwise including on `dev`, is not touched by anything here, and is **not** added to `known-red.txt`.

### One seam still owed

`agent.go`'s `Agent.Close` should call `a.SettleWrites()` — one line; the method is written, exported, and already used by `newTestAgent`. Until it is there, the window in which a process could exit holding an owed write is the length of that one write rather than zero. I did not edit `agent.go`.

PR #880, opened 2026-09-11
Task start B: a proposal's card goes up while its reply is still streaming (staged tools)

santoshkumarradha:
**Lane B of the task-start wave.** `propose_task` is now a staged tool. When its call closes while the rest of the reply is still streaming, the card and its countdown can go up at once. Admission waits until the reply is whole, and the proposal is withdrawn cleanly if the reply does not go through.

## Measured before building
The Spark's 5,562 transcripts contain ten replies with two or more task calls. The stream time left after the first task call closes:

| task calls | est. from wall-time share | est. at 100 tok/s |
|---|---|---|
| 2 (n=8) | p50 11.3 s, p90 60.8 s | p50 4.6 s, p90 12.5 s |
| 3 (n=1) | 13.2 s | 6.1 s |
| 4 (n=1) | 18.7 s | 29.7 s |

The gain for each proposal is capped by its countdown (15 s by default). The work now starts at the later of the countdown ending and the reply ending; before, it started at the reply ending plus the countdown. Method and data: `docs/design/task-start/DESIGN.md`.

## What this PR does
- **`internal/exec/bare/stage.go`**: `StagedTool`, `Staged` (`Commit` or `Withdraw`, exactly one), `Hold` (decided once: release or withdraw), `RunStaged`, and `Tool.Stages()`. The property that lets a tool start early is the tool's shape: `Stages()` reads an unexported field that only `StagedTool` sets, so no tool can claim it with a flag.
- **`propose_task` split** (`task.go`):
  - `stageTask`: refusals, ground, slot, id, preflight, and the card with its clock running.
  - `stagedProposal.Commit`: reads the answer, applies the redirect and the model shortlist, compiles `admissionContext`, admits.
  - Two things moved to keep quality unchanged. The **admission context is now compiled at the commit**, so the node's brief quotes the whole reply, byte-for-byte what the batch compiled. The **answer wait runs from the moment the card goes up**, so a "no" and an expired clock are read in the order they happened.
- **Withdrawal**: the question is withdrawn in one sentence, and the card settles through the new `TaskNotice.Withdrawn` field (`internal/tui3/task.go`: a few lines in lane V's file; V has been told). Late answers are dropped and the fan slot is handed back.
- `quick_task` is **deliberately not staged**. Its commit is the worker's first request, and that request's brief quotes the reply, so nothing about it can run early without changing the brief.

## The loop.go seam (not in this PR)
`loop.go` belongs to another session's lane, so this PR ships only the machinery; without the seam it changes nothing about when a proposal starts. The five-move seam, its loop-level tests (`stageearly_loop_test.go`) and the manual lines that describe the behaviour are one commit on **`speed/ts-B-seam`** (`1eb37f681`), stacked on this PR's commit. The design doc lists every move. Apply that commit after this PR merges, with the owner of loop.go agreeing.

## Tests
- `internal/exec/bare/stage_test.go`: the hold's laws.
- `internal/session/stageearly_test.go`:
  - the card goes up and the clock runs out, yet nothing is admitted before release;
  - the admission quotes the reply;
  - withdrawal takes the card and the question back;
  - withdrawal hands the fan slot back;
  - a "no" given before the clock expired wins;
  - on the belt, only `propose_task` stages.
- `internal/tui3/taskwithdrawn_test.go`: a withdrawn card settles, overrides this window's own clock verdict, and draws nothing when this window never had the card.
- Seam branch: `stageearly_loop_test.go` covers four cases:
  - the first card is up before the second call closes;
  - a cut reply takes back the proposal it had started, and exactly one task is admitted;
  - a reply that ends without the call takes it back;
  - a write before a proposal keeps its order.

  All pass under `-race -count=3`.


santoshkumarradha:
Follow-up on the review, all pushed:

- **Suite:** `make test-touched BASE=origin/speed/task-start` at `11c0f7433` **PASS** (integrator's serial run, 11:29–11:39 EDT; session 174s, tui3 297s, thirteen packages).
- **Structural law added** (`f2568d085`): `TestOnlyStagedToolSetsTheStagedField` walks `internal/exec/bare` with `go/ast` and allows exactly one writer of the unexported `staged` field, `StagedTool`. It is on `make test-laws`, and it was mutation-checked (a second constructor setting the field fails it).
- **A race fixed** (`cfe53a069`), found while rebasing the seam: `ResolveTask` takes a proposal off `a.taskAnswers` and then sends the answer, so a countdown firing in that window left `select` to pick between the answer and the clock — a "no" given before the deadline could be overruled by it. Three failures in ten `-race` runs of this wave's consent-order test. `taskSilenceAdmits` now refuses to approve a proposal that is no longer registered, under the lock the answer was claimed with. Green at `-race -count=15`. **This is older than the staged proposal** and reachable on `dev` today.
- The two commits above land after the PASS sha, so a re-run at `cfe53a069` is worth having; both are confined to packages that run already covered.
- **Seam** (`speed/ts-B-seam`, now `9936ce4a9`): rebased onto `origin/dev@af5aa9950`, loop.go merged with no conflict, light gates and the seam tests green at `-race -count=6`. No PR opened for it, as instructed.
- **Filed #908** for the pre-existing tui3 defect this wave makes visible: with two proposals in one reply, the first card is settled `expired · the turn ended` while the engine is still asking about it, and the word is never corrected — a task that runs keeps a card reading `expired`. Reproduced on plain `dev`.

Issue #830, opened 2026-09-10
Groom: TUI question component UX — no navigation between questions, long pause after final question, awkward up/down arrow behavior


santoshkumarradha:
## Summary

The question component (the TUI surface where the agent asks the person questions during a task) has several UX problems that need grooming before any implementation work starts.

## Observed problems

1. **No way to move between questions** — once a question is shown there is no way to go back to a previous question or jump forward. Answering wrong means living with it or restarting the whole flow.
2. **Long pause after the final question** — after answering the last question, the TUI sits for a noticeable period with no visible feedback before anything happens. Feels frozen; we should show what is happening (or fix the latency itself).
3. **Up/down arrow behavior is weird** — arrow-key navigation doesn't behave the way a person expects (selection highlight, wrap-around, or focus semantics are off). Needs a defined key contract.

## What this issue is

A **grooming** item, not a task: the expected behavior for each of the three problems has to be decided (navigation model, post-final-question feedback, arrow-key contract) before anyone breaks it into tasks.

## Grooming questions to settle

- Should people be able to navigate back and edit earlier answers? If yes, what happens to downstream answers?
- What should happen between the final answer and the next visible state — spinner, progress, or eliminating the pause?
- What is the intended key contract for the question list (up/down, tab, vim keys)? Should it match the rest of the TUI conventions?

## Notes

- Reported by hand during use of the TUI; no replication steps beyond "answer questions in a chat/task flow."
- Area: `internal/tui3` question component.

santoshkumarradha:
## Recon findings (2026-09-11, dev @ 39039b990)

Four read-only passes: engine, rendering + keys, freeze/latency, design contract. Nothing changed yet. This issue tracks the redesign + fixes; close it with the PR.

### Defects — cause first

[...]

PR #914, opened 2026-09-11
questions: an answer is a message — keep working while asking, a countdown you can stop, and undo

santoshkumarradha:
Lane A of the question-views wave (#830), **against `dev`** and standing on its own
there. An answer to the model's own `ask` was only the RETURN VALUE of the call
that asked; it is now a MESSAGE TO THE CONVERSATION that also releases a waiting
call when there is one. Three things the owner picked fall out of that one
mechanism, and a fourth — the undo — rides the same door.

## The mechanism

`internal/session/asklane.go` — the lane's book (`Agent.asks`, one `modelAsk` per
question the model raised), one rendering (`askAnswerText`: a sentence, then the
whole `Answer`), and two deliveries: the call parked on it, or the conversation's
own queue through **`Agent.accept`** — the road a job's ending, a watch's firing
and a landed task already take (step boundary, or a wake on an idle
conversation). No second delivery path. `ResolveQuestion` is still the one door.

**It absorbed #870's book.** `internal/session/askwait.go` held the parked wait
with three endings and a doc warning that two maps holding one question is a
defect; this is one map with those three endings kept word for word plus the
fourth — a call that never waited. `THE ENTRY IS THE OWNERSHIP` became `THE CLAIM
IS THE OWNERSHIP` (`modelAsk.claimParkedLocked`), because a settled entry is kept
so its answer can be changed. Talking past a question still retires the asks a
person's sentence is in the way of — the ones something waits on — and leaves a
non-blocking question standing, which is what asking without waiting means.

**Deleted:** `askWaits`, `askedOfThePerson`, the policy clock inside
`executeAsk`'s `select`, the ask arm that dropped an answer when nobody was
parked, and the refusal in `SetAutonomy` that made a clock length compulsory.

## What a person gets

- **Keep working while asking.** `blocking: {turn: false}` returns at once; the
  question outlives the call and the turn. Omitted or `true` waits exactly as
  before — the zero value waits.
- **A countdown you can stop.** The clock belongs to the lane, not the call, so
  it runs on a question nobody is parked on; it takes the pick through the same
  door, and what the model reads says the answer is **provisional**. Opt-in: no
  clock unless the person set one (`/autonomy`), length by one derivation
  (`autonomyClock`). Any key stops it for good — `Agent.HoldQuestion`, wire
  `Question.Hold`, sent from a `tea.Cmd` — and the question says so itself while
  the clock runs: `any key stops the clock · you can still change the answer
  afterwards`.
- **Change your answer.** `c change` on the receipt asks the question again; the
  new answer carries `Answer.Revises`, the record says `changed from …`, and the
  model is told what changed. Irreversible decisions refuse it in words.
- **Undo a permission.** `u undo` on the receipt of an `always` hands that grant
  back whole — the session memo, the saved settings line, and a connected
  account's capability, which returns to asking rather than to a no nobody said.
  `Agent.undoGrant` is the one door for the stores the engine owns;
  `app.forgetAlways` is the surface's own line. What may be changed is ONE
  reading (`Question.Revisable`), asked by the receipt that offers the key and by
  the door that takes it.

Wire **15** (both halves fail as silence on a v14 engine). Lean prefix 30,833 →
30,968 bytes against a 31,500 budget; fixed prefix unchanged.

## Proof

`make pr-ready BASE=origin/dev` green — laws, and `internal/session`,
`internal/tui3`, `internal/manual`, `internal/e2e`, `internal/remote` with
`-count=1`.

Laws added: a non-blocking ask returns at once and its question stands; a parked
and a messaged answer are the same bytes; a provisional pick says so and runs with
nobody parked; a hold clears the deadline in the engine; a change reaches the model
as a correction (parked and unparked); an irreversible decision refuses a change;
an answer the lane no longer holds is refused in words; an answered question
leaves the desk; a question nothing waits on is not `waiting on you`; a clock that
runs out after `Close` decides nothing; an answer a closed conversation cannot read
records nothing; the second answer to one question records nothing; the countdown
says how to stop it and only while it runs; `u undo` takes the standing yes back
and only a standing yes is offered one.

Driven in tmux as a person against `deepseek/deepseek-v4-flash`, twice — once on
`--no-host` and once over the DEFAULT host road (#690 is closed by #856, and the
idle-woken answer reached the screen). The model chose the non-blocking form,
finished the other files while the question stood, and used the answer when it
came; a change from the receipt made it rewrite the file; the clock took its pick,
said so as provisional, and the model asked the person whether to keep it; a key
press left the next countdown `paused` and undecided a minute later; `u undo`
printed `taken back · this tool asks again` and the next call asked again.

Two earlier drives found two defects the unit tests had not — a correction posted
into a spent channel, and a reopened question that still said `waiting` — both
fixed and pinned.


santoshkumarradha:
### Seams for the other lanes

This PR leaves five hooks that are somebody else's file. Each is small and named here so it
is not discovered in a merge.

- **P — the saved permission rule.** A revised `always` drops the ENGINE's memo
  (`Agent.reviseConsent`), but the rule written into the person's settings is the surface's.
  Call the inverse of `internal/tui3/consent.go`'s `app.rememberAlways(tool, shape)` from
  `app.consentAnswered` when `answer.Revises` is set and the key is not the widening one.
- **P — the offer to stop asking.** `session.AutonomyOffer(records, rules)` is the WHEN: it
  answers the `AskKind` whose third same-shaped, same-way, person-made answer has just been
  given, and false otherwise. An explicit row in `autonomy.json` silences it for good, so it
  is offered once per kind per project — taking it AND declining it both write that row,
  through `Agent.SetAutonomy`, the one write door. The receipt's dim line and the `y`/`s`
  keys are P's.
- **P — how long a receipt offers `c change`.** The key lives only while the receipt is drawn
  (`questionRecordFor`, ~30s, `questionRecordsShown`). A provisional pick somebody comes back
  to a minute later therefore has no key. If the pick means "for as long as the receipt
  shows", a receipt for a decision the person did NOT make may want to stand longer.
- **P — `paused`.** The engine says a held clock by clearing `Question.Deadline` while
  `Policy.Kind` stays recommend-then-auto; `app.questionClockWord` reads exactly that and
  returns `paused`. `Question.Deadline` is the one source of truth for the deadline.
- **F — the surface's second dial.** `internal/tui3/questiondelivery.go` still answers a
  recommend-then-auto question from the surface when the person is away, beside the engine's
  clock. First-answer-wins makes it harmless, but it is two clocks for one law and the
  engine's is the one that works with no window attached. F's file; not touched here.
- **F — reconcile on reattach.** A session host that restarts loses the ask book (stated as a
  limit in the report). The ghost row on a still-attached window closes honestly only if the
  surface withdraws what the replay no longer lists.
- **One line, somebody else's file, not taken:** `internal/tui3/task.go:1549` calls
  `HoldTask` directly; `Agent.HoldQuestion(kind, token)` is the door it should become.

Files of another lane this PR does touch, and how much: `internal/tui3/question.go` (~40
lines: the `revising` mark and the `Revises` stamp, the `held` hook in `raiseQuestion`, the
`paused` word, `questionRecord.question`, `Was` on the receipt), `internal/tui3/app.go` (two
fields and the hold drain in `Update`), `internal/tui3/autonomysheet.go` (`recommend` with no
duration), `internal/tui3/question_test.go` (one field on the fake agent),
`internal/remote/{wire,server,client,callclass,questionlane}.go` (wire 15 and the hold door).

Issue #898, opened 2026-09-11
The frame law is blind to a package function that takes *app as a parameter


santoshkumarradha:
## What is true today

`internal/tui3/framedisk_law_test.go` walks everything `app.View` can reach and
fails on a call that reads a file, starts a process, globs a directory or
touches the network. It landed in #875 and it works — it caught two live reads
nobody knew about (`.loadNoticeLedger`, `.readJobLogTail`, both deleted) and it
catches all three of the defects that lane was opened for when they are planted
back.

But its walk makes a surface edge out of `x.method()` **only where `x` is
spelled the same as the enclosing method's own receiver name**
(`surfaceGraph.callsIn`). A package function that takes the surface as a
PARAMETER therefore severs the graph at its own door: the function's body is
walked, because a bare call is resolved to the package function of that name,
and then every `a.…` call inside it is read as "a method on some other value the
frame happens to hold" and dropped.

The reviewer of #875 traced a live one:

```
app.View → app.placeDraw → placeFrameWithBar(a, …)   ⊘
  a.composerRows → a.composerWhereLine → a.composerWhere
    → a.composerOpensAt → a.errandPlace → errandHomeDir → os.Getwd
```

`os.Getwd` is in that law's own forbidden set as of #875, and the law cannot see
this call. `pages.go:1130` is the cut.

This is written down as the fifth blind spot in the file's head comment
(`d8cd923b1`) because the comment opens by saying that a law whose reach a
reader cannot predict is a law that lies by omission. Writing it down is not
closing it.

## What to do

Make `callsIn` parameter-aware: alongside `receiverName(fn)`, collect the names
of the function's parameters whose type is `*app`, and treat a call on any of
them as the same surface edge. It is about eight lines off `fn.Type.Params`, and
the same names feed `nodeName(recv, …)` with `"app"` as the receiver type.

`internal/tui3` has a lot of `func f(a *app, …)`, so **expect this to turn things
up** — that is the point, and it is why it was not done in the same commit as the
law itself. Each one is then a real question with three honest answers: the read
moves to `open`/`tick` behind the memo (`learned[T]`), or the call is deleted, or
it goes on `frameDiskDoors` with the line saying why it is once-per-epoch rather
than per-frame. The allowlist is one entry today and should stay short.

`TestTheLawSeesEveryShapeOfIndirectionItClaims` is where the proof goes: plant
`func planted(a *app)` calling a forbidden thing and assert the walk reaches it,
[...]

PR #918, opened 2026-09-11
The leaf's grant is one number, and the measurement saying it is too small

santoshkumarradha:
## What was true

`defaultLeafTokens` was spelled in five places: the constant, `aforge exec`'s `--token-budget` default, `aforge run`'s, `chatLeafTokens` in the chat surface — whose comment promised it *"mirrors the headless run defaults exactly"* — and again inside `cmd/aforge`'s usage test. That is an intention where this repository's one-source-of-truth law wants an interpolation, and it meant five numbers had to be recalibrated together by somebody who knew all five were there.

## What is true now

The grant is `exec.DefaultLeafTokens` and every door reads it from there. `TestTheLeafGrantIsSpelledOnce` reads declarations with `go/parser` — not a grep, so a figure inside a string or a comment is not a false positive — and fails the build on a second spelling anywhere under `cmd/aforge`, `internal/exec` or `internal/session`. Proven to bite: planting `const plantedGrant = 250_000` turns it red naming the file and line.

**The value is unchanged at 150,000**, and that is the interesting part.

## The measurement

Issue #898 (the frame law widened to see `func f(a *app)`) was run headless on `z-ai/glm-5.3`, run `04c2404b26072e41`, 2026-09-11. Neither half of the calibration held: the median call carried **14.2k** prompt tokens, not 11k, and the leaves that did the work ran **14 to 41** turns, not 8 to 16. Every one was landed mid-edit.

| leaf | turns | spent | of grant |
| --- | --- | --- | --- |
| Issue 898 frame law fix | 14 | 184,411 | 150,000 |
| finish-issue-898 | 41 | 190,524 | 164,462 |
| Answer remaining offenders | 14 | 268,971 | 211,852 |

A grant above the constant is the constant plus `gatheringGrant`'s dependency term; the overshoot past each is the landing reserve at `landingTokenShare`, working as designed. The cost was the re-planning around every landing: **five rounds, seven nodes, 43 minutes, $2.11**, and a delivery gate that refused at the end for want of time — while the work itself was correct and committed after the first two leaves. Every split is paid for twice, once in a fresh planning round and once in the context the next leaf has to be told again.

## Why the number did not move

This branch opened by raising it to 250,000. That turns three tests red, and none of them should be moved to let a figure through:

1. **`TestObservationWindowIsSizedFromContextNotSpend`** compares the 32,768-byte context-derived window against `DefaultLeafTokens/6`, so the grant may not reach **196,608** until the window is re-derived. Moving the comparison would un-pin the category error the test exists for.
2. **`TestACacheDiscountedRunawayLandsOnItsMoney`** holds a 98%-cached runaway under the 1.4M raw tokens the audited melt-downs reached. At about **5.8× the grant**, 250,000 pushed 1,442,112 — exactly what the grant forestalls.
3. That test's closing cost/raw separation **coincides at 190,000**, where the simulated bound and the real run both land at 85 turns, and the test says so itself.

So the grant is load-bearing rather than a knob: moving it means moving the observation window's derivation and re-measuring the runaway bound. **#920** carries that work and this evidence. Until then a lane that needs a bigger grant for one run passes `--token-budget` to `aforge exec` or `aforge run`.

The constant's own comment and `PERF.md`'s "A leaf's bounds" both carry the reading, so the next person to reach for the number meets the ceiling before they move it.

## Checks

`go build ./...`, `gofmt`, `go vet` clean. `internal/exec` and `cmd/aforge` suites on the Spark.


santoshkumarradha:
Opus review findings applied, and the Spark acceptance run is green.

```
ok  github.com/Agent-Field/aforge-v2/internal/exec   57.939s
ok  github.com/Agent-Field/aforge-v2/cmd/aforge      61.007s
```
(spark, worktree at origin/perf/leaf-grant-measured @ 6594b34b8, `go build ./... && go vet && go test -count=1`)

What changed since the review:

- The law reads `*ast.BasicLit` rather than `*ast.ValueSpec`. Three of the four copies it exists to catch were call arguments, and a flag default is exactly where the figure gets written a second time. Proven to bite by planting `f.Int("token-budget", 150000, "")` in `cmd/aforge`, which it named by file and line.
- Scoped by `leafGrantWatched` so `internal/exec`'s own thirteen fixtures, which choose a size rather than restate a default, are not policed while the doors are.
- **New: `TestTheManualSaysTheGrantTheLoopApplies`.** `internal/manual/chat/adaptive-runs.md` tells a person in words that a run stops at 150,000 tokens. That is the fifth spelling and the only one a person reads without opening the source, and markdown has no integer literals for an AST law to find. The page must carry the current figure or the build fails, so a recalibration is two edits and the build names the second.
- The change entry no longer claims more than the law delivers. It now states the blind spots plainly: the package's own fixtures, a hex or differently-grouped figure, and a stale copy of a *past* figure, which a recalibration still owes a hand sweep for.
- The 250,000 probe was re-run rather than recalled. It turns **two** invariant tests red on **three** separate bounds, not three tests. The entry says so.

The grant is held at 150,000. #920 carries the recalibration.

Issue #891, opened 2026-09-11
Three seams #873 measured and did not take: the mid-stream rate test asks the wrong machine, a late first token teaches only a losing arm, and a refusal that cannot name an unknown machine


santoshkumarradha:
Three things #873 measured and deliberately did not take, because each crosses a file that lane was scoped away from or wants a ruling rather than a patch. They are written down here so they do not live only in a report.

#873 (`dev@834bdce26`) made the chooser refuse a machine whose expected time to a usable answer is longer than the role's own declared deadline, and made that refusal reach `provider.ignore` rather than `provider.order`. **All three below are about the half of the same question that lands AFTER the request has gone out, or about the one policy decision that is not ours to make.**

---

### What happened

2026-09-11 10:02, `dev@c9f24f2b6`, `tag:task` node 7. After a bash `cd …` tool result the next step was served by one machine at roughly 6–8 tokens a second, and ran **219,014 ms** — three and a half minutes during which the person watched a `cd` row and read it as stuck. Two machines on the same model were writing at 90–120 tok/s. From the same call log, that machine's own rate over the previous night:

```
01:04–01:11   45 rows at 40–105 tok/s (median ~62)
09:33:33      9.8 tok/s
09:37:16      5.9 tok/s   (197 s stream)
09:52:49      6.1 tok/s   (93 s)
09:59:46      7.9 tok/s   (543 s, 4,307 tokens)
10:06:18      6.8 tok/s   (219 s)
```

The stream was producing tokens throughout, so `#853`'s "a producing stream is never cut" correctly held. What did not happen is a move to one of the machines that were forty times faster.

### Why

`hazard.keepingUp` (`internal/lane/control/hazard.go`) judges the live per-token gap against `h.plan.Gap` — **the belief about the machine that is answering**:

```go
func (h *hazard) keepingUp() bool {
	return !h.plan.Gap.Known() || h.weight < 1 || h.drift/math.Sqrt(h.weight) <= h.z
}
```

Six tokens a second is not surprising for a machine believed to write at six tokens a second. `rateCollapsed` is therefore false, the ceiling runs on the ordinary clock, and nothing is done. The test can only ever catch a machine getting worse than ITSELF; it cannot see a machine that is simply far worse than what the request could have had.

(#873 closed the half of this that happens before the send: a machine believed that slow is no longer chosen for a role somebody watches. This is the half that matters once a healthy-looking machine degrades mid-stream, and the series above is five consecutive steps of exactly that.)

[...]

PR #925, opened 2026-09-11
A named refusal is a move: a veto that did not take stops the walk, an account's ceiling waits once, and a person's pick outranks the chain

santoshkumarradha:
## The evidence

The owner's own task, 2026-09-11 14:39:17 – 14:40:47, `tag:task` node 2 on
`deepseek/deepseek-v4.1-flash`. Eight consecutive sends came back
`(via Wafer: … is temporarily rate-limited upstream)`, every one carrying
`retry_after: 2`, spaced by the dispatcher's own doubling (0.7 s, 1.4, 2.8, 5.6,
11.2, 22.4, 44.8) — ninety seconds of `waiting · rate limited` while six other
machines on the same model were answering in under five, one of them thirty
seconds earlier in the same minute. Then the work hopped to a model the chain
chose, while the person had already picked one in the room and been told the
change was taken.

Both halves are reproduced as failing `lanestub` scenarios before anything moved
(`internal/provider/namedrefusal_test.go`): on `dev` the first stages eight sends
and 1 m 30 s of waits, the second seven sends of identical bytes.

## What changed

**A named refusal is asked before the status.** `taxonomy.Evidence.Named` is the
one spelling of "one machine owned this refusal", and `classOf` asks it before
any number: a 400, a 502 and a 429 relayed from one pool are one fact and earn
one answer. A 429 that named nobody stops borrowing `the provider could not
serve it` and gets its own shape, `ReasonPaced`.

**A veto that did not take is nowhere else to go.** The dispatcher's exclusion
law used to key on `relayed` — a 4xx that would not otherwise have been retried
— so the commonest named refusal in the log was exempt from it. It now keys on
the veto the body actually carried: when the machine that answers is one this
very body named in `provider.ignore`, routing cannot save this request, and the
refusal is handed back for the ladder and the session's one model hop. The
snapshot is taken at the encode rather than read back at the refusal, because
the veto list is shared by every arm of one question.

**An account's own ceiling is not an open set.** `control.Plan.AccountRefused`
is `ShapeRefused`'s sibling: a router with a pool behind the model paced the
whole key and named none of it, so every machine is behind the same ceiling,
nothing can come off the next body, and the open-set walk's premise is gone. It
earns the comeback it named, once, and then the model. A base that has never
published a pool is deliberately not this — an endpoint with one machine saying
"come back later" is answered by coming back later.

**A person's pick outranks the fallback chain.** `TaskNode.standingModel` is the
model somebody chose in a node's own room and that it is not running on;
`nextNodeModel` moves there instead of walking the adapter's list. A chain
answers "what next when nobody has said", and somebody had.

**And the room says what is true.** `model · <id> · the next request takes it`,
where it used to say `its next turn takes it` — a boundary that can be twenty
minutes off on a long step, which is why the person typed `continue` to try to
force it.

## Laws

- `internal/taxonomy`: a named refusal is `Transport`/`ReasonRefused` at every
  status; an unnamed pace is `ReasonPaced`; `Named()` reads the upstream alone.
- `internal/lane/control`: an account ceiling earns one comeback and then the
  model, is not answerable by a relaxed shape, and has no move at all when it
  named no comeback.
- `internal/session`: a rescue moves to the model the person picked; a model the
  proposal named is not a standing pick.
- `internal/provider`: the three `lanestub` scenarios above, plus the control
  case — a veto the router honours still moves with no wait at all.

## Manual

`lanes.md` "When a machine is too busy" loses the claim that an account-wide
rate limit is waited out for the whole of the role's give-up, and gains the
veto-that-did-not-take case. `tasks.md` and `models-and-cost.md` stop saying a
task's model change lands on its next turn.

## Deliberately not in this PR

Issue #891 §3 — the chooser demanding `provider.only` instead of advising with
`provider.order` — was built and then taken back out. It is blocked on a seam
this PR should not cross; the exact finding is in the report and in a comment on
#891.


santoshkumarradha:
Reworked at `87be0bd780b42508f2163a7cdb63eda6dd2e7571`, rebased onto `dev@eaa3d69de` (the task-start wave #929 and #918 are under it; the rebase was clean, including `task_run.go`, which #929 rewrote around the standing pick).

**Defect 1 — `the next request takes it` was false.** Built the cut half as ruled, then took it back out on the second ruling so this can land without `loop.go`/`agent.go`; it is handed to lane L11 whole, as a patch with a header saying what it is and what it owes the manual when it lands. What is here now is true of the code that is here: the room reads **`the next turn takes it; a rescue goes to it first`**. `its next turn takes it` was never false — it was ALL the room said, so over a grinding step it read as "nothing happens until this finishes". The second clause is what this PR actually landed and is the half that answers the person who typed `continue`. `tasks.md`, `models-and-cost.md` and the change entry say exactly that and nothing more.

**Defect 2 — `demandedAlone` asked the advisory set.** Now `demandedLane`; `demandedAlone` is deleted rather than narrowed, since `demandedLane` was already the one door for "did this request demand one machine". Two tests: `TestOnlyADemandExemptsAMachineFromTheVetoLaw` (a table over hedge arm / strict pin / demand of two / candidate set of one / candidate set of two / nothing — **fails on the old reading**, on exactly the "a candidate set of one is a belief, not a demand" row) and `TestOneMachineBelievedInIsStillNotADemand` on the wire.

One correction to the reading, for the record: staged on the wire, the old exemption cost **one** extra send, not eight — `control.Next` bounds a one-wide set to a single comeback, so the observable harm was a wrongly shaped ending (`after 2 attempts:` rather than "nowhere else to go") plus a doc comment claiming something false. The fix is right either way, and the table test is where the teeth are rather than the scenario.

**Defect 3 — `models-and-cost.md` documented the deleted road.** Rewritten to the two roads (a named 429 is a move with no wait; an unnamed one is one wait for what it asked for, then another model), with the veto-that-did-not-take stop added, and the stale budgets in the section below it gone too — "four times in all", "2s, then 4s, then 8s", "a full four tries of its own", and the `2 of 4` now explained as machines rather than tries.

**Follow-ups taken:** 4 (the account wait names nobody; its test has a head on the plan so the assertion can fail), 6 (a named comeback is the whole wait — it went through `backoffFor`, which returns the larger of it and the doubling, so 3s asked for was 11s paid on a fourth attempt while the person read `back in 3s`), 7 (`workPolicy` asks `Named()`), 9 (`sentVeto` starts at what the caller's own encode saw).

**5** taken as the second option you offered: `sentVeto` is documented as what it is — every machine that had refused this call — with both cases where that is wider than the body's `ignore` list named, and why handing back is right in both. Snapshotting the encoded object means returning the preferences out of `encodeRequest`, which has callers past this loop; better shape, not this change.

**8 is not taken**, deliberately, and is the one I would most like a second opinion on. Feeding `Evidence.Upstream` from the door's answer means carrying that answer on `APIError` as a fourth mark beside `Routing`/`Account`/`Withdrawn` — which is right, and changes what `Named()` is true of for every pinned call. It is in `~/af-defect-ledger.md` as an open entry rather than smuggled in here.

Report at `~/af-rec-l9.report.md`.

santoshkumarradha:
Second pass addressed at `0ceb54126bb805a6c2672b5e3353682d80754023`, rebased onto `dev@7840426d7`.

**Defect 11.** You were right, and the reading was sharper than "the wording is too broad" — `a rescue goes to it first` was ahead of its mechanism in **two** independent ways, either of which alone made it false, and I had missed both. Taken as ruling (b), scoped to task steps, in my own files.

1. **The gate refused the exact failure the sentence was written for.** `movesForFailure` returns false for every Transport verdict and a 429 is Transport by this branch's own `classOf`, so the owner's own scenario — a step grinding on `temporarily rate-limited upstream` — never reached `nextNodeModel` at all. It now asks the standing pick **before** it classifies: what this gate exists to refuse is *buying* a dearer model to answer for somebody else's bad minute, and a model a person named is not a purchase. That is the rule saying what it is about, not a second reading of the class bolted beside it — written the other way round (classify, then let a pick past a Transport verdict) there would be two answers to "may this node move" for a later wave to drift apart. `TestAPacedStepNobodyChoseForStaysOnItsModel` holds the other half: nobody chose, nothing moves.

2. **The pick was invisible to the very move it was made for.** Exactly as you traced it: `retargetLocked` writes `spec.model` and clears `ran`, so `chosen == runModelLocked()` compared the pick against itself. Read honestly, my own passing test was telling me this — it had to stage a prior `runOn` by hand to see the pick at all. It is a **fact** now, not a derivation: `TaskNode.picked`, written by the retarget, spent by `TaskNode.startedOn` where the worker is actually built. And the model a failed run used is read **once** by the run loop (`ranOn`) before another goroutine can rewrite it, so the pick is weighed against what actually failed rather than against itself. `TestAPickNamingTheModelThatFailedIsNotAMove` pins the degenerate case.

`TestTheFirstMoveAfterAPickGoesToThePick` walks the run loop's own road — the gate, then the move — against a real relayed-429 `APIError`, and it fails **twice over** when reverted: `a step being paced stayed on the machine pacing it` with the gate's reading removed, and `the pick reads "" the moment it is made` with the fact turned back into the derivation.

**Scope, as ruled.** A conversation's hop is `loop.go`'s `nextFallback` and is L11's, so `models-and-cost.md:1344` now says a `/model` in the chat lands on your next message and points at the tasks page; `:57` and `:1306` are scoped to a task's own room; `room.go:275` is the task room and keeps both halves, with the comment naming the two functions that make it true.

**Defect 12** fixed — the change entry's second line now says what the code does, including *why* `sentVeto` is deliberately wider than the body's `ignore` list.

Spark green at this exact sha, all seven packages (`session` 174.970s, `tui3` 308.896s, `provider` 64.751s). `light gate` pass; `touched packages` running.

PR #938, opened 2026-09-11
chat: a call refused on its own arguments says so, and the schema's sentence stays with the model

### LINKED ISSUE #890: chat: a refused `propose_task` spills its raw schema error into the conversation — "checks must each be ONE command with no shell composition"
## What happened

2026-09-11, `dev@333acc67d`, live drive of the chat surface, scenario 6. The model called `propose_task` with a composed shell command in `checks`. The engine refused the arguments. The person's screen got both halves of a conversation that was never theirs:

```
╰─ not started · the call was refused
```

and, opened underneath it, the refusal text itself:

```
Invalid arguments: checks must each be ONE command with no shell composition — "cd 1-check && ./run.sh" is not
```

`checks`, `shell composition` and the quoting of the offending argument are the schema talking to the model. The person is being shown a repair instruction addressed to somebody else, about a field they have never heard of, on a card that says nothing started.

## Replication

**Deterministic (no model).** `internal/tui3`, through the reducer the chat uses — `internal/tui3/forming_test.go` already drives this exact shape (it settles a forming card with `Hint: "Invalid arguments: brief is required"`):

Feed the feed a `session.EventToolBegin` for `propose_task`, then

```go
session.Event{
    Kind:   session.EventToolFailed,
    Tool:   "propose_task",
    Hint:   `Invalid arguments: checks must each be ONE command with no shell composition — "cd 1-check && ./run.sh" is not`,
    Output: `Invalid arguments: checks must each be ONE command with no shell composition — "cd 1-check && ./run.sh" is not`,
}
```

and render.

```sh
go test ./internal/tui3 -run TestAToolArgumentRefusalIsNotDrawnToThePerson -count=1
```

**Expected today:** the rendered frame contains `not started · the call was refused` **and** the whole `Invalid arguments: checks must each be ONE command with no shell composition` sentence, in an entry that opened itself.

**Expected after the fix:** the frame contains the card's own settled words and no `Invalid arguments:` text.

**Field (real models).** `OPENROUTER_API_KEY`, the chat surface, a `/task` whose acceptance is naturally a composed command (`cd X && make test`). Small models compose in `checks` routinely. One turn, cents.

## Where — the leak, in two hops

1. **The text is written for the model.** `declaredCheckList` in `internal/session/task_checks.go` returns `"Invalid arguments: checks must each be ONE command with no shell composition — " + strconv.Quote(...) + " is not"`. That is correct and must stay: it is one repair away from a working check, which is the stated reason the door refuses rather than dropping silently.
2. **The engine puts it on the wire as a person-facing hint.** `internal/session/loop.go`, the batch-result loop (search `Kind: EventToolFailed`), sends `Hint: clip(firstLine(results[index].text), hintLimit)` and `Output: capOutput(results[index].text)` — the tool's own result text, unchanged.
3. **The surface draws it and opens it.** `feed.closeTool` in `internal/tui3/feed.go`, reached from `f.closeTool(ev, toolFailed, firstNonEmpty(ev.Hint, errText(ev.Err)))`: it writes `e.text = e.text + " — " + why` and then sets `e.open = true` under the comment **A FAILURE OPENS ITSELF**. That law is right for a bash that failed; a schema refusal is not that kind of failure.
4. The card beside it is `app.refuseFormingCard` / `taskFormingRefused` in `internal/tui3/task.go` — `not started · the call was refused` — which is the correct and sufficient sentence.

## The fix

**A tool-argument refusal is a result the model reads and repairs, and never a sentence drawn to the person** — unless the model gives up on the call, in which case what the person reads is the model's own account of giving up, in its own words.

The row for a refused call says what a person can act on: the call was refused, and the model is fixing it. The schema's sentence stays where it belongs — in the tool result, in the transcript, and behind `ctrl+o` for somebody who asks for it — rather than being drawn expanded by default. Concretely, a failure that is an **argument refusal** (the engine knows: it never reached the tool's work) does not take the `A FAILURE OPENS ITSELF` road, and its hint is not the raw result text.

The existing sentence `not started · the call was refused` is already the right words and does not change.

## Acceptance

- **e2e:** `go test -tags e2e ./internal/e2e/` (the tmux suite) — a task proposal refused on its arguments draws `not started · the call was refused` and the screen never contains `Invalid arguments:`.
- **e2e:** the control — a genuinely failed `bash` still opens itself with its own output, so `A FAILURE OPENS ITSELF` is intact for the case it was written for.
- **Unit:** `internal/tui3` — a `toolFailed` entry whose failure is an argument refusal renders with `open == false` and with no substring of the refusal text in its drawn row; `internal/session` — the refusal still reaches the model verbatim as the tool result, asserted on the transcript.
- `internal/manual/chat/how-tasks-run.md` (which today quotes `Invalid arguments: checks must each be ONE command with no shell composition` as something the screen shows) is corrected to say where that sentence goes, and the change entry's `invalidates` names the belief that a refused call spills its schema error into the conversation.

santoshkumarradha:
Closes #890.

A `propose_task` the belt refused on its arguments spilled the schema's own sentence into the conversation — field names, the offending argument quoted back, *checks must each be ONE command with no shell composition*. That is a repair instruction addressed to the model. It landed on the person as somebody else's mail, beside a card that already said the call was refused.

The row now carries the one fact a person can act on and stays shut. The schema's sentence stays where it already was: in the tool result the model reads and repairs from, and in the transcript behind `ctrl+o`.

## The mechanism, not a message match

`session.ArgumentRefusal` answers with **the loop guard's own eyes** — `argumentRepair`, the same function the guard uses to decide a call is one repair away from working. So the guard and the surface cannot disagree about which failures are the schema talking to the model. A bash that failed out in the world is still the one row whose detail is why the person is looking at the screen, and it still opens itself.

## Where this came from, and what I changed in it

**An aforge lane wrote this** — `~/af-lane.sh 890`, `aforge do` on `z-ai/glm-5.3` through OpenRouter, unattended on the Spark. I reviewed the diff and made two corrections before opening it.

**Dropped:** the lane had also edited `cmd/aforge/usage_test.go` and `vocabulary_test.go` to put a mutex around the help-output seams. That is unrelated to this issue, and the mechanism does not do what its comment claims — it takes the lock in the setup helper and releases it before the test body runs, so two parallel tests still clobber each other. If that seam is genuinely racy it deserves its own issue and a fix that works.

**Kept and committed:** the lane replaced a `time.Sleep(5 * time.Second)` in its own new end-to-end subtest with a wait for a word the turn only produces when it finishes, the way `testAskHere` already waits. That removes five seconds from every run and the flake that arrives when a turn takes six.

## Proof

Spark, rebased on `52af732bb`:

```
ok  github.com/Agent-Field/aforge-v2/internal/session  175.555s
ok  github.com/Agent-Field/aforge-v2/internal/tui3     308.365s
ok  github.com/Agent-Field/aforge-v2/internal/e2e        0.393s
ok  github.com/Agent-Field/aforge-v2/internal/manual     1.679s
```


santoshkumarradha:
One correction to the note above, now that I have checked rather than assumed.

The seam change I dropped is not a fix for a real problem. Its comment claims `go test` runs tests in a package concurrently "whenever they call t.Parallel" and that a cleanup in one test restored a seam while another held it. **No test in `cmd/aforge` calls `t.Parallel`** — `grep -n 't.Parallel' cmd/aforge/*_test.go` returns nothing — so tests in that package run one at a time and the failure it describes cannot occur. There is no issue to file.

That is worth recording as a thing to watch for in lane output: the fix was plausible, the comment was written in this repository's voice, and it was about a failure mode that does not exist here. Reading the diff is what caught it.

PR #951, opened 2026-09-11
tasks: a call that is still open says so, with its own clock and rate

### LINKED ISSUE #837: The task room draws nothing about a call that has been open for ten minutes
## What happened

2026-09-10, build `839884259`: during one request that ran 12 m 04 s (a task node on `z-ai/glm-5.3-flash`, 32 532 reasoning tokens), the task room drew nothing about it. The header said

```
working · 10m 21s · 4 tool calls
```

the whole time, with no sign that a single call had been open for ten minutes or how fast it was writing. The engine writes exactly that fact to the task's beat sidecar: `request_started` newer than `request_finished` means "inside a call". **Nothing in the surface reads it.** `readTaskBeat` in `internal/session/task_beat.go` has no caller outside its own test.

## Replication

**Deterministic (no model).**

```sh
grep -rn 'readTaskBeat\|RequestStarted' internal --include='*.go' | grep -v _test.go | grep -v internal/session/task_beat.go
```

prints nothing today: the sidecar has a writer and no reader.

**Field (real models).** Open a task whose node makes one long call (any reasoning model at `effort: high` on a long plan). Watch the task room header during the call: it shows the task's elapsed time and tool-call count, and nothing about the open call.

## Where

- The writer: `taskBeat.began` / `taskBeat.ended` in `internal/session/task_beat.go` (fields `request_started`, `request_finished`).
- The task room's live header in `internal/tui3`: the line that read `working · 10m 21s · 4 tool calls`.
- The chat foot already draws an in-flight call's elapsed time and live token rate (since #747). That is the vocabulary to reuse.

## The fix

The task room reads the beat. While `request_started` is newer than `request_finished`, the header's live segment shows the open CALL's elapsed time and its live token count and rate, spelled exactly as the chat foot spells them. When the call ends, it goes back to the task's own elapsed time. No new words: the chat foot's sentence is reused.

## Acceptance

- **e2e:** the tmux suite (`internal/e2e`) opens a task against a stub whose node holds one call open for several seconds while streaming, and asserts the room shows the chat foot's in-flight segment during the call and drops it after. The needle goes into `internal/e2e/tuiwords_test.go`.
- **Unit:** a `tui3` render test feeding a beat row with `request_started` after `request_finished` draws the segment; the reverse does not.
- The manual page for the task room (`internal/manual/chat/`) says what the segment means.

santoshkumarradha:
Closes #837. Draft until the manual page and change entry are on it.


santoshkumarradha:
Spark proof, rebased on `d726ff259`:

```
ok  github.com/Agent-Field/aforge-v2/internal/tui3    309.802s
ok  github.com/Agent-Field/aforge-v2/internal/e2e       0.325s
ok  github.com/Agent-Field/aforge-v2/internal/manual    1.691s
```

`internal/session` was red on that run, and it is **not this branch**. A clean dev worktree running `go test ./internal/session/ -count=1` alone is also red, with a different test each time. Filed as #959 with the three runs that reproduce it. The four tests this branch's run turned red all pass in isolation on both dev and this branch.

## What I corrected in the lane's work

An aforge lane wrote this over two runs on `z-ai/glm-5.3`. The feature code was right both times; four things around it were not.

1. **The first run was plumbing with no consumer.** It carried the pulse all the way to `taskRoom` and nothing drew it, so the issue was not closed. I wrote what was done and what was missing onto #837, naming the consumer its own comment had identified, and re-ran the lane against that. The second run finished it.
2. **The export missed a caller.** `readTaskBeat` → `ReadTaskBeat` left `task_calltrail_test.go` behind and `internal/session` would not vet. The lane never ran `go vet` on that package.
3. **Its two new tests failed, and the feature was fine.** They built beat rows from `time.Now()` while this surface runs on a fixture clock weeks away from the wall clock, so the request start landed in the app's *future*, the elapsed came out negative, and `countUpWord` returned empty — which on a failure reads exactly like the header refusing to draw. `beatRoom` now takes how long the call has been open and derives its instants from `a.now()`.
4. **No change entry and no manual page.** Both written here.

PR #962, opened 2026-09-11
A split drawing reaches the boundary its own cut opened

### LINKED ISSUE #956: a split drawing can miss the boundary its own cut opened: the turn pays another generation, and the drawing is dropped when that step is the last
## What happened

Found by reading rather than on a screen: lane FLAKE met it while fixing the five
load-sensitive reds and reported it in **PR #952 §5 ("Found, not fixed")** instead of
working around it, because closing it needs `checkpoint.go`/`steer.go`/`loop.go`
semantics that lane did not own. Base `dev@c742b8bc8`.

The mechanism is **#871**'s one reading beside the work. A mark's drawing rides beside
the next step (`markAside.start`), and its single power over the work is `act` →
`Agent.cutGeneration(errMarkCut)`: stop the request in flight so that **the boundary
where the drawing can be spent arrives at once** rather than whenever the step happens
to end. Two orderings break that promise, and both end the same way — the turn pays for
a whole extra model step that the cut had just bought its way out of, and where that
step is the turn's last, the drawing is written down as a carry-on and the handover
never happens.

**(1) The cut is delivered before the answer is takeable.** In `readBeside` the reading's
goroutine runs `act(answer)` and only then `close(side.settled)`:

```go
go func() {
    answer := ask(readCtx)
    side.answer = answer
    if act != nil && readCtx.Err() == nil {
        act(answer)          // ← markAside's act: a.cutGeneration(errMarkCut)
    }
    close(side.settled)      // ← only now can take() see it
}()
```

That order is deliberate and #871's doc states why: an answer taken *before* the cut
landed would be spent against a step the interruption was still about to cut. But
`sidecar.take` is **non-blocking by construction**. So the real sequence is: cut
delivered → the provider unwinds → `completeWithRetryReasoning` returns `errMarkCut` →
the loop's `continue` → `Agent.checkpointSettle` → `aside.take()`. A loop that gets
there before the reading's goroutine reaches `close(settled)` reads *pending*, the
boundary passes, and a fresh request goes out. The cut was paid and bought nothing.

**(2) A drawing that lands with nothing in flight cuts nothing.** `Agent.cutGeneration`
returns false when `a.generation == nil`, and `a.generation` is installed by
`beginGeneration` under `a.mu` a long way *after* the boundary's take: the loop takes at
the top of the round (`checkpointSettle`), then parks on owed jobs, drains steering,
guards the request size, compacts tool history and assembles the transcript before
`beginGeneration` runs. A drawing landing anywhere in that gap cuts nothing and is spent
only at the boundary after the next step. `cutGeneration`'s own doc gives the reason
this is supposed to be harmless —

> a generation that has not started yet needs none (the transcript it will be assembled
> from already carries the change)

— and **that reason is true of the recall and false of a drawing**. The recall's block is
recorded into the transcript, so the next request carries it whether or not anything was
cut. A drawing rides no transcript: it exists only to be spent at a boundary, and the cut
is the only thing that brings the boundary forward.

**And then the drop.** There is exactly one other taker, `runTurn`'s exit defer:

```go
if landed, ok := marked.takeAtTheEnd(); ok {
    a.journalMarkRead(landed.read, landed.mark, landed.rounds, landed.read.sketch.carryOnDecision())
}
```

`carryOnDecision` returns `checkpointDecisionContinue` (`"continue"`) for **any** sketch
that does not hand back — including one whose `split()` is true (`parts >= 2 &&
!handsBack`). So a drawing that said *this has parts*, and whose landing missed the
boundary its own cut opened, is journalled `{"decision":"continue"}`,
`handOverRunningTurn` is never reached, no ceiling row is written, and a turn with
independent parts left in it ends in words. The ledger line is kept — the defer's comment
is right that a paid-for reading must not be a line nobody can count — but the *decision*
is lost, and the file then says the reader decided to carry on when it decided the
opposite.

The lane's instrumented traces (twelve CPU burners, `GOMAXPROCS=2`, prints reverted)
show that road is reachable and is where a drawing goes to die: `markReading take
pending` at eight consecutive boundaries and then `markReading takeAtTheEnd`. In those
particular runs the cause was the fixture's own guess (fixed by #952), not this ordering —
the evidence for this defect is the code above, which is why the replication below is
written to be run rather than quoted.

## Replication

**Deterministic (no model).** Two, both with fixtures that already exist in
`internal/session`; a stranger needs no key, no network and no box of ours.

*(a) The ordering, at the sidecar.* Call `readBeside(ctx, ask, act)` with an `act` that
delivers its interruption and then blocks on a channel the test holds, and call
`side.take()` at the instant the interruption is observable. It returns `(zero, false)`.
That is precisely the boundary the cut opened, reading *pending* — no goroutine racing,
no load, sub-second, and it is the product's own ordering with nothing stubbed.

*(b) The consequence, through the real door.* `checkpointAgent` + `scriptedCompleter` +
`stubbedGraph`, with `askedForSketch(messages)` to recognise the mark reader's call and
`journaledMarks(t, path)` to read the rows back — the same three helpers
`TestAMarkNobodyCouldReadIsJournaledAsAFailure` uses today. Script a drawing with
independent parts (two or more parts, no hand-back, so `checkpointSketch.split()` is
true) and release the sketch call's answer only after the completer has served the turn's
final answer, so the landing is strictly after the last generation. Run it through
`Agent.Submit`. **What a developer sees today:** the mark row reads
`{"decision":"continue"}` with a many-part `sketch` on it, `journaledCeilings` is empty,
and `stubbedGraph` sees no node — a mastermind call that said "this has parts" written
down as a carry-on, and the work never handed anywhere.

```
go test ./internal/session/ -run '^TestASplitDrawingIsSpentByTheBoundaryItsCutOpened$' -count=1
```

**Field (real models, `OPENROUTER_API_KEY`).** Any unattended run long enough to cross a
mark with divisible work in it — `aforge do "<a multi-part task>" --yes-spend` — and then
read the session journal for the fingerprint, which needs nothing private:

```
grep '"mark"' <session journal> | jq 'select(.mark.decision=="continue" and (.mark.sketch|test("\\|")))'
```

A `mark` row whose drawn sketch carries separators but whose decision is `continue` is
either this bug or a hand-back — and a hand-back journals `waiting`, not `continue`, so
the row is unambiguous. Budget a few minutes and a few cents.

## Where

All in `internal/session`, by searchable string (line numbers are at `dev@c742b8bc8` and
will move):

- `readBeside` in `sidecar.go` — `act(answer)` then `close(side.settled)` (lines 111-118);
  `sidecar.take`'s non-blocking `select` with its `default` (lines 129-142);
  `sidecar.takeAtTheEnd` (line 154).
- `markAside.start` in `checkpoint.go` — the `act` that calls `a.cutGeneration(errMarkCut)`
  (lines 2773-2787); `markAside.take` / `takeAtTheEnd` (2790, 2800).
- `Agent.checkpointSettle` in `checkpoint.go` — `landed, ok := aside.take()` and the
  `if !ok { return false }` that lets the boundary pass (lines 2826-2849); the split arm
  calls `handOverRunningTurn`.
- `Agent.checkpointRound` in `checkpoint.go` — `aside.start(ctx, a, mark, rounds)` at the
  first two marks (line 2714), `aside.end()` at the ceiling (2729).
- `Agent.runTurn` in `loop.go` — the boundary take at the head of the loop
  (`a.checkpointSettle(...)`, line 785), the `errMarkCut` arm that `continue`s (853-864),
  and the exit defer whose `marked.takeAtTheEnd()` journals with `carryOnDecision()`
  (lines 730-749).
- `Agent.cutGeneration` and `Agent.beginGeneration` in `steer.go` — the `a.generation == nil`
  return false (182-190) and the install under `a.mu` (148-155); the install's call site is
  `loop.go`'s `attemptCtx, generation := a.beginGeneration(attemptCtx)` (line 1712).
- `checkpointSketch.split` and `checkpointSketch.carryOnDecision` in `checkpoint.go`
  (1202, 1263), `checkpointSketchParts = 2` (299), `checkpointDecisionContinue = "continue"`
  (1849).

## The fix

**The drawing's landing is the fact the boundary waits on, and the lock that already
decides whether there is a generation to cut is the one place that can order it.** Both
windows are the same missing ordering seen from two ends, so they take one fix and not
two: `a.mu` is already taken by `beginGeneration` to install a generation and by
`cutGeneration` to cancel one, so let the drawing's landing be published through that same
lock — the answer is readable before the interruption is delivered, the interruption is
delivered under the lock, and the boundary's take is therefore ordered *after* the cut by
the lock rather than by a race, which is what #871's act-before-settle rule was reaching
for and could not reach with a non-blocking channel test. The second window closes with no
extra mechanism, because `beginGeneration` is then the one place that can see a landed,
unspent drawing while holding the lock and hand the loop back to its boundary instead of
paying for a step nobody will read. No "was there a cut" flag, no retry counter, no
widening of `take` into a wait, and no second taker: one ordering under one lock, after
which `takeAtTheEnd` stops being where split drawings go to die because a split landing can
no longer be missed by the boundary it opened. (If a drawing genuinely can only land after
the turn's last step, the exit defer must journal what it actually says — `split` — rather
than `carryOnDecision()`, so the file never records the opposite of the reading.)

## Acceptance

- **e2e:** through `Agent.Submit` with the scripted completer of replication (b) — a
  drawing with independent parts landing in the window that follows a mark's cut: the
  journal carries a mark row with `"decision":"split"`, a ceiling row with
  `"decision":"moved"`, and exactly one node ran; the turn makes **no** further model call
  between the cut and the handover (counted off `journaledCalls`), which is the figure the
  cut is bought for.
- **e2e:** the control — a carry-on drawing (one part, or a hand-back) landing in the same
  window still journals `"continue"` / `"waiting"`, launches nothing, and the turn carries
  on exactly as it does today.
- **Unit:** the ordering law at the sidecar — a reading whose `act` has delivered its
  interruption is takeable at the boundary that interruption opened (replication (a) turned
  round: `take()` returns the answer), plus the structural half already on the laws gate
  (`sidecar_law_test.go`) extended so the only `takeAtTheEnd` on a `markAside` stays the
  turn-ending one.
- Nothing on this road is person-facing wording, so no manual page changes; the change
  entry's `invalidates` names what was believed — that a mark's cut always opens the
  boundary its own drawing is spent at, and that a drawing journalled `continue` was a
  reader that decided to carry on.

*Reported out of PR #952 §5; the mechanism is PR #871's.*

santoshkumarradha:
Closes #956.

A mark's drawing rides beside the next step and its one power over the work is the
interruption: cut the request in flight so that **the boundary where the drawing can be
spent arrives at once** rather than whenever the step happens to end. Two orderings took
that promise away, and both ended the same way — the turn paid for a whole extra step, and
where that step was its last the drawing was written down as a carry-on and nothing was
handed anywhere.

**The answer was published only once the interruption had been delivered.** `readBeside`
ran `act` and then closed `settled`, and `sidecar.take` is non-blocking by construction —
so the loop arriving at the very boundary the cut had just bought read the reading as still
in flight and let it pass. The answer is published first now. The order that protected
against a late `act` is kept by one claim instead: `sidecar.spent` is CAS'd by the
interruption and stored by the taker, so whichever arrives first is the one that happens,
and an answer that has already reached a boundary raises no interruption into whatever the
turn does next.

**A cut that arrived between two steps found nothing to cut and was dropped.** Whether that
is safe is a property of the reading behind the cut and is now declared by the cause
(`boundaryCut`), not decided at the call: a recall's block is written into the transcript
before its cut is raised, so the request being assembled carries it either way — while a
drawing rides no transcript at all and the boundary is the only place it can ever be spent.
So a boundary cut is **owed** rather than spent on nothing, and `beginGeneration` — already
the one place that decides whether there is anything to cut — spends it the instant there
is. An owed cut belongs to the turn that asked for it, by number, so a cut one turn never
spent cannot reach the next person's message.

**And the one road where a drawing genuinely lands too late says so.** A turn that answered
in words has no boundary after it. The exit defer still waits for the drawing and still
writes the row — a reading somebody paid for must not be a line nobody can count — and the
row now says `late` rather than `continue`, which was the file recording the opposite of
what the reader decided.

### The bar

- **The one abstraction**: `sidecar`'s answer-and-interruption claim, and the cut door's
  `boundaryCut` property. Every reading beside the work (recall, mark, route judge, and
  whatever is added next) gets the boundary contract for free; every cut declares in one
  place whether it may be dropped.
- **Deleted**: the act-before-settle ordering and the paragraph arguing for it; the
  unconditional `return false` for a cut with nothing to cut; `carryOnDecision` on the
  turn's exit road.
- **Guarded by**: `TestTheAnswerIsReadableBeforeTheInterruptionIsDelivered` (go/ast, on the
  laws gate, with a plant that is the pre-#956 order), plus four runtime tests and one
  control in `cutboundary_test.go`.
- **Not a band-aid**: no flag asking "was there a cut", no retry counter, no widening of
  `take` into a wait, and no second cut path — two windows, one ordering, in the two places
  that already owned it.

### Proof

Every test ran on the Spark at this PR's head; the sha and the output tail are in the
report. Three of the new tests were verified red on the pre-fix code before the fix landed,
including the issue's own replication (b), which reproduces `{"decision":"continue"}` on a
many-part sketch.


santoshkumarradha:
Rebased onto `origin/dev` @ `741b1dad1` (#963). Head is now `f2fcb089139ac55dc5bd772ba7aaf4e1e186ab18` — same two commits, clean rebase, no conflict; none of #963, #954, #951 touches sidecar.go / checkpoint.go / steer.go / loop.go.

**CI at that head: three green** — [run 34656084090](https://github.com/Agent-Field/aforge-v2/actions/runs/34656084090): `check` pass, `light gate` pass (4m03s), `touched packages` pass (4m43s). `touched packages` is the whole `./internal/session` suite with `-count=1`, so the six runtime tests ran there; `light gate` runs the two structural laws again through `scripts/laws.sh` (`laws: 103 files in 26 packages`).

Spark `make pr-ready BASE=origin/dev` is queued at the same sha behind the per-box suite lock; its result follows here.

On the L11 seam (#935, reworking): this change adds **no** `generation.cancel` call site of its own — `grep -n 'cancel(' internal/session/*.go` finds only `steer.go:223` (the door) and `steer.go:361` (`Agent.Steer`, which #935 deletes) — so a law that matches cuts by type rather than by receiver spelling names nothing here. The owed cut is spent by calling the door (`a.cutGeneration`), never the handle `beginGeneration` is holding.

santoshkumarradha:
**Both gates green at the same sha: `37b1530a1afa0e9c1c95209550495c926288ef7a`** (rebased onto `origin/dev` @ `84016ad09`, which carries #973's manual fix).

**CI** — [run 34659227107](https://github.com/Agent-Field/aforge-v2/actions/runs/34659227107): `check` pass, `light gate` pass, `touched packages` pass.

**Spark** — `spark:~/af-l13` detached at that sha, `make pr-ready BASE=origin/dev`:
```
laws: 103 files in 26 packages          (all 26 packages ok)
touched: ./internal/session
./scripts/one-suite.sh go test -timeout 15m -count=1 -p 1 ./internal/session
ok  	github.com/Agent-Field/aforge-v2/internal/session	175.157s
```

### One red on the way, measured rather than rerun

The first `touched packages` attempt at this sha failed on `TestAHeadlessRunLeftGoingCheckpointsAtTheSameRoundsAsAChatOne/watched_conversation` ("the mark reader was asked 2 times, want 3") — a test that reads marks, so it was measured against dev rather than waved away. Back to back on the same box:

```
go test -count=200 -run TestAHeadlessRunLeftGoingCheckpointsAtTheSameRoundsAsAChatOne ./internal/session
dev  84016ad09 : 2 failures in 200
head 37b1530a1 : 0 failures in 200
```

The sketch that test grinds is `checkpointChainSketch` — a carry-on, never a split — so no cut is raised in it and this change cannot reach it. That is the fourth such control on this branch (the earlier three, on `TestARacedYesIsLookedAtEarly…`, `TestARatifyIsShownAndWaitsOnNobody` and `TestTheCeilingsQuickNodeHasTheToolsShape`, all gave the same rate at base and head). All of them are in `~/af-defect-ledger.md`; nothing was added to `.github/known-red.txt`, which only shrinks.

### Seams with the two branches in flight

- **#923 (`feat/turn-promotes-itself`)** touches none of loop.go / sidecar.go / steer.go / session.go, and my three checkpoint.go hunks sit on anchors its rewrite keeps: `carryOnDecision` (still there), the decision const block (which it extends with `told` the same way I extend it with `late`), and `markAside.start` (still the one `readBeside` starter). `loop.go:745`'s `marked.takeAtTheEnd()` is byte-identical on their branch.
- **#935 (L11)** — this adds no `generation.cancel` call site of its own, so a law matching cuts by type rather than receiver spelling names nothing here; the owed cut is spent by calling the door, never the handle.

Happy to rebase onto either as soon as it lands; re-running both gates at the resulting sha. **Not merging.**

santoshkumarradha:
Review's fixes are in. Head is now `75eac176a7b2e881d490bb6ef325345990cdfb16`, rebased onto `origin/dev` @ `9a7ab78cc` (the #952 conflict in sidecar.go is resolved below).

**1 — the claim's one-instruction hole, which this PR made cost a request.** Between `close(settled)` and the CAS the interruption can still win while a taker at that same boundary takes anyway. Nothing is lost, but at a boundary there is no live generation, so that cut is now OWED rather than dropped and on the carry-on arms (`handOverRunningTurn(...).moved == false`, `checkpointSettle`'s `ctx.Err()`) it stops the next step for a reading already read. Closed at the fact, not the race: **the boundary that takes the drawing withdraws the cut it no longer needs** — `Agent.dropOwedCut`, one branch under the lock that already owns `cutOwed`, called once from `checkpointSettle` right after a successful take, spelled by cause so one reading never withdraws another's. It is exact rather than racy: the take and the spend are both on the turn's own goroutine, so by the time `beginGeneration` could spend the cut, the take has happened.

Neither ordering suggested in the note actually closes it, and I want that on the record rather than silently not done: **CAS-before-close** means no taker can ever have claimed (a taker claims only after the settle), so the interruption always wins and the double survives; **`take` consulting `spent`** either refuses the drawing forever — the next boundary would find `spent` already true — or has to wait for `act`, which is the law `take` exists to keep. The claim stays, because it is what suppresses the interruption in every other interleaving; the withdrawal is the other side of the same fact.

**2 — the `spent` doc** now says what it does: the claim GATES THE INTERRUPTION AND DOES NOT ORDER THE TAKE, and it names the one-instruction window and the withdrawal that covers it.

**3 — the law no longer bites on a name.** It finds the reading and the interruption **by their shape** (`readBeside` takes two function parameters; the reading answers, the interruption does not), requires the closed channel to be the `settled` selector, and **fails closed** when the door never delivers its interruption. `TestTheLawNamesADoorThatNeverDeliversItsInterruption` plants all three holes at once — an interruption renamed `raise`, never called, closing `side.done` — and the law names them.

**4 — the claim has a test in its own direction**: `TestAnAnswerAlreadyClaimedAtABoundaryRaisesNoInterruption` (reading held on a channel until after `side.claim()`; #952's `besideWatch.quiet` is what says it landed — no sleep), plus `TestACutOwedForADrawingTheTurnHasTakenIsNotSpent` and `TestADrawingWithdrawsOnlyItsOwnOwedCut`. Both new mechanisms verified red-before by surgical revert.

**5 — the stopped-turn arm is filed as #984**, with a replication a stranger can run and an end-to-end acceptance, and it is named in the change entry (including that this PR makes that arm more reachable).

**The #952 rebase**, since it landed in my file: `besideWatch` put `watch.landed()` exactly where `close(side.settled)` moved to. Resolved by keeping both orders — the settle closes first (#956), the interruption is delivered second, the watch hears it **last** — so a test waiting on `quiet` is waiting for the interruption too, which is what makes fix 4 deterministic. Their law and mine now sit side by side in `sidecar_law_test.go`; both pass.

**CI at this head: three green** — [run 34668393898](https://github.com/Agent-Field/aforge-v2/actions/runs/34668393898). Spark `pr-ready` is re-running at the same sha; the first attempt hit `TestASplitSketchAtTheFirstMarkHandsTheTurnOver` on its own 60 s `ran.await` deadline (the message that test prints for exactly this case). Checked rather than assumed: `handOverRunningTurn` takes `a.mu` twice inside `checkpointSettle` itself, so the lock is provably not held where `dropOwedCut` takes it — no deadlock is possible there — and the test is 30/30 green alone at this head and 30/30 alone at dev `9a7ab78cc`. Result follows. **Not merging.**

PR #963, opened 2026-09-11
questions: a clock armed on a question is let go when the session closes

santoshkumarradha:
One line in `Agent.Close`, and the test that could only pin half of it before.

`Close` already lets go of everything a steer armed, and says why in one sentence: **nothing armed outlives the session that armed it** (`steer_grace.go`). A clock armed on a question was the one thing it did not let go of, so a person with a `recommend` rule an hour long left a live `time.Timer` and its goroutine standing behind a closed conversation for that hour.

```go
 // Nothing armed by a steer outlives the session that armed it
-// (steer_grace.go).
+// (steer_grace.go), and nor does a clock armed on a question (asklane.go).
 a.stopSteerGraceLocked()
+a.stopAskClocksLocked()
```

The verb it calls is this lane's own (`asklane.go`), a two-line seam onto `askedOfThePerson.stopClocksLocked`, so that `agent.go` names an ask-lane verb rather than reaching into another file's map.

## Nothing a person could see changes

`askClockRanOut` already refused to decide anything once the session had closed — that is the **guard**, and it stays. This is the other half: the clock **does not fire at all**. The two are different facts and they now have a test each.

- `TestAClockThatRunsOutAfterTheSessionClosedDecidesNothing` (already there) pins that a clock which fires anyway writes nothing.
- `TestClosingTheSessionStopsTheClockOnAQuestionNobodyAnswered` (new) pins that it does not fire: `time.Timer.Stop` reports whether it was the call that stopped a timer still running, so a second `Stop` answering **false** is the proof that `Close` made the first. The question itself is left exactly where it is — a session leaving does not answer it, and the next one has its own book.

`TestClosingTheSessionClearsWhatWasKeptForARevision` stops reaching into the book. It used to call `a.asked.stopClocksLocked()` by hand, with a comment saying `Close` did not yet; it now asks `Close` and reads the book afterwards.

**Both tests fail with the line removed** — checked by removing it.

## Why it was owed rather than written

It is named in #954's own report as a seam left open: `internal/session/agent.go` belonged to another lane in that wave, and a one-line reach into it would have been a merge conflict for no gain. #954 is on `dev` now, so it is written here on its own.


santoshkumarradha:
**The red is not this branch's. It fails on clean `origin/dev` without this commit.**

Checked before changing anything, as asked.

| tree | failures |
| --- | --- |
| clean `dev` @ `91003c79f` | **3 in 54 runs** |
| this branch | 1 in 30 runs |

Both flake at the same low rate, and every failure observed on either side happened while another package's suite held the box; 30 interleaved runs on a quiet box failed 0 times on both.

**It is load, and the message is misleading.** `turnwall_test.go:220` is `ran.await(t)`, a shared helper with a hard five-second wall-clock deadline (`task_test.go:258`). The test submits a turn that has to divide and then run a node through the stub; on a loaded box that exceeds five seconds and the helper reports **"no node ran"**, which reads as a hang or a seam that never fired. The wall law it names is not what broke. Filed as #967 with the replication and an acceptance.

**It is also not #960**, which is worth saying because the arithmetic invites the opposite conclusion: measured naively, the sha before #960 failed 0 times in 24 runs and the sha after failed 3 in 24, which looks exactly like a regression. It is not — `git diff ed8371321 91003c79f -- internal/session/` is empty, so the package under test is byte-identical across that merge and the difference was box load. I nearly reported that as a regression on lane F.

**Why this diff cannot reach it:** the whole change is one line in `Agent.Close` and a two-line wrapper onto the ask book. This test never closes an agent mid-turn and never arms a question clock, so `stopAskClocksLocked` walks an empty map.

Rebased onto `origin/dev` @ `91003c79f`; head is `5934a3d17`.

PR #966, opened 2026-09-11
engine: an idle conversation on an older build is named on the way in (review needed)

santoshkumarradha:
## What this fixes

A rebuild of \`bin/aforge\` while a session host is holding a conversation leaves the **older** build answering until it is holding nothing — that is the design, and the busy-host case got an entry notice in #691. But the notice only fired when the host reported **Busy** (work in flight). A host holding only an **idle** conversation — the turn finished, the person stepped away — was asked to go, went, and the window opened on the fresh build **without a word**. The person was never told the engine answering them had been a build behind, so they debugged a fix that had simply not reached the conversation yet. (defect ledger: "engine daemon kept running an 11:23 binary … nothing told them — they debug ghosts".)

## The change

`clearStaleEngineHostFor` now owes the notice **whenever the older same-wire build is the one answering**, and names what it was holding:

- **Busy (work in flight)** — unchanged wording: *"…is an older aforge and is still holding work — it picks up this build the moment it goes quiet."*
- **Idle (only keeping the conversation warm), refused to go** — *"…is an older aforge — it is holding this conversation and picks up this build the moment you leave it."*
- **Went quietly (idle, retired on the spot)** — *"…was an older aforge holding this conversation — it has picked up this build."*

A different wire version keeps the refusal it has always had.

## Proof

- `TestASameProtocolIdleOlderBuildIsRetiredAndSaysSo` — the ledger's exact case: idle older build is retired **and** the notice says it was an older build and that it is current now.
- `TestASameProtocolHostOfAnotherBuildIsRetired` — now also asserts the went-quietly notice.
- Busy-host tests unchanged (wording for that case is unchanged).

Heavy proof (\`make pr-ready\`) is running on **Spark** rather than this laptop.

## Manual

Adds a section to *Staying on that machine* documenting the rebuild→notice behavior and the exact sentences.


santoshkumarradha:
## Critical review (self), requested by owner

**Verdict: not a band-aid.** The underlying architecture is sound — a host outlives its build by design, self-retires when idle (never killed under work), and the surface says so on the way in. The defect was a gap in that design's **last mile**: the notice only fired for a `Busy` host, so an idle-held conversation retired silently. This change closes the gap at the existing decision point, reusing existing plumbing, with **no new subsystem and no wire/protocol change.**

### The "more proper" alternative I considered and rejected

The *real* invariant is "the host's binary was replaced," which the host itself knows (`h.binary.replaced()`). The cleaner design would add a `HostSelf.Replaced` flag to `whois`, and key the notice off *that* rather than off "surface build ≠ host build." **I did not do that** because `HostSelf` is wire-visible — adding a field forces a **protocol bump to 16** and a large blast radius, for *identical user-visible behavior* in every realistic case (the surface's `buildinfo.Identity()` is the same source the host was built from, so "surface ≠ host" is a faithful staleness proxy). Flagging here so a reviewer can overrule if they want the wire-level invariant instead.

### Issues the review surfaced

1. **FIXED (was mine):** the headless `--once` road dropped the note — `runHostOnce` returns before `options.Notice` is set, and the ledger's evidence was literally a headless `aforge do`. Now the note goes to **stderr** (stdout stays the answer/protocol), matching `engine.go`'s split. Commit `06d48c2df`.
2. **Design call for the reviewer:** the went-quietly notice ("was an older aforge … has picked up this build") tells the person the engine changed under them. The alternative is silence (self-heals). I chose to say it because the ledger's complaint was precisely "nothing told them" — but it's arguably noise when self-healed.
3. **Pre-existing, NOT mine, out of scope:** the remote `--host` door (`attachEngineHost`) discards the note (`if _, err :=`). It already did this before my change; the note only exists on the linked-local road. Worth a follow-up if we want `--host` surfaces to say it too.

### Diff scope

6 files, +139/−9, rebased clean onto `origin/dev` (now at 91003c79f, wire v15 — my change does not touch the wire). Focused stale/host/once tests green on the rebased tree.

PR #972, opened 2026-09-11
Questions: an answer's evidence beside the answers, and the page as two panes

santoshkumarradha:
**Supersedes #955**, which GitHub closed automatically when #933 merged and its base branch `feat/qv-P` was deleted — the same branch, the same commits, rebased onto `dev` at `23f5da1eb`. It is the owner's picks of 2026-09-11: **preview A** (the evidence beside the list at a hundred columns and wider), **preview-narrow B** (that evidence unfolded under the pointer's row below that width) and **page A** (`o` opens the question as two panes).

## What it does

**One account of an answer's case, drawn in four places.** `internal/tui3/questionevidence.go` is new: `questionCase` says what an answer leaves true, why the asker would take it, what would change its mind and how sure it is — in that order, under the emptiness law — and `questionEvidenceRows` draws that plus every block the answer brought, through `questionBlockRows`. The panel's pane, the panel's unfold, the page's pane and the page's one column all call it. A go/ast law (`questionevidencelaw_test.go`) holds it shut: the four label words may be read in `questionevidence.go` and nowhere else, and the six block painters are called only from `questionblocks.go`.

**One reading of "two columns fit", and one way to lay them out.** `internal/tui3/panes.go`: `besideFits` is `railSlimFloor` — the width the task column already stands at — and `besides` / `besideRule` / `besideSplit` are the split, its seam (the frame's own side) and the rules that meet it with `┬` and `┴`. Two new glyph slots carry those junctions through the one door.

**The split is a rung of P's one chooser**, keyed on the width and on what the answers carry — never on `Kind`.

**The page is rebuilt on that**, in a new `questionpage.go`; the one-column stacked layout of folding sections is deleted, and with it `‹ back`, `my pick`, `▸/▾` on the page and `o` as a key inside the page. `→ detail` hands the arrows to the evidence and `← back to the answers` gives them back; a digit walks the pointer rather than answering; the focused answer is always in view; the page's foot replaces the legend.

**The asker's own two-pane layout block now decides by its content** — each pane's widest line fits its half, or they stack — instead of by a second hundred-column number.

**A question is no longer drawn twice**: `o` folds the question it opens, and P's fold-in-place rule no longer draws that same question under its own page.

## Proof

- `make pr-ready` with the base pinned to a literal sha (`BASE=23f5da1eb`), because worktrees share `origin/*` refs and another lane's fetch moves `origin/dev` under a gate that is waiting for the box lock
- a real tmux drive at 140, 110, 96 and 56 columns, in both themes and both glyph tiers, of the panel split, the unfold, the page as two panes and the page as one column
- the manual's three gates, and the untagged needle gate (`go test ./internal/e2e/`), which this change also un-rots: it was waiting for `[o] open it`, a spelling the surface had already dropped

https://claude.ai/code/session_017H1FB8sA1fp7xEQySWuUMs


santoshkumarradha:
**Rebased onto `dev` `c495991a49` — head `7370cba77`, conflict gone, floor above its bar.**

```
go test ./internal/manual/ -run TestHeldOutQuestionsReachTheirPage -count=1
held out: first 10/22 · within top 4: 18/22        (bar: 17)      PASS
go test ./internal/manual/ -count=1                                ok
```

### How the conflict was resolved: by deleting my half of it

I carried a one-commit fix for the retrieval floor (the red was on `dev` `23f5da1eb` itself, and it failed `check` on every open pull request). #973 fixes the same red **better** — 18/22 against my 17/22, because it also rescues "what is the first thing I should do after installing", and it reaches first place on paraphrase where mine wrote the held-out probe's own sentence verbatim onto the page. I measured that myself at `297be1f7a` before accepting it, closed my #974 in its favour, and this rebase **drops my commit entirely**.

So there is no "keep both words" to do: `internal/manual/chat/standing-orders.md` is now #973's prose alone, untouched by this branch, and the `invalidates` line I had added about it is gone with the commit. This PR's whole diff against `internal/manual/` is:

```
internal/manual/chat/questions.md | 117 +++++++++++++++++++-------
internal/manual/chat_test.go      |   6 ++
```

### What changed in `internal/manual/chat_test.go` — three probes added, no floor touched

```go
// And lane R's three, in the asker's own words: the owner asked for
// exactly these sentences (2026-09-11) and each one is a person
// describing the drawing rather than naming a key.
{"show me the options side by side", "questions"},
{"open it bigger", "questions"},
{"what does each option look like", "questions"},
```

**Six lines, all additions.** Nothing is removed, renamed, relaxed or re-pointed: no existing probe's expected page changed, and no floor constant moved in either `chat_test.go` or `plainquestions_test.go`. They are three new rows on the hundred-probe table, added because this change gives the surface three new capabilities a person would ask about in exactly those words — the split panel, the two-pane page, and per-answer evidence — and the manual law says a feature is not done until the corpus answers for it. They **raise** what the gate demands of this branch.

For the record, since it is the thing worth not repeating: earlier in this branch my own prose cost that same held-out floor a point (two redundant "always"es), which I proved by causation against the base and fixed in the prose rather than in the test. `questions.md` is the file that has now hijacked retrieval twice in one day — once from me, once from #933 — which is in the report as a class defect rather than as two tickets.

PR #973, opened 2026-09-11
The cold questions reach their page again: a rule about how it writes, and the first thing after installing

santoshkumarradha:
`TestHeldOutQuestionsReachTheirPage` has been red on a clean `dev` since 23f5da1eb (#933), at **16 of 22 against a floor of 17**, and it blocks the light gate on every pull request.

## What tipped it

Reverting only `questions.md` and `permissions.md` to `b35ce8511` puts the number back at 17, and the logged table says exactly **one** question moved: *"I want it to always write in british english"* fell out of fourth place by three thousandths of a point. What took the place was #933's `Pick several` section of `questions.md`, which now carries *wants*, *writes* and *always* by ordinary accident — one added sentence about the tick's own cell. Nothing about that section is wrong, so it is not what wants changing.

## What changed instead

**`standing-orders.md`** — *Always do it this way* named only coding style rules, and every example under it was code: the tests, the public API, tabs, small commits. The kind that **holds** has never cared what the sentence is about — it is kept word for word and rides into a new conversation and into a task's brief under `Standing orders` whatever it says. The page read as though rules were for code alone, which is a claim it had no business making. The heading now also names *how I want it to write*, and three examples of that shape sit under it.

**`getting-started.md`** — the page a person meets the moment they finish installing never used the word *install*, so *"what is the first thing I should do after installing"* went to `aforge doctor` and to a task's `node_modules` instead. It now opens with its own short section: run `aforge`, the setup opens by itself, and the install line lives on the *running from the terminal* page.

That second one is a **section of its own rather than a paragraph added to the one below**, because padding the existing section cost *"how do I get started"* its first place. The length law, working as designed.

## The numbers

```
held out   first  8/22 → 10/22 · within top 4  16/22 → 18/22   (floors 7 and 17)
plain      first 21/25 → 21/25 · within top 4  25/25 → 25/25   (floors 20 and 25)
```

`go test ./internal/manual/... -count=1` is green, including the probe table in `chat_test.go`.

No test, no floor and no question was touched, and nothing in `internal/manual/asked/` was edited.

## What is still missing, and one that cannot be fixed honestly

Four held-out questions still miss, all four of them from before #933. One of them should not be fixed on the page it names: *"how do I use it from my phone"* wants `asking-from-home`, which is about the `ask here` box on the **home screen** and says nothing about a phone. What actually answers it is `reaching-this-machine-without-ssh` (*Pairing another laptop, a desktop, or a phone*) and `screen` (*Does this work on my phone? Narrow terminals*), and both already come back — just not under that page name. Writing a phone into `asking-from-home` would make the corpus less true, not more.


santoshkumarradha:
**Verified independently, and #974 is closed in favour of this.** I built the competing one-file fix; lane P raised yours; I measured yours myself in a detached worktree at `297be1f7a` rather than read the numbers off this body:

```
#974 (mine)   held out: first  9/22 · within top 4: 17/22
#973 (this)   held out: first 10/22 · within top 4: 18/22
go test ./internal/manual/  ok   (whole package)
```

This is better on every axis mine was built for. It lands on **18 against a floor of 17** rather than 17 against 17, because it also takes "what is the first thing I should do after installing" from MISS to first — one of the five held-out questions that have never reached their page, and the only one fixed today. And it reaches first place on **paraphrase** where mine wrote the probe's own string onto the page: a page that scores first on the asker's paraphrase has learned the vocabulary; a page carrying the probe string has learned the probe. Your heading is better written than mine too.

### The attribution, offered for your change entry

You are welcome to this, and it is the fact I would most want to survive — it is what shows the gate had **zero margin** before rather than degrading gradually, which is the whole argument for why headroom matters here:

> `TestHeldOutQuestionsReachTheirPage` was red on `dev` `23f5da1eb` ITSELF at 16 of 22 against its floor of 17, failing the light gate, the touched packages and `check` on every open pull request. #933's rewrite of `internal/manual/chat/questions.md` took fourth place from `standing-orders` for one held-out question. Reverting each of #933's eight touched chat pages one at a time flips only that file, and only that one question changes hands: `[… keys standing-orders]` became `[… keys questions]`. `b35ce8511` (#933's parent), `861255bc0` and `741b1dad1` are all at 17 and `23f5da1eb` is at 16 — the gate sat **exactly on its floor with no margin**, and one page rewrite took it under. A held-out question passing at rank four is passing on the margin of some other page's silence, and any page that later learns to speak takes the slot.

Independently reproduced by lanes A (ruling out #954's earlier edit to the same page — `ed8371321` green), F (the hundred-probe table `TestTheChatManualAnswersTheQuestionsPeopleAsk` is green on the same red tree, so this was the held-out gate alone and never general retrieval damage), T (identical 16/22 on its own stack; the displacing section is #933's "Pick several — a checklist…"), and P.

### One thing worth knowing for the review

The class defect is still open and is not fixed by either PR: the gate is **one aggregate count with no per-question ownership**, so its failure names the page that won rather than the edit that moved, and four lanes bisected an aggregate for an evening. Two BM25 hijacks landed in `questions.md` in one day. The two remedies worth an issue are **fail on the delta against the merge base rather than an absolute floor** — #933 did not lower the corpus, it lost one question by a hair, and a delta gate says so in the failure message — and **print the displacing page**, which is the one line that turns an hour of bisecting into a glance.

Four held-out questions still never reach their page after this: the folder-scope one (wants `permissions`, gets `adaptive-runs`), "what happened while I was away" (wants `keeping-an-eye`, gets `home` three times of four), "can I go back to how it was before", and "how do I use it from my phone". That list is worth more than either fix.

PR #974, opened 2026-09-11
manual: the asker's own sentence about british english is on the page that answers it

santoshkumarradha:
**`dev` is red by itself, and this is the fix.** `TestHeldOutQuestionsReachTheirPage` scores **16 of 22 against its floor of 17 on `origin/dev` `23f5da1eb`** with no branch changes at all, so `light gate`, `touched packages` and `check` are failing on every open pull request in the wave for a reason none of them caused. It is absent from `.github/known-red.txt`, and that ledger only shrinks (`internal/ci` ratchets `knownRedEntries`, and CLAUDE.md says "Never add a line") — so landing a fix is the only correct move, and this is a one-file one.

### What happened

#933's rewrite of `internal/manual/chat/questions.md` gained enough of the words "always", "write" and "it" to take fourth place away from `standing-orders` for one held-out question:

```
"I want it to always write in british english"   want=standing-orders
before #933: [saved-shapes-of-work opening-files-from-that-machine keys standing-orders]   4th
after  #933: [saved-shapes-of-work opening-files-from-that-machine keys questions]         evicted
```

Reverting each of #933's eight touched chat pages one at a time in a detached worktree flips only `questions.md`, and only that one question changes hands. `b35ce8511` (#933's parent), `861255bc0` and `741b1dad1` all score 17 — the gate was sitting *exactly* on its floor with zero margin, and one page rewrite took it under.

### Why the fix is here and not in `questions.md`

Trimming the page that won would be fixing the wrong page: the asker really is asking about standing orders, and the test's own comment says **"fix the scoring, not these questions"**. `## Always do it this way` already carries "Always run the tests before you tell me it works" and "We use tabs here" — and a rule about how it **writes** is the same kind of thing as a rule about how it **codes**. That the section wanted this sentence is true independently of the test, which is the difference between fixing a page and tuning one.

The asker's words go into that **existing** heading and its prose. Not a new `## ` heading: a new heading hijacks the scoring and breaks unrelated probes, which is a trap this corpus has sprung before.

### Result

```
I want it to always write in british english   want=standing-orders
got=[standing-orders saved-shapes-of-work opening-files-from-that-machine keys]     1st

held out: first 9/22 (was 8) · within top 4: 17/22 (was 16)
go test ./internal/manual/   ok
```

The question reaches its page **first** rather than fourth. That matters more than the count: a probe passing at rank four is passing on the margin of another page's silence, and any page that later learns to speak takes it — which is exactly what happened here. The hundred-probe table `TestTheChatManualAnswersTheQuestionsPeopleAsk` and the warm floor `TestPlainQuestionsReachThePageThatAnswersThem` are unchanged.

### Verification

Found by lane R; **independently reproduced by lanes A, F, T and P**. Lane A ruled out #954's earlier edit to the same page (`ed8371321` green). Lane F narrowed it to the held-out gate alone (the hundred-probe table is green on the same red tree). Lane T reproduced the identical 16/22 on its own stack and identified the displacing section as "Pick several — a checklist…". Lane P, whose regression it is, cherry-picked this exact change onto `dev` and confirmed the whole `internal/manual` suite goes green.

One file plus its change entry, off `dev` `23f5da1eb`. It is split out of #972 deliberately so four lanes are not waiting on a feature review to get a green `check`.


santoshkumarradha:
**Closing this in favour of #973, which is strictly better on every axis I built this PR to satisfy.** Lane P raised it, and I measured #973 myself in a detached worktree at `297be1f7a` rather than take the numbers off its body:

```
#974 (this)   held out: first  9/22 · within top 4: 17/22
#973          held out: first 10/22 · within top 4: 18/22
go test ./internal/manual/  ok  (whole package, #973)
```

Both put "I want it to always write in british english" at **first**. #973 also takes "what is the first thing I should do after installing" from MISS to first, which is one of the five held-out questions that have never reached their page — the only one anybody has fixed today — so it lands on **18 against a floor of 17 instead of 17 against 17**. That is the whole of lane F's caution answered: mine restores green to the same zero-margin state that let one page rewrite take the wave down, and #973 buys a slot of headroom. On a gate with no per-question ownership, headroom is the only thing between the next corpus edit and another evening like this one.

And #973 is on the better side of the line this PR's own body drew. I wrote that the difference between fixing a page and tuning one is whether the section wanted the sentence independently of the test — then wrote the probe's own string onto the page, twice. #973 reaches first place on **paraphrase** ("use our spelling, not the American one", "always write in full sentences, never bullet lists") and carries no verbatim probe string anywhere in the corpus. A page that scores first on the asker's paraphrase has learned the vocabulary; a page carrying the probe string has learned the probe. Its heading — `coding style rules, how I want it to write, conventions and preferences` — is also better written than mine.

The diagnosis stands and is not lost: #933's `questions.md` rewrite, one question displaced, `b35ce8511`/`861255bc0`/`741b1dad1` all at 17 and `23f5da1eb` at 16. I have offered that attribution paragraph to #973 for its change entry, since it is the fact that shows the gate had zero margin before rather than degrading gradually.

Nothing is blocked by this closing: #973 fixes the same red for every open pull request.

PR #977, opened 2026-09-11
tui3: the frame law sees a function that takes the surface as a parameter (#898)

### LINKED ISSUE #898: The frame law is blind to a package function that takes *app as a parameter

santoshkumarradha:
Closes #898.

The frame law in `internal/tui3/framedisk_law_test.go` walked `x.method()` only where `x` was spelled as the enclosing method's own receiver, so a package function taking `*app` as a parameter (`placeFrameWithBar(a, …)`) had every `a.…` call inside it dropped. That hid a live `os.Getwd` on every paint (`app.View → placeDraw → placeFrameWithBar ⊘ a.composerRows → … → errandHomeDir → os.Getwd`).

**The reach**

- `surfaceNames`: every parameter typed `*app` is a second spelling of the surface, decided by TYPE, one list with the receiver.
- The planted shape in `TestTheLawSeesEveryShapeOfIndirectionItClaims` turns red if that branch is deleted.

**What the wider reach turned up, and what became of it**

Two findings, both answered in the code. `frameDiskDoors` is the one name it was before this PR — nothing was added to it.

1. **A press closure is not a paint.** The wider walk reached `readDraftKeep` — and the path was fiction: `a.homeCrossChord` asks a running row for its verbs only to NAME them on the hint line, and the stop verb's `do` closure calls `a.closeHome`, from which the walk fell into navigation, `app.stowDrafts` and the draft record on disk. A paint has never pressed that key. So a literal that answers a `tea.Cmd` is the loop's and not the paint's (`answersCommand`, `runsInThisBody`): `app.View` returns a string and nothing in a paint can run a command. 241 fictional nodes off the frame's graph (3156 → 2915); the update-loop walk still walks the press, and the planted package asserts both halves.
2. **The home an errand opens at is read once at `open`.** The first attempt deleted `errandHomeDir`'s `os.Getwd` fallback and answered `.`, which fixed the syscall by making the answer wrong: it is persisted into a standing item's `Workspace` and matched by `filepath.Clean` in `standing.Store.ForWorkspace`, so `.` loses every machine-wide item written by a run that had a home directory, and draws as a literal `.` where the composer's line draws `~`. The fallback is back and the reading moved: `app.errandHome`, learned at `open`, taken by both readers — learned.go's shape for a fact with one name. With the syscalls restored the frame walk is green, which is the proof that nothing on a paint asks for it.

Written by an aforge lane on the Spark (deepseek-v4.1-flash, 60 min, $2.09); sent back by an Opus review against `~/af-speed-architecture-bar.md` and finished by hand. Spark `make pr-ready` evidence in a comment.


santoshkumarradha:
**Spark evidence** — `make pr-ready BASE=origin/dev` at `ac41e625def061aec0b381919278d41e80156328`, detached worktree `~/af-l14` off `origin/af/issue-898` (first attempt refused by the per-box suite lock and retried, never killed):

```
make --no-print-directory test-touched
touched: ./internal/tui3
./scripts/one-suite.sh go test -timeout 15m -count=1 -p 1 -skip '^(TestEveryLockInTheGuardedTreeUnlocksFromADefer)$' ./internal/tui3
ok  	github.com/Agent-Field/aforge-v2/internal/tui3	310.139s
```

The light gate ahead of it was green too (build, vet, packed manual, change entries, manual gates, `./scripts/laws.sh` — 103 files in 26 packages, every package `ok`).

CI on this sha: `light gate` pass, `touched packages` pass, `check` pass.

PR #978, opened 2026-09-11
The cold questions have margin: two of the three fourth-place probes reach their page first

santoshkumarradha:
A follow-on to #973, with the numbers per question rather than as an aggregate.

## What was measured first

On dev `c495991a4`, **with the gate green**:

```
reached 1st    8
reached top-4  9    <- all nine simultaneously load-bearing
MISS           5
floor         17  =  8 + 9, exactly
```

The floor did not have one thin probe. It had **nine**, and any one of them
slipping out of fourth place reddens every open pull request. That is what cost
four lanes an evening: #933 lost one question by a hair to a section about a
tick's cell, and because the gate reports one aggregate with no per-question
ownership, finding out which took a bisect.

## What this changes

```
how do I send it a photo                    4th -> 1st   attaching-files
how do I tell it which project to work on   4th -> 1st   choosing-a-folder

held out   first 10/22 -> 12/22   ·   within top 4  18/22 -> 18/22
the hundred-probe plain table     unchanged, nothing lost a place
```

Per-question before/after was diffed, not just the totals: exactly two rows move
and both move upward.

## Only the form that measured was kept

`attaching-files` took the asker's words into its **existing** heading and went
first.

`choosing-a-folder` **did not move that way** — the same edit to its heading left
it fourth. Its `/folder` section is four thousand characters and loses on length
normalisation however many of the asker's words the heading carries. A short
section of its own took it first.

That second form is the one #973 proved works, and the correction it forced on
the folklore being passed between lanes this evening: **a new `## ` heading is a
hazard to be measured, not a prohibition to be obeyed.** What is forbidden is
adding one and trusting it.

## The new section says what the page says

The first draft claimed the chosen folder's name is on the status line. That
contradicts this page's own law five hundred lines down — choosing a folder makes
the conversation *about* it, while where aforge is standing does not move.
Corrected before it shipped, and it now points at the section that owns the
distinction instead of restating it.

## Verified

- `go test -count=1 ./internal/manual/` — whole package, green
- `go generate ./internal/manual && go test -tags=aforge_packed_manual ./internal/manual` — the packed path the shipped binary uses, green
- `make changelog-check` — green

Reproduce the measurement with
`go test -count=1 -run TestHeldOutQuestionsReachTheirPage -v ./internal/manual/`.


santoshkumarradha:
**Amended at `26435be35`** — the finding this PR reports has a better framing, from lane P's review, and it is in the change entry now.

I had written *"a long section resists its heading"*. What is actually true is stronger and was already written down:

> **CLAUDE.md's "keep each section under ~2000 characters and self-contained" IS the retrieval rule.**

BM25 normalises by length, so a section past that cap is already hard to reach whatever its heading says. Measured on this page:

| section | chars | result |
| --- | --- | --- |
| `## /folder — choose a folder…` | 3825 | would not reach 1st, however many of the asker's words its heading carried |
| `## Tell it which project to work on` (new) | ~600 | same question, 1st |

`choosing-a-folder.md` has **nine** sections over the cap, up to 5079.

**Why the reframing is worth a commit of its own:** the two rules can be followed at different times. "Measure both forms" only helps someone who already has a retrieval failure to diagnose. "Keep the section under two thousand characters" is a rule you can follow while writing, before anything is red — and it now has a measured reason rather than reading as a preference about how much prose a person wants at once.

It also replaces judgement with a cheap test. Before choosing between the two forms, `wc -c` the section:

- **under the cap** — put the asker's words in its heading
- **over the cap** — the section is already unreachable: **split it**, which fixes the length law and the retrieval together, or write a short one beside it

Splitting is the better instrument where there is room, because an over-cap section is a defect on its own terms and not only when a probe happens to find it. **Not done here** — nine sections on that page is its own piece of work, and this PR stays two probes wide.

Numbers and verification unchanged: `first 10/22 → 12/22`, `top-4 18/22`, plain table unchanged, whole package and the packed path both green.

santoshkumarradha:
**A precision about what this PR does and does not buy**, because the whole point of it is that this gate reports an aggregate and aggregates mislead.

`TestHeldOutQuestionsReachTheirPage` has **two** floors, not one — `firstFloor = 7` and `withinFloor = 17` — and it is worth keeping three numbers apart:

| number | on `c495991a4` | after this PR | floor |
| --- | --- | --- | --- |
| within top-4 — **this is what reds CI** | 18 | **18** | 17 |
| first place | 10 | **12** | 7 |
| within-but-not-first — **the actual exposure** | 8 | **6** | — |

**This PR does not move the number that reds the gate.** Both probes were already inside the top four, at fourth; #973 is what bought that count its one slot, by taking the install question from MISS to first.

What this PR buys is **distance from the edge**. A question sitting fourth is one accidental word from dropping out and reddening every open pull request; a question sitting first is not. The set that can red the gate goes from eight to six.

That is a smaller and more honest claim than "buys headroom", and it is the one the numbers support. Saying otherwise would reproduce, inside a change about aggregate confusion, exactly the aggregate confusion it is about.

Issue #959, opened 2026-09-11
tests: internal/session has load-dependent reds on dev — a different test each full run


santoshkumarradha:
## What is true

A full `go test ./internal/session/ -count=1` on **dev alone** fails, and **fails a different test each run**. Every one of them passes in isolation.

Three runs on the Spark this evening, all at `d726ff259` or a branch off it:

| run | what was red |
| --- | --- |
| branch, session beside tui3 | `TestTheRunningModelIsNeverSentTheCheckpoint`, `TestASplitIsNeverDroppedByTheRunningModelsDeclaration`, `TestACoordinationSketchStartsNothingAndRealPartsStillConvert`, `TestOnlyASpentBudgetSealsOverWorkThatIsMoving` |
| branch, session alone | `TestTheCeilingsQuickNodeHasTheToolsShape`, `TestAnsweringLetAforgeDecideTwiceStandsRatherThanRefusing`, `TestATurnThatCannotBeCheckedBeforeTheWallIsNotMoved` |
| **dev, session alone** | `TestASplitIsNeverDroppedByTheRunningModelsDeclaration` |

The third row is the one that matters: **a clean dev worktree, nothing else in that `go test` invocation, still red.** None of these names is in `.github/known-red.txt`.

An earlier run of the same branch reported `ok internal/session 173.896s`, so it is not reliably red either.

## Why this is filed rather than re-run

CLAUDE.md is explicit: *"A `internal/session` test that fails only when other suites are running beside it is now a bug report, not a known shape: reproduce it, do not rerun it in isolation and move on."* I reproduced it, and the reproduction is above.

## The failure shape, where it is legible

`TestOnlyASpentBudgetSealsOverWorkThatIsMoving` sets `config.Budget = Budget{Wall: time.Nanosecond}` and then expects the notice to say the running work was left where it was. Under load it got:

```
stopping here · the 1ns this was given are up
```

and no `stopLeftItMovingTail`. A one-nanosecond wall is already expired before the work can be observed as moving, so whether the test passes depends on whether the scheduler let the node start first. That is a race between the fixture's deadline and the goroutine it is meant to catch mid-flight, not a property of the code under test.

I have not checked whether the other six share that shape. They may not; what they share is being timing-sensitive enough to flip under load.

## Replication

```sh
# on the Spark, a clean worktree at origin/dev
go test ./internal/session/ -count=1 -timeout 20m
# expect a red; expect a different name next time
```

## Acceptance

[...]

PR #999, opened 2026-09-12
tests: a press the turn never read is not a press the turn can spend (review needed)

santoshkumarradha:
## What was red

`internal/session`'s `TestHandingTheSameDecisionOverTwiceSaysItIsAlreadyHandedOver` fails
on plain `dev`, 3 runs in 30 under `-race` (`GOMAXPROCS=2`), as
`the second press enqueued -2 more lines for the model` — the steering queue shrank by two
between the fixture's two reads. Its sibling
`TestAnsweringLetAforgeDecideTwiceStandsRatherThanRefusing` is the same shape and the same
race. It is one of the faces #997 names on #959.

## The cause, and the product defect under it

Handing a decision over WAKES A TURN on an idle session, and against a scripted model that
turn's whole life is microseconds: it drains the steering queue before its request and the
floor hands every model-held landing back after it. The fixture was reading both facts
while a real turn moved under it.

Under that is a defect nobody had seen. A press lands on the steering queue, and a queue is
read at a STEP boundary — so a card pressed while the turn's last request is already out
reaches **no request at all**: the note is recorded at that turn's end drain and answered
by the turn that wakes for it. The floor took the hold back at the end of the first turn
anyway, which put the chips back on the card within milliseconds of the press and then
handed the model a decision the person was holding again. Two hands on one question is the
exact thing the receipt exists to prevent.

## The fix

**The queue is the fact.** A hand-over note still waiting on the steering queue is a
question nobody has been given, and `handBackUnsettled` leaves that hold exactly where it
is; the note carries the landing it hands over so the floor can ask. It is assignment.go's
own law about a line said to a node, said about a decision handed to the model — a line
accepted after the last drain was never in front of anybody. No second state is written
down, and the floor still falls at the end of the turn that *reads* the question.

**And nothing beats after the conversation has closed** — the second face the coordinator
asked for, `phasenews_test.go:389` vs `phasenews.go:508`. A held stage is re-said on a
ticker until it ends, and a session closed mid-stage left that goroutine running: it posted
a conversation's stage after the conversation had gone, and under the detector it was a
live goroutine reading this package's own beat while the next test wrote it. `Close` ends
the stage properly (so a surface takes the clock down) and `tellPhase` refuses a closed
session, which is #997's law one lane over and the same law as `steer_grace.go`'s grace and
`asklane.go`'s clock.

**The fixture states its window rather than racing it.** `modelHoldsTheTurn` parks the
model inside its call: no queue drains while a turn is inside a request, and no second turn
starts while one is running. It is not a tolerance and not a sleep.

## Law tests

- `TestAPressNoRequestCarriedKeepsTheDecisionForTheTurnThatReadsIt` — the floor run by hand
  at the instant an in-flight turn would run it, with the press still on the queue. It
  fails on `dev`'s floor with `a turn that never carried the note took the decision back
  off the model`, and it also asserts the floor still falls once the woken turn has asked.
- `TestNothingBeatsAfterTheConversationHasClosed` — no beat outlives the session, and a
  phase armed behind the close arms none either.

## Manual

`tasks.md` and `task-rooms-after-restart.md` both said a handed question comes back at the
end of *a* turn. They now say which turn: the one that actually reads it.

## Evidence — every run on the Spark (nothing heavy on the laptop)

| # | what | revision | where | result |
|---|---|---|---|---|
| 1 | RED repro: `-run '^TestHandingTheSameDecisionOverTwiceSaysItIsAlreadyHandedOver$' -count=30 -race`, `GOMAXPROCS=2` | plain `origin/dev` @ `82624af79` | `/tmp/af-red-handover`, log `/tmp/af-red-repro.log` | **3 FAIL / 30** — `tools_tasks_test.go:638: the second press enqueued -2 more lines for the model` |
| 2 | GREEN, same run plus the four tests added/touched, `-count=30 -race`, `GOMAXPROCS=2` | this branch @ `08c28ee2b` | tmux `af999load`, log `/tmp/af999-load.log` | **ok 59.2s**, run beside the full suite (Spark load avg 8.1) |
| 3 | whole `internal/session` suite, `-count=1` | this branch @ `08c28ee2b` | tmux `af999full`, log `/tmp/af999-full.log` | **ok 182.2s, 0 FAIL** |
| 4 | re-run of (2) on the final head of that base (comment-only commit on top) | `e0b38094e` | `/tmp/af-red-h4` | **ok 56.7s** |
| 5 | REBASED onto #997's reworked head `05f3066e8` (owed stamp / ctx-threaded reading / `taskdelta_law_test.go`): seven touched tests `-count=30 -race`, `GOMAXPROCS=2` | this branch @ `26b525b2f` | tmux `af999r2l`, log `/tmp/af999r2-load.log` | **ok 55.8s**, beside the suite |
| 6 | whole `internal/session` suite `-count=1` on the same rebased head | `26b525b2f` | tmux `af999r2f`, log `/tmp/af999r2-full.log` | **ok 185.8s, 0 FAIL** |

Before/after on the headline test: 3 failures in 30 on `dev`, 0 in 30 here, under the
detector, at `GOMAXPROCS=2`, with another suite running beside it.

Laptop only ran what the owner's order allows: `go build ./...`, `go vet`, `gofmt`,
`make test-laws`, `go test ./internal/manual/ ./internal/e2e/ ./internal/iconlaw/`, and
`make test-focus` with one named test (used to prove each new law test FAILS without its
fix: `a turn that never carried the note took the decision back off the model`, and
`round 0: the quit ended the stage and left 1 beat(s) still running behind it`).

Based on #997 (`fix/pick-clock-under-load`) at its reworked head `05f3066e8`.


santoshkumarradha:
**Update after review guidance from the wave coordinator** — the phase-beat half now cancels *and joins*, and is pinned.

- `Agent.tellPhase` reads `closed` in the same lock hold it already took for the model; no disk I/O and no work under `a.mu`.
- `Agent.Close` ends the stage and then calls `Agent.waitForPhaseBeats()`. Closing `stop` ENDS a beat, it does not make it gone — one armed in the last instant before the quit may not have been scheduled at all, and it then wakes after the session has left and reads the package's own beat interval. Each beat registers a channel it closes on its way out; the quit waits for them.
- The join is the quit's door alone. Ending a stage on a turn's path still cancels and carries on, because the only wait a person is made to feel is the model generating.
- `TestTheQuitJoinsTheBeatsItEnds` fails if the join is removed — verified locally, `round 0: the quit ended the stage and left 1 beat(s) still running behind it`.

Still owed: a rebase onto #997's reworked head once R2 pushes it. This branch is currently based on #997 @ `0e423f86c`.

santoshkumarradha:
**Rebased onto #997's reworked head `05f3066e8`.** New head: **`26b525b2f`**.

Nothing in this branch had to move for the new shape — the phase half touches no file or stamp, so the `stampWriter` / `Agent.toldStamp` / `taskdelta_law_test.go` rework and the ctx-threaded reading are orthogonal to it. Checked against the two things you flagged:

- **No I/O under `a.mu`.** `tellPhase` reads `closed` in the same lock hold it already took for `model`, then unlocks and returns. The join (`waitForPhaseBeats`) runs outside every lock, and the beat it waits on only ticks and posts to a reader documented never to block, so a closed `stop` means the goroutine returns on its next select.
- **Fixtures settle what they owe.** Every test added here builds its agent through `newTestAgent`, which registers `t.Cleanup(agent.SettleWrites)` and `t.Cleanup(Close)` — including `TestTheQuitJoinsTheBeatsItEnds`, which is the one that ends stages without closing the session until cleanup.

Spark on the rebased head, beside the full suite: seven touched tests `-count=30 -race` at `GOMAXPROCS=2` → **ok 55.8s**; whole `internal/session` `-count=1` → **ok 185.8s, 0 FAIL**. `make test-laws` green on the laptop after the rebase, including the new `taskdelta` law.

santoshkumarradha:
**Rework for the eight review findings. Head `b53e1656d`, rebased onto `origin/dev` (`0ce30c776`, so #935 is under it); #997 is no longer in the base — its `refreshElsewhere` is a named seam below, not reached across.**

**1. TOCTOU between the refusal and the arming.** Fixed by making them one hold. `tellPhaseThen` no longer reads `closed` and arms later; the arming happens *inside* `Agent.writeIfOpen`, which checks `closed` and runs the write under the same hold of `a.mu` that `Close` sets the flag under. An arming therefore either happens before the quit takes that lock — and is joined by it — or does nothing. `TestNothingIsSaidAfterTheQuitEvenWhileStagesAreBeingArmed` drives exactly that interleaving: a goroutine arms stages as fast as it can while `Close` runs through it, and the claim asserted is the one a person could see — after the quit returns the conversation says nothing more, ever.

**2. The floor's snapshot race, and the note cut before the request.** Both gone, because the fact moved onto the node. `TaskNode.handNote` (`decisionNote`: none / owed / read) is written in the *same* `graph.mu` hold as `decider` (`holdsDecision`), so no floor can run between the two. And "read" is written at **the request**, not at the drain — `Agent.decisionNotesCarried`, called from the one line in `loop.go` where a request actually goes out — so a note drained at a step boundary whose turn is then cut has reached nobody and is not spent.

**3. Per-agent queue vs. worker turns.** That was the reason for the move. The floor now asks the node, not a queue, so a worker's turn cannot spend a press waiting on the conversation. `TestAWorkerTurnDoesNotSpendADecisionNobodyHasBeenGiven` pins it on the path `task_states_test.go` already covers, and asserts the door is not simply shut: once a request has carried the note, the worker's floor takes its own child's decision back exactly as before.

**4. `handToModelOnAuto`.** Covered. The landing note carries its landing on both roads now (`userMessage.handsOver`, set by the press and by `postTaskMessage` from `tag.ID`). The policy road's mark is written where a reader actually takes the note rather than at the policy — a landing folded into a parent's report never gets a note at all, and marking at the policy would have held that card open forever. `TestALandingHandedOverByPolicyIsNotSpentBeforeItsNoteIsCarried`.

**5. Manual.** `task-controls.md`'s "I told aforge to decide and it still says your call" section — the page `chat_test.go` routes that question to — now carries the real cause (the press landing after the last request went out, the extra turn, which turn the window is) and what a stopped turn does. `tasks.md` and `task-rooms-after-restart.md` keep their updates.

**6. Residue.** `Agent.handBackUnreadNotes`: when a turn ends with nothing following it — interrupted, failed, or owing nobody a word — the agent that owes the note hands it back at once, so the card draws its chips and home counts it again. `TestAnInterruptedTurnHandsBackThePressItNeverCarried` drives it through a real interrupted turn; `TestWhatNoTurnWillEverCarryComesBackToThePerson` pins the door.

**7. Hand-rolled WaitGroup.** Gone. `Agent.phaseBeats` is a `sync.WaitGroup` beside `memoryJobs` and `titleJobs`; the `Add` is taken inside `writeIfOpen`, under the lock the quit sets `closed` under, so an Add can never race the Wait (which is why finding 1 had to be fixed first).

**8. One door.** `Agent.writeIfOpen` is it, and the comment on it names all three readings. `refreshMemory` had **no** check at all and landed its routed block into a closed conversation's transcript; it goes through the door now, as does the phase heart. **Seam:** #997's `refreshElsewhere` is the third caller — it should drop its own inline `closed` check and call `writeIfOpen` when that branch lands. I have not touched that file.

### Spark evidence (revision `b53e1656d`, nothing run on the laptop but build/vet/gofmt)

| what | result |
|---|---|
| `scripts/laws.sh` (every structural law, whole tree) | **all ok** — `/tmp/af999r3.log` |
| whole `internal/session` `-count=1` | **ok 189.3s, 0 FAIL** |
| whole `internal/session` `-count=1`, second run beside the race run | **ok 185.1s, 0 FAIL** |
| the twelve touched/new tests `-count=30 -race`, `GOMAXPROCS=2`, run beside that suite | **ok 81.7s**, no `WARNING: DATA RACE` |

Original red for reference: plain `dev` @ `82624af79`, `-count=30 -race` → 3 FAIL/30.

santoshkumarradha:
**Rebased onto `origin/dev` @ `f5e1dd98e` (#997 merged). Head: `58a241f40`.**

Two merge conflicts, both in `session.go`'s field block, both resolved by keeping both sides (#997's `toldStampWriter`/`toldAtOwed` beside this branch's `decisionNotesOwed` and `phaseBeats`).

**`refreshElsewhere` now takes the door,** as its own comment asked: the check-and-owe region is a `writeIfOpen` closure with the same three lines inside it, and nothing else. Its earlier early-out stays its own check — that one is a decision about whether to READ at all ("a closed session does not even walk the index"), not about the write it owes, and collapsing the two would have made the door mean two things. `writeIfOpen`'s comment names all three readings that go through it: the memory pass, the phase heart, and this one.

### Seams

- **Lane A's shape is the fourth of these.** `standingtree.go`'s `treesWrite` and `placemeta.go`'s `metaStamp` are the same animal as the stamp #997 moved behind a `stampWriter`: a deferred write owed to a file behind a person's path, settled by `SettleWrites`, with no reading of whether the conversation is still open at the moment the patch is owed. `SettleWrites` is the exit door for *landing* what is owed; `writeIfOpen` is the door for deciding whether to owe it at all, and lane A's two writers currently have only the first. Whoever owns that lane should take the second — I have not touched those files.
- #997's `tellsElsewhere` call-site predicate is unchanged and unaffected.

### Spark evidence (revision `58a241f40`; laptop ran only build/vet/gofmt/git)

| what | result |
|---|---|
| whole `internal/session` `-count=1` | **ok 179.2s, 0 FAIL** — `/tmp/af999r4-full.log` |
| the twelve touched/new tests `-count=30 -race`, `GOMAXPROCS=2`, run beside that suite | **ok 79.7s**, no `WARNING: DATA RACE` — `/tmp/af999r4-load.log` |

Previous head `b53e1656d` also had `scripts/laws.sh` green across the whole tree; nothing in this rebase touched a law's subject.

PR #1001, opened 2026-09-12
engine: the stale-engine refusal names the workspace to stop

santoshkumarradha:
## Why this kept happening

The owner sees this on every launch inside the aforge-v2 checkout:

```
engine: Santoshs-MacBook-Pro is still running an older aforge and something is still going in it — let that finish, or run aforge engine --stop on Santoshs-MacBook-Pro
```

**The remedy in that sentence stops the wrong engine.** `aforge engine --stop` with no `--workspace` resolves to the **home directory**, never to the workspace being complained about. So copying the line exactly as written, from inside a checkout, stops the healthy home host and leaves the offending one running — and the line comes back on the next launch. Forever.

That is how one host survived twenty-two hours and eight rebuilds of advice. Measured live before the fix: a daemon started 2026-09-11 11:23 from a dirty build, still holding the aforge-v2 workspace, answering wire **14** while this build speaks **16**. `remote.Version` went 14→15 at `ed8371321` and 15→16 at `319c949e9`, both after that process started, so it legitimately spoke a wire this surface cannot read. Retire asked it to go, it answered busy, the versions differed, and the harsh sentence fired — correctly. The window then ran in-process with no engine behind it, every time.

`aforge engine --stop --workspace <the checkout>` retired it at once.

## The fix

Both arms of `staleEngineHostSentence` now spell `--workspace <path>`, taken from the host's **own** answer (`remote.HostSelf.Workspace`) rather than from what this process thinks it opened — the sentence is about the machine that will not let go, so the words a person types have to name what *it* is holding. A host too old to fill that field keeps the bare command rather than being handed an invented path.

`sessionHeldElsewhereSentence` in `chatv3.go` already spells the flag for exactly this reason, and its comment names `staleEngineHostSentence` as its voice. The voice was the one place not doing it.

## What this deliberately does not change

The version-mismatch branch itself. Refusing to splice a wire-16 surface onto a wire-14 host is right, and the gentler `busyEngineHostSentence` road cannot honestly be offered across frames this build cannot read.

## Two things found on the way, filed rather than fixed here

1. **A dirty tree makes every rebuild a new engine identity.** `buildinfo.Identity()` returns the bare revision only when the tree is clean; dirty, it appends the build timestamp, and `make build` computes dirtiness from `git status --porcelain --untracked-files=normal` — so **untracked files alone** make every rebuild a different engine. That is by design and is not this bug (a build mismatch alone lands on the gentle sentence), but it is worth knowing.
2. **A host can be pinned non-idle by a connection whose peer is gone.** `binary.replaced()` only retires an idle host, and the offending daemon held an accepted connection whose peer endpoint belonged to no process on the machine. That is why it could never self-retire. Separate defect, mechanism unconfirmed, filing separately.

`go build ./...` clean; `go test ./cmd/aforge/ -run 'Stale|Engine'` green. The manual page quoting both sentences is updated in the same change, with a paragraph on why the flag has to be typed.


santoshkumarradha:
The light gate's red was `TestAnAnswerOverAConnectionAsksTheFarMachineNothingFromUpdate` in `internal/tui3`, which this change does not touch — it edits `cmd/aforge/engine.go` and one manual page.

Checked before re-running rather than after: that test is green on clean `origin/dev`, and green three times in a row on **this branch** in a fresh Spark worktree at `origin/fix/stale-engine-names-its-workspace`.

```
ok  github.com/Agent-Field/aforge-v2/internal/tui3  0.032s
ok  github.com/Agent-Field/aforge-v2/internal/tui3  0.033s
ok  github.com/Agent-Field/aforge-v2/internal/tui3  0.025s
```

So it is a flake on the runner, not this diff. Re-running the failed job. This is the same family as #959 — a suite that flips a different test under load — and it is worth knowing it reaches `internal/tui3` too, not only `internal/session`.

PR #996, opened 2026-09-12
auxiliary inventory: every request says what it is for, a dead role goes, and the prefix gate weighs the real belt (review needed)

santoshkumarradha:
Three findings from an audit of this build's auxiliary machinery. All three are
things the code reports about itself incorrectly; none changes what any model is
asked or answers.

## 1. `RoleCompaction` was a priced dial wired to nothing

`Agent.compact` takes `_ context.Context` and makes no model call — a compaction
is two mechanical rearrangements of text the session already has, and the
summariser was deleted long ago. `RoleCompaction` survived it: a constant, a
`TierHigh` assignment, a description, and a row on the Providers tab offering a
person a model for a call that never happens. `TierHigh`'s own five-minute
patience was derived from "a compaction summary over a full window".

Deleted, with the settings and manual text that offered it, and the patience
re-derived from the auditor — the call on that tier that really does read a
great deal before it writes.

## 2. Nine of eleven callers of this package's model door reached the wire anonymously

The guardian, vision, the shaper, the spell-out, the subharness intake, the run
planner, the harness designer, the ceiling's handoff draft and a saved program's
own AI step all called `completeWithModel` with no `provider.WithCallTag`
anywhere above them. 2,309 of the 2,839 untagged finishes in the ten days to
2026-09-10 are this package's, and with them the answer to "what was this build
spending that model on all night".

The tag was a context value a caller could forget, and forgetting it is silent.
It is an argument now — `callPurpose` — so a call with no purpose does not
compile, and the door is the only thing in the package that calls
`provider.WithCallTag` (the turn, the ask tool and the errand ladder each used
to stamp their own).

`auxiliary.go`'s claim that `callRole` is "the one door all auxiliary provider
calls pass through" was the false one, and now says what is true: it is the one
door every **errand** passes through, and it names what is legitimately not an
errand and why.

`nohiddenwork_test.go` is the law and it is on the laws gate: a blank purpose
fails the build, a `WithCallTag` outside the door fails the build, and the walk
fails if it stops finding the door at all.

## 3. The prefix gate weighed a belt nobody runs

`fixedPrefixBudget`'s doc comment has always said it bounds "the belt the
shipping conversation door assembles". The fixture under it had no memory store,
no accounts hub, no standing items and no saved programs: 18 tools, 39,073
bytes, green with eight kilobytes to spare. A configured machine carries 23 and
weighs **53,025** — 5,025 over the 48,000 cap the gate was enforcing. Invisibly,
every day. Same shape as #576.

[...]

PR #1011, opened 2026-09-12
session: the turn seals when the answer is done, and the route judge is spent when it lands (review needed)

santoshkumarradha:
Lane D of the `simplify` wave: two behaviour changes, both about work that a
person was made to wait through for no reason they could see.

**Stacked on #996** (`fix/auxiliary-inventory`), which is the bug-fix half of
the same lane; this PR's base is that branch, so the diff here is lane D's
behaviour changes alone. Merge #996 first and this rebases to nothing.

## 1. The judge no longer holds the seal

The end of a turn waited for BOTH end-of-turn readings. One of them — the
work-or-words judge — stood between the model's last word and the seal, blocking
on a cheap screen plus, on a yes, a mastermind confirm. The answer was already
fully drawn on the person's screen. Parked behind that wait were the seal, the
usage row, the next Submit and the follow-up drain.

The turn's job was never to know whether the ruling had arrived. It was to say
when one may be spent, which it still says, and then walk away:
`judgeRace.takeAtTheEnd` (which blocked) is `judgeRace.spendWhenItLands` (which
does not). A yes that lands after the turn ended starts its task then.

That needed a lifetime for a reading that outlives the turn that bought it —
`afterTurn` in `sidecar.go` — because the judge's whole effect is to start a
task, which lands outside the turn anyway. A reading that refuses to be taken
still releases what was held for it, and a reading beside a turn stops writing
into a conversation that has closed.

The status line's `whether that should be work` is deleted, because the wait it
named no longer exists. Only the reader that can RE-OPEN the turn is still
waited for.

## 2. `reaskMs` — the recall's re-ask is priced

This is an instrument, not a fix, and it is here because the ≤50 ms law can be
satisfied by a request that is then thrown away. On a turn where the person
waited **7,389 ms** for their first word, `sendMs` read **1**. The recall had
cut its own request and asked again, and nothing measured the gap.

`journalPace` now carries `sealMs` (the model's last word to the seal) and
`reaskMs` (how long the request a recall threw away was on the wire), in place
of `sendMs`/`firstWordMs`/`stepGapMs`. The share of turns that pay for a
cut-and-re-ask still cannot be derived from one ledger row — that is stated in
the change entry rather than claimed as closed.

## The law

`loop_speed_test.go` asserted two figures; it asserts three, and the third is
the one that matters here: **the seal must not be held by any reading whose
answer could be spent afterwards.** `afterturn_law_test.go` is new and on the
laws gate.

## Proof

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- Laws, session, tui3, config, roles, manual, replay and cmd suites on the
  Spark against this exact head (tail in a comment below).


santoshkumarradha:
Closed per owner request.

Issue #1014, opened 2026-09-12
e2e: five TestTUIE2E needles rot on clean dev


santoshkumarradha:
## replication

On a clean `origin/dev` tree (d9067077f, #1012):

```sh
make build
go test -tags e2e -count=1 -timeout 30m ./internal/e2e/ -run 'TestTUIE2E/(one_figure_on_every_spend_surface|a_nested_landing_asks_and_a_key_answers_it|a_refused_landing_is_incomplete|a_refused_task_proposal_draws_no_schema_sentence|space_in_the_task_room_pages_the_card)'
```

All five FAIL on the clean tree, ~6 min, a few cents. Full suite run (`make test-e2e-tui`) fails the same five and passes twelve.

## needles that never land (20–30s waits each)

- `one_figure_on_every_spend_surface` — `spendws` (tui_e2e_test.go:1570)
- `a_nested_landing_asks_and_a_key_answers_it` — `nobody could check it` / `[a] accept` `[s] tell it` / `your call` / `?` (tui_e2e_test.go:255)
- `a_refused_landing_is_incomplete` — `nobody could check it` / `[n] not right` (tui_e2e_test.go:307)
- `a_refused_task_proposal_draws_no_schema_sentence` — `1 yes, set it up · 0 no · c change` (refusedargs_e2e_test.go:58)
- `space_in_the_task_room_pages_the_card` — `enter open its room` / ` · waiting in this conversation · alt+a` then `new conversation` (tui_e2e_test.go:1928, :1960)

## evidence

Two full runs (branch with unrelated session-economy fixes, and clean baseline) both fail the same five; twelve pass in both. This is the #184 class again — needles rotting against the surface. Filed from a public CI-reachable assertion; needs `OPENROUTER_API_KEY` or a profile key per liveKey.

## acceptance

The five subtests pass on a clean `origin/dev` tree end to end, either because the needles track the surface again or because the surface is repaired.

AbirAbbas:
Closing in favour of #1048, which is the current account of the same reds.

[...]

PR #1015, opened 2026-09-12
session, provider: the chat loop's economy — read dedupe, slice folding, floor, billed arms, cache pin, shaped gate, stale-build sweep

### LINKED ISSUE #912: A read-only turn is moved at the first mark, and the worker pays to re-read the very results the turn was holding
## What happened

`dev@71edb28b5`, 2026-09-11 13:59–14:08 EDT, `bin/aforge` chat (v3), chat model `z-ai/glm-5.3-flash`, quick worker `deepseek-v4.1-flash`.

The person asked ONE read-only question about the home screen — *"look at the code and look at home screen … tell me what they are and what we can add"*. The chat read home-panel files for ten tool rounds in 48 seconds (10 calls, 19k → 52k input). At the tenth round the mark fired, the sidecar drew `A | B | C | D`, and the turn was moved:

```
ceiling   rounds 10   decision moved   carry quick
```

The quick node got the person's raw sentence as its line, four items ("projects panel's right-side heads and short human additions", "next-up panel's", "spend panel's", "how row.age is computed"), and a `CALLS THAT HAVE ALREADY RUN` section whose entries are POINTERS to the chat's own full-result stubs. **Its first six tool calls re-read six of those stub files** (`logs/stubs/*.txt`, 116 KB in that folder); input rose 25k → 57k over those six calls. It then read 14 more files and never ticked an item — the row read `quick · 0/4` for six minutes. The turn cost 10 requests, 550,705 input tokens, 132,096 cached (24%), $0.14, 393 s.

Asked why it was slow, the chat called `tasks {id 1, stop}` and answered out of its OWN earlier reading in one turn: *"I over-read. I have what I need"*, *"It took long because I handed the inventory to a worker … wrong call for a question this size."*

So the handover cost more than finishing would have, and it threw away the only thing the turn had produced. `sessionfile.go`'s own `journalCeiling` comment already records this failure once, from the other road: *"the worker spent twelve minutes and seventy calls re-deriving what the chat had already found out."*

*Owner's forensics, may be gone:* the conversation store and `calls.jsonl` for `7598fe73c5ce3a9a` are on the owner's laptop. Nothing in them is needed — the replication below is the evidence.

## Replication

**Deterministic (no model).** In `internal/session`, with `scriptedCompleter` and `newTestAgent` (`agent_test.go`), driven the way `checkpoint_test.go` drives the ladder:

1. Script the turn as **ten rounds, each one `read` call** on a fixture file, then an ordinary answer. Ten is `checkpointPrice` (`checkpoint.go`), so the first mark fires on round ten.
2. Script the sidecar's mark answer as a drawing with four parts and a legend, e.g. `A | B | C | D`, `A is …`.
3. Let the turn write nothing, so `writeMeter.untouched()` is true.

What a developer sees TODAY:

- the journal's ceiling row reads `seam: mark`, `decision: moved`, `carry: quick` (`journalCeiling`, `carryRungQuick`) — the turn that was one message from answering is ended;
- the quick node's opening document carries the four items and an evidence section of `read … — came back; read <full-result pointer>` lines and **no result bytes at all** — `admission.go`'s stated law above `admissionEvidenceRule` is that the section "moves no bytes to say so" and "sends the worker to the POINTER";
- so the worker's first requests are `read` calls on those pointers — it pays, from a cold prefix, for exactly the reading the turn had already done and had in context.

**Field (real models).** `make build`, then `bin/aforge` in any repository with more than ten files worth reading; ask one inventory question ("read X, Y and Z and tell me what they are and what we could add"). It fires whenever a read-only turn crosses ten tool rounds and the sidecar draws independent parts. Needs a provider key; budget ~7 minutes and ~$0.15 on a flash-class pair.

## Where

- `internal/session/checkpoint.go` — `Agent.checkpointRound` (the mark ladder; `checkpointPrice = 10`, `checkpointRatio = 2`, `checkpointMarks = 3`), `checkpointMeter.round` / `markAt` / `checkpointMarkAt`, `Agent.readMark`, `checkpointSketchAsk`, `Agent.checkpointCeiling`, `Agent.handOverRunningTurn`.
- `internal/session/checkpoint_quick.go` — `Agent.quickFromDrawing` (its gates are `checkpointSketch.split`, `Agent.turnWroteNothing`, custody), `sketchItems`, `Agent.handOverAsQuick`.
- `internal/session/admission.go` — `admissionEvidenceHeading` / `admissionEvidenceRule` and `AdmissionHandle.line` (the `"; read " + h.Result` pointer); `internal/session/admission_compile.go` — `Agent.admissionContext`, `admissionEvidence`, `admissionEvidenceFloor`.
- `internal/session/prompts/quick.md`, `internal/session/sessionfile.go` (`journalCeiling`).

**The unit is a tool ROUND and not a write.** `checkpointMeter.round`'s own comment says why a round is the unit — it is the one the loop always knows, and it prices the person's waiting. But the write seam (`writeSeamFires`, asked first and at every boundary) is the only governor that reads what the turn DID, and a read-only turn crosses nothing but the round count. A turn that has read ten files and is composing an answer is therefore indistinguishable, at the mark, from a turn ninety rounds into a grind.

**And the value of a read-only turn is precisely the thing that is not carried.** The quick road already establishes that the turn wrote nothing (`turnWroteNothing`, the write seam's own counter read for zero) — which is the same fact as "the tool results ARE this turn's product". The handover then carries the drawing and pointers, and the worker rebuilds the context from the stub files at the uncached rate.

## The fix

The road is right for a turn that has fifty files still to touch. It is wrong for a turn whose reading is already in the room. Two changes, both narrow to the write-free case:

1. **A write-free turn hands over its READING, not a set of addresses.** The digest the mark's reader was just given already holds it — `checkpointDigest`'s `WHAT CAME BACK, NEWEST FIRST`, bounded and newest-first — so that compiled reading rides into the quick node's opening document in place of pointer lines, inside the room `admissionEvidenceFloor` already reserves. The worker's first request is then work. The existing "read the pointer" precedence stays for everything the bound could not fit.
2. **And a write-free turn whose drawn parts are an inventory of what it has already read is not moved at all.** Where every part of the drawing is answered by results already in the digest, the mark journals its carry-on and the turn finishes — which is what the person was waiting for. A round count alone does not end a turn that has read N files and has written nothing.

The person-facing line does not change: `this has parts · a quick task is taking them here, in this folder: <name>`.

## Acceptance

- **e2e:** through `aforge do --json` with the stub completer serving ten `read` rounds and a four-part drawing: the run's journal holds one `ceiling` row, and where it says `decision: moved`, `carry: quick`, the quick node's FIRST `call` row is not a `read` whose argument is one of the chat's own full-result pointers. Exit 0.
- **e2e:** the control, same door: a turn that WROTE (one `edit`) across the same ten rounds still takes the full road — the ceiling row's `carry` is a ladder rung and not `quick`, and the worktree handover is unchanged.
- **Unit:** every quick handover taken off a write-free turn carries a non-empty reading — assert on `Agent.admissionContext`'s compiled evidence for that node that at least one entry carries result text, and that the whole section stays inside `admissionEvidenceFloor()`.
- The manual page that covers the ceiling and quick tasks (`internal/manual/chat/how-tasks-run.md`) quotes the new behaviour, and the change entry's `invalidates` names what people believed before: that a quick node always opens on pointers.

### LINKED ISSUE #913: A node's requests walk six machines and re-bill the whole context: a held prompt cache is worth only its dollars in the comparison that abandons it
## What happened

`dev@71edb28b5`, 2026-09-11 13:59–14:08 EDT, one quick-task node, model `deepseek-v4.1-flash`. Ten requests for one node walked **six serving machines**, and the prompt cache was lost at every switch. The `calls.jsonl` rows for that node, in order:

```
Wafer        60,178 ms   ttft 37.6 s   no heartbeat
GMICloud      4,500 ms   cached  7,552
GMICloud      4,000 ms   cached 29,440
GMICloud      3,900 ms   cached 33,536
Novita            —      cached None
Novita            —      cached  7,552
Novita            —      cached  7,552
Venice       51,005 ms   ttft 40.6 s   no heartbeat
Morph       139,245 ms   ceiling
hedge → SiliconFlow/GMICloud — GMICloud answered 6.8 s, cached 39,424
Morph        95,979 ms
```

Of the node's 393 seconds, **346 were four requests on Wafer, Venice and Morph**. GMICloud's `cached_tokens` grew exactly as a warm prefix should — 7,552 → 29,440 → 33,536 — and then the next three requests went to Novita, which held only the fixed prefix (`None`, then 7,552, 7,552). Over the node the cached share was **24%** (132,096 of 550,705 prompt tokens); on one machine held throughout it would have been ~85%.

*Owner's forensics, may be gone:* the laptop's `calls.jsonl` for conversation `7598fe73c5ce3a9a`. The replication below stands in for it.

## Two halves, and only the second one is this issue

**(a) machines that had just failed inside the wait kept being admitted** — Wafer, Venice and Morph were each asked again after a no-heartbeat or a ceiling. **This binary predates #873.** `71edb28b5` is 2026-09-11 10:25 EDT; #873 merged as `834bdce26` at 11:04 EDT and `git merge-base --is-ancestor 834bdce26 71edb28b5` is false. #873 ("a machine is asked only when the answer is expected inside the wait its work will sit through") is exactly this half, and #891 already carries the seams it did not take. **So (a) is not filed here** — it wants a re-measure on a binary at or past `834bdce26`, not a second report.

**(b) a walk AWAY FROM A HEALTHY CACHE HOLDER costs nothing in the decision that makes it.** #873 changed `choose.go` and `frontier.go` and touched not one cache line (`git show 834bdce26 -- internal/lane/choose.go internal/lane/frontier.go | grep cach` is empty), so this half is unchanged on `dev` today.

And the claim "nothing prefers the cache holder" would be wrong, so it is worth stating what IS there before saying what is missing:

- `internal/provider/affinity.go` holds an **affinity pin** per prompt lineage + model (`endpointPins`, `Client.noteEndpointAffinity`, `Client.heldEndpoint`) — the endpoint that last answered leads the next request's order. But `Client.releaseEndpoint` drops that pin on **any** request that errored, 4xx'd, 5xx'd or had its stream cut, and its own comment says which endpoint caused it "is not knowable from here". A wait the ROUTER blew through therefore releases a pin the machine that holds the cache never broke.
- `internal/lane/choose.go` holds a **prefix memory** (`RememberPrefix`, `cachedTokens`, `prefixNote`, `PrefixHold = 5 * time.Minute` in `value.go`) and `PriceWithCache` discounts the believed-cached tokens.

The gap is what that belief is WORTH in the comparison. `scoreOf` (`internal/lane/value.go`) minimises `$·λ + perceived seconds`. A held cache appears only inside the dollars — on this node, ~30k re-billed prompt tokens is a fraction of a cent — so at any ordinary λ a machine that looks a second or two quicker beats the machine holding 33,536 warm tokens, and the walk pays the whole context again. Nothing in the frontier treats "this machine answered the previous request of this lineage with a growing cached count" as evidence in its own right, and nothing makes abandoning it need the evidence a strike needs.

## Replication

**Deterministic (no model).** `internal/lane`, with `lanestub` and the fixture beliefs the chooser tests already use (`choose_test.go`, `frontier_test.go`), plus `RememberPrefix` / `ForgetPrefixes`:

1. Two lanes on one model, both healthy. Lane **H** has served this request's `Prefix` and reported a prompt length — so `cachedTokens` credits it — and answers inside the role's wait. Lane **F** has never served this prefix and is believed slightly faster.
2. Ask the chooser for the next request of the same conversation, with the prompt grown a little.

What a developer sees TODAY: **F** is chosen while **H** is healthy and warm — the discount enters as `$·λ` and loses to a small difference in perceived seconds. Assert the same table with **H** failing to answer inside the wait: it must be abandoned there, which is the behaviour that must not regress.

The provider half is already replicable in `internal/provider` — `affinity_test.go` builds usage frames carrying `"prompt_tokens_details":{"cached_tokens":N}` — so a request that fails on a machine that never held the pin releases the pin anyway (`Client.releaseEndpoint`).

**Field (real models).** `bin/aforge`, one task whose worker makes ≥8 requests on a large context, with `OPENROUTER_API_KEY`; read the node's rows in `calls.jsonl`. It fires whenever a machine on the serving set stalls mid-node. A few minutes, a few cents. The control is the same node forced to one machine.

## Where

- `internal/lane/choose.go` — `prefixMemory`, `RememberPrefix`, `cachedTokens`, `prefixNote`; `internal/lane/frontier.go` — `scoredAt`, `frontierFor`; `internal/lane/value.go` — `PriceWithCache`, `PrefixHold`, `scoreOf`; the role's wait in `internal/lane/roles.go` and `contract.go` (`Request.Prefix`).
- `internal/provider/affinity.go` — `endpointPins`, `Client.noteEndpointAffinity`, `Client.heldEndpoint`, `Client.releaseEndpoint`, `Client.overPriceCeiling`.
- `internal/provider/hedge.go`, `internal/provider/armwatch.go` (`pathFaultReason = "no heartbeat"`), `internal/provider/calllog.go` (the `CachedTokens` column, `internal/provider/billing.go`'s `Billed.CachedTokens`).
- `docs/design/routing/` (#857's assessment) for where this axis is named.

## The fix

**Cache affinity becomes a scored axis rather than a tiebreak that money alone carries.** A machine that answered this lineage's previous request with a growing `cached_tokens` is worth, in the comparison, the tokens the walk would re-bill — measured against its own believed rate, not against a constant. Where that difference is smaller than the difference in expected seconds, the walk is right and happens; where it is not, the incumbent keeps the request.

**And walking away from a cache holder needs the evidence a strike needs.** `Client.releaseEndpoint` should release on a failure attributable to the pinned endpoint — one that answered late, cut its stream, or did not answer inside the wait — and not on every error the router returns for a machine that was never asked. The price ceiling release stays as it is: a warm cache was never worth any price.

Neither half is a threshold anybody picks: both are a difference in billed tokens against a difference in believed seconds, out of numbers the ledger already keeps.

## Acceptance

- **e2e:** one node driven through the real door with the stub pool (`internal/provider/pool`, `make test-remote`-style stub serving), where the warm machine stalls once and then answers: the node's `calls.jsonl` rows show `cached_tokens` **non-decreasing across consecutive requests served by the same healthy machine**, and the number of distinct `served` values for the node is bounded by the number of machines that actually failed to answer inside the wait.
- **e2e:** the control on the same door — a machine that goes quiet is abandoned: the next row's `served` is a different machine, and the run does not return to the quiet one until it has answered again.
- **Unit:** the chooser's comparison — with two lanes identical but for one holding a believed prefix, the holder is chosen iff the re-billed tokens outweigh the expected-seconds difference; and a role whose wait the holder misses never returns the holder.
- The manual page that covers routing and what a call costs quotes the new behaviour, and the change entry's `invalidates` names what people believed before: that the affinity pin was the whole of the cache preference.

santoshkumarradha:
## What this fixes, measured

A planning chat (issue #912) burned **2.4M input tokens vs 41k output (59:1), $1.41, 14.5 min**: 18 sliced `read` calls on one 5180-line file, endpoints rotating to 5 different machines, first `[taking stock]` note at round 3, compaction that freed 6%, and 44% of spend invisible in the transcript. Every fix below answers one measured number from that session (+ the follow-up router audit).

## The fixes

- **First-rung floor** (`checkpoint.go`) — the route race used to drag the first `[taking stock]` mark down to the boundary it landed on (fired at round 3 over 43 KB); the mark now waits at `checkpointPrice/checkpointRatio` (round 5), with a probe test pinning the floor against race-at-any-round.
- **Read dedupe** (`heldreads.go`) — a `read` whose range is fully covered by what this conversation already holds answers with a one-line pointer instead of re-paying bytes; spans carry a fingerprint of the message text, so stub/fold rewrites invalidate coverage by themselves; changed files (mtime+size) read fresh.
- **Fold old slices** (`turnfold.go`) — repeated reads of one file keep the newest slice whole and fold the older ones to `read offset/limit` pointers, inside the existing per-step fold.
- **Billed arms leave call lines** (`sessionfile.go`, `loop.go`) — hedged rescue arms and cut-stream retries, previously banked but invisible, leave a `call` row with an `arm` marker; a mid-session model hop now seals one usage line per answering model (was: all calls rolled onto the last model).
- **Cache pin in the enforced set** (`lanes.go`) — the affinity pin used to die whenever the speed frontier dropped it (`provider.order` advisory, honored 29%); it now joins the demand unless patience-refused or the person pins a lane themselves.
- **Shaped takeover gate** (`routefirst.go`) — the OpenRouter-takeover gate counted any refusal as evidence, including refusals of *our own narrowed demand*, so a takeover fed on itself; only a refusal of a bare (router-default) request counts now.
- **Stale-build launch sweep** (`taskpresence.go`, tui3) — the third binary-shadow incident in this repo's history: one fail-open presence sweep at launch names live sessions on an older rev, with the pids — warn, never blocks, never kills.

## Verified

- `go build ./...` green; focused tests for every fix green with `-count=1`; `make test-laws` green (25 packages).
- Live sessions on the branch: single endpooint held with `cacheRead ≈ input` (97–100%), dedupe pointer firing, marks at the ladder geometry, takeover storm gone.
- TUI e2e: 12 subtests pass; the 5 reds **also fail on clean `origin/dev`** — filed as #1014 (not our regression).
- Effort/wire auto verified unchanged: no `reasoning` field is stamped for default chat turns; OpenRouter auto governs.

## Notes

- Change entries 1004–1010 in the wave's vocabulary, one per fix.
- #1014 tracks the five rotting e2e needles on dev.

Fixes #912
Fixes #913


santoshkumarradha:
Heads-up from the stacked #1017: `docs/changes/unreleased/1015-stale-build-sweep.md` has `surface: session, tui3` as a plain scalar, which fails `make changelog-check` (unmarshal into []string). Fixed as a drive-by on the stacked branch; flagging so you can fix here before merge too.

santoshkumarradha:
Two more pre-existing gates trip on this branch (found while running pr-ready against it for #1017): (1) `TestEveryLawIsStatedOnceAndInItsOwnPlace` on `read.offset-limit` — the description cut left its registry sentence unreferenced; (2) gofmt drift in `internal/tui3/app.go` (double blank line). Drive-by fixes for the second are on my branch; the first needs a sentence put back or the registry line deleted.

PR #1036, opened 2026-09-14
session, tui3: a task-start question you answered is not asked again when you come back to the tab

santoshkumarradha:
## What a person sees now

Approve a task, look at another tab, come back: the assignment is there with
**your answer under it** — `start it · approved` — and nothing to answer. The
conversation behind it takes the keyboard again, so it scrolls.

Before this, coming back re-asked the same question, every time, for as long as
the turn that proposed the work was still running.

## Cause

A turn's backlog is every event that turn has sent, kept verbatim so a surface
arriving mid-turn reads the work from its first line (`eventHub.attach`). A
**card** in it is the one event that is not a report of something that happened
but a question about something that has not — and the replay handed it over
whatever had since been decided.

So `EventTaskProposal` went out again on every re-attach. `app.proposeTask` found
no card for that id (the atomic replay's entries stop at the turn's floor, so the
running turn's card exists only in the backlog), built one, and raised the
question — a question the engine had already recorded an answer for and would
therefore never close. And because a question takes the keyboard where it is
drawn (#919), the conversation underneath it could not be scrolled while it
stood. **Both symptoms, one cause.**

## Fix

- `questionAsked` names the question a card put in front of somebody, in the
  lanes' own keys — consent, a task proposal, a standing card, a harness offer,
  a connect ask. `eventHub.attach` is handed what is still being asked
  (`Agent.stillAskedLocked`, the banked words read under `a.mu` at the moment of
  the attach) and leaves out a card whose question is no longer one. The backlog
  stays a faithful log of the turn; what is filtered is the REPLAY.
- A proposal **restates itself when it is settled**: `taskWait.decided` sends the
  card again with `TaskNotice.Decided` on it, the moment anybody answers — a
  person, another window, or the countdown — beside the restatement
  `taskWait.withdraw` already made for the other ending, at the one place both
  roads pass through. A card that states its own outcome is not asking anything,
  so the replay keeps it in the open card's place.
- The surface draws it as what it is. `taskVerdict` is the one reading of
  approved-and-redirect that the window whose key answered and the window
  replaying a decision both take.

The first commit alone would have left the forming block saying
`not started · the call was refused` over work that was running
(`app.refuseFormingCard` reads a `propose_task` result with no card as a
refusal), which is the second commit's whole reason.

## Before / after, on the real binary

Driven as a person in tmux on the Spark against `deepseek/deepseek-v4.1-flash`:
two tabs, a task-start question answered by hand while the turn went on running,
then `tab` away and `tab` back three times.

**Before** (`~/af-taskq-evidence/before4/3-midturn-back-1.txt`, and identical on
rounds 2 and 3):

```
  ╭─ ? ◆ Go program printing first 20 primes, with test ───────────
  │ Write a small Go program in /tmp/atq/before4/w that prints the first twenty primes, plus …
  ╰──────────────────────────────
─ conversation about saying only the word one · deepseek/deepseek-v4.1-flash · ⠿ auto ·
╭─ ? wants to start a task: Go program printing first 20 primes, with test ─────
│ ▸ 1  start it
│   2  no
╰─ ↑↓ choose · enter take it · esc later ──────────────
  c change · ? ask back · 1–2 jump
 › say what to change… (enter sends it)
```

The second symptom is visible in that same before-run: the `tab` I sent to
switch tabs went to the QUESTION instead — round 2 shows the pointer moved from
`1 start it` to `2 no` and the tab never changed. A question takes the keys where
it is drawn, so nothing aimed at the conversation reached it.

**After** (`~/af-taskq-evidence/after5/3-midturn-back-1.txt`, identical on rounds
2 and 3):

```
  ╭─ ? ◆ Go program printing the first twenty primes, with test ────────────────
  ╰─ start it · approved
    ⇉ Go program printing the first twenty primes, with test
    · this one wants more hands · handing it over with everything found so far
─ conversation naming with one-word response constraint · deepseek/deepseek-v4.1-flash · ⠿ auto
 ›
```

No question. The card carries the answer. The box is the conversation's again,
so the keys are too.

## Spark evidence

Worktree `spark:~/af-taskq`, revision `7ccfdd12f`. `go build ./...`, `go vet
./...`, `make test-laws`, `go test ./internal/session/ ./internal/tui3/
-count=1` — all green (`spark:~/af-taskq-tests2.log`, session 192s, tui3 312s);
`./internal/remote/` and `./internal/enginehost/` green at `20fc1fee7`
(`spark:~/af-taskq-tests.log`). Captures under `spark:~/af-taskq-evidence/`.

## Tests

- `internal/session`: `TestAnAnsweredProposalIsNotAskedAgainOfASurfaceThatArrivesLater`,
  `TestAProposalNobodyAnsweredIsStillAskedOfASurfaceThatArrivesLater`.
- `internal/tui3`: `TestAProposalDecidedBeforeThisWindowArrivedIsDrawnAnsweredAndNotAsked`,
  `TestADeclinedProposalDrawnOnALaterAttachSaysSo`, `TestAProposalNobodyHasDecidedIsStillAsked`.

All deterministic — a held clock, no sleeps, no model.

## Manual

`internal/manual/chat/questions.md` gains
*"You answered it, then switched tab and came back — is it asked again?"*, in the
words somebody would search for, and it says what is still there when you return
and what is not.

https://claude.ai/code/session_0182eyxXP8Xe7ADNKpYDe4wv


santoshkumarradha:
This fix covers the replay half: a card whose question was answered is not shown again when a window re-attaches. The same re-raise also happens on the **landing** path, which rides `EventTaskUpdate` and is outside the five card kinds this filters — filed as #1077, with a session where one landing card was accepted twelve times in an hour and the engine recorded every answer.

<sub>drafted with [agentfield ai](https://agentfield.ai/github?utm_source=github&utm_medium=comment&utm_campaign=drafted_with)</sub>

PR #1040, opened 2026-09-14
tui3: the switcher card and its tab row are one reading, on alt+k

ZeroPoint95:
**One idea: the switcher card and the tab row are the same list, read two ways.** Six commits follow from that; the seventh is spacing.

**The card is now `alt+k` (`opt+k` on a Mac).** #1062 is folded in and closed as subsumed — it moved the switcher off `ctrl+k` so that key could be kill-to-the-end-of-the-line, and it deletes the `Chats ▾` control on the tab row. Both PRs edited the same three files and #1062 deleted `hopOpenAll`, which this one edits, so one resolution beat two.

## What you'll notice using it

1. **`enter` works from home.** It used to switch the conversation *behind* home and leave home standing, so the key looked dead while it had quietly swapped what the screen was drawn over. `hopTake` now ends in `hopLand`, which comes down off whatever place is showing — home, tasks, standing, memory, spend, search, settings.
2. **Rows are in tab order.** Leftmost tab is row 1, so `3` on the card is the third tab. The cursor still starts on the last conversation, so `alt+k` `enter` still means "back where I was".
3. **Close a tab, the row leaves the list** — down behind the fold, still held and still running, `enter` brings it and its tab back. By the `×`, by `ctrl+w` in the conversation, or by `ctrl+w` on the card.
4. **The foot is `→ show closed` / `← hide closed`**, and one dim `closed` marks where the tabs stop, spaced the way the head's `open` is.
5. **Long names fit.** The name column takes the spare width instead of `nothing new` taking it, with two cells of air so a name that fills its column never runs into the clause.

## Wording fixed

- `ctrl+w put away` → **`ctrl+w close tab`**. `put away` is home's `ctrl+e`, which ARCHIVES a conversation until its name is typed — a durable act one keystroke away from this one. The card's own receipt already said `tab closed`.
- **Every `✕` row says why it refuses**: `open in another window`, or `that folder is gone`. `hopRest` asked the disk about the folder for the glyph but left the field the clause reads false, so one of the two cases drew the refusing mark with nothing beside it.

## Removed

- The wrapped title block under the list. It only ever repeated the selected row's own name, only appeared when that name happened to be long, and cost the rows under it their places on a short card.
- The `Chats ▾` control at the right end of the tab row, with #1062's own reasoning: `→` opens the fold and a second door is a second thing to keep in step. `hopOpenAll` goes with it.

## Found in review, fixed

- **`ctrl+w` on a conversation this window never opened closed nothing and said it had.** `hopTabbed` asked `tabShut` alone, and nothing ever dismissed a conversation this terminal has not held — so the map was silent and silence read as "it has a tab". It now asks `row.open` too. `hopReading` never noticed because every row it splits is one the keeper holds; `hopAway` did.
- **The seam's three lines were charged whether or not it was drawn.** On a short card the window lost three rows, the rows it lost were the closed ones, and `→ show closed` made the list SHORTER and showed nothing. They are given up in order now — the air, then the word — and a card too short even for the word draws the closed rows without it.

## Under it

- `hopReading` splits its rows on `hopTabbed`, which asks `tabShut` and never the drawn strip — the strip is a frame behind the keystroke, so a card rebuilt from it after its own `ctrl+w` kept the row it had just closed.
- `hopStripOrder` lays the rows out the way the strip is; the recency walk survives as the tie-break for a conversation with no tab.
- `hopFirstStop` is a method now: it walks the previous-stack, because row zero stopped being the conversation `tab` would go to.
- `hopOpenRows` is the figure the reading returned rather than a walk of rows, which gain the rest of the machine when the fold opens. The head's `3 of 12` is how many TABS are on the row.
- `hopSubjectCol` is gone: the clause has the fixed column (`hopNoteCol`) and the name takes what is left, up to `hopSubjectMax`.

## Paperwork

`make pr-ready` green on the head commit — the light gate plus the full `internal/tui3` suite with `-count=1`, 317s. Fourteen tests added or rewritten in `hop_test.go`, each verified to fail without its change. Six change entries carry the invalidations.

`keys.md`'s switcher section broke the manual's retrieval gate by growing too long after three changes landed in it; the layout and the ordering are now sections of their own. The card drawing on that page is regenerated from `hopCardLines` rather than hand-patched, which is also how the one conflict with #1041's `codeaf` rename was resolved — nothing else in the branch spelled the product name.

## One defect found on the way, already fixed on dev

`internal/tui3`'s suite walked the developer's real `$HOME` — `askFolderStore` → `scanFolderRoots` starts at `os.UserHomeDir()`, and `TestMain` deliberately leaves `HOME` alone — so on a cold cache or a loaded box that scan blew the harness's five-second budget and panicked, taking the whole test binary down and reporting the package red at ~60s. It cost this branch several false reds. #1058 landed the seam (`folderRootScan`, answered with nothing in `TestMain`) while this was open, and it is gone.


AbirAbbas:
Verified this against `dev` at the current head (green CI, adversarial read, and driven on the real binary). Two things block it, and there's a conflict to resolve first.

**Blocker — `ctrl+w` on a never-opened conversation prints a false `tab closed` receipt.** `hopAway`'s guard is `!a.hopTabbed(row)` (`internal/tui3/hop.go:1764`), and `hopTabbed` (`hop.go:526`) returns `!a.tabShut[key]`, which is **true** for any conversation this window has never held — nothing ever set `tabShut` for it. So `ctrl+w` on a below-fold machine row falls through, calls `tabShutKey()` on a key with no tab, and sets `a.hop.say = "tab closed · <title>"`. Nothing actually closes. This is a regression from `dev`, which refuses correctly, and it contradicts the manual page this same PR writes.
- Repro: `ctrl+k`, `→`, put the cursor on a below-fold conversation this process never opened, press `ctrl+w` → the card says `tab closed · <title>` and the tab row is unchanged.
- Fix looks like one line in `hopTabbed`: `if row.here { return true }; return row.open && !a.tabShut[a.convKey(row.file)]`. Adding `row.open &&` changes nothing in `hopReading` (every row it splits is already `open`), it only fixes `hopAway`. Worth a contract test for the never-opened case — the existing `ctrl+w` test only covers the tab-you-closed path.

**Should-fix — `→ show closed` shows nothing on terminals ≤23 rows.** The seam's line reservation (`1 + 2*seamAir`) is subtracted from `available` unconditionally, but the seam itself is only drawn inside the scroll window. When the window is short the reservation is spent and no closed row is reached, so opening the fold makes the list shorter and shows nothing — while the foot still flips to `← hide closed`. Closed rows first appear around 24 rows.

**Conflict with #1062.** This PR keeps the switcher on `ctrl+k`; #1062 moves it to `alt+k` and makes `ctrl+k` a kill command. They conflict in `internal/tui3/hop.go`, `internal/manual/chat/keys.md` and `screen.md` (this PR edits `hopOpenAll`, #1062 deletes it). They're mutually exclusive — worth deciding which direction wins before either lands.

AbirAbbas:
Reopened — I closed this by mistake a minute ago, working from a stale view in which #1062 was still the live branch. This is the right vehicle: #1062 is folded in and closed, the `ctrl+w` refusal and the seam fix are here, and the `how long does stop take` probe is fixed here too. Please ignore the earlier note about carrying anything onto #1062. Waiting on CI for the new head.

PR #1043, opened 2026-09-15
spend: one cut of the ledger at a time, four columns, and doors that open the thing

ZeroPoint95:
**The spend place (`alt+3`) was a chart, two ragged tables and a door that went to the wrong screen. It is one cut of the ledger at a time, in four columns, with a control that swaps the cut — and the doors open the thing.**

## Before

before

## After

> The screenshot below is from an earlier round of this branch — the layout is current, but the headings now read `by model` / `by topic` and only one is drawn at a time.

after

```
 14 days came to $2.10 · 28.9M tokens · loudest day: $2.03 sep 14 (bandit-algorithm)

 ← by topic →                                       ← the cursor is on the heading
 · bandit-algorithm                    task          7  $0.63
 · deepswe model discussion            chat  aforge-v2  $0.38

 by standing order
 · repo-watch                    88 firings  $0.01 a run  $0.88
```

---

## The bug you can't see in a screenshot

`enter` on a row of *what it was for* **did not open that thing**.

- a task opened the tasks *place* — then left you to find your own row in a list of everything this machine has ever run
- a conversation opened **home**, the switcher, with the conversation you named nowhere on it

Both go through the doors the rest of the surface already uses now. A task opens **its own record card**. A conversation is **brought forward** if this terminal has it, otherwise opened beside the one you're in. A promise opens the standing place **on that order**. A thing the record no longer holds **refuses where it stands** instead of walking you to a page about everything.

A test was asserting the defect. It waited for `pageHome`.

---

## One cut at a time

`by model` and `by topic` are the same money added up two ways — every dollar under one heading is also a dollar under the other — so a page drawing both asked you to read one bill twice.

The page draws one. It opens on `by topic`, the question you walk in with. Walk the cursor onto the heading and it wears arrows: **`←`, `→` and `enter` step between the two.** The arrows appear only there, because `→` on every other row opens that row's verbs — a key is drawn where it is bound. An empty window keeps the control, or `by model` would have no way back.

Standing orders keep a heading of their own under `by topic`: a promise is one of the things money was *for*, so it's a third heading rather than a third cut.

## Why the bars went

A bar was a model's share of the dearest one — a reading the figure at the end of the same row already gives. It cost a reserved column, plus a second reservation in front of it so a role word on one row couldn't shove it out of line. The list is sorted dearest first. That is the chart.

## Why the figures moved right

What a row **is** is read from the left. What it **cost** is read by comparison with the row above — and a comparison is made on a figure's right-hand edge. So the facts travel together in a block and the names run out to meet them.

That block ends **where the chart ends**, on the `today` point. Money used to be flushed to the frame, which on a wide terminal put the one figure every row is read for forty cells from the counts it belongs with.

**A row gives up its words before its figure.** A long name pushes the fields behind it, but only as far as the cells the money needs.

## Numbers

- **Money is a column of cents with a floor.** Under a cent reads `$0.01` — not `$0.0068`, not the words `under a cent`. The head line, `/cost` and the Spending tab keep exact arithmetic; a floor is a column's rule, and a total in a sentence has nothing to line up with.
- **Thousands are marked everywhere.** `$4,210.55`, `$50,000`. The page used to draw `128,400 calls` and `$4210.55` on one row.
- **Token volumes got a `B` rung.** A fortnight of agent work read `7062.1M`. Now `842` · `12.4k` · `1.2M` · `3.2B`.

## Words

- `a task` → **`task`**, `a conversation` → **`chat`**. Column words, not sentences.
- The headings are `by model`, `by topic`, `by standing order`. They were sentences that named the page's subject again before getting to the point — and the tab already said `spend`.
- `under a cent a run` is gone; a promise's per-firing figure goes through the money column's rule, so it reads `$0.01 a run`.

## Smaller things

- **Twenty rows under `by topic`, not three.** Three is a headline, not an answer to the question this page is for.
- **The loudest day lost its row** and became the head line's last clause.
- **The head reserves the zoom clause** before its own optional facts. A key crowded off this line is not hidden, it is **unbound**.
- **A missing project draws nothing**, not `.` — `filepath.Base("")` answers `"."`, and that was reaching the screen.
- **Narrow frames give up whole fields**, in a named order of worth. Never a figure with its tail cut off. Never the money.

---

## What the review caught

Three blockers, all real, all fixed in `6d3d656`:

1. **A task row could open another conversation's task.** The join matched the index's `SessionID` — the conversation that ran the work — against `SubjectSpend.Session`, which is the task node's *own journal id*. The exact-pair arm never fired on real ledger data, so every row fell through to an id-only match, and ids restart with every conversation. `SubjectSpend` carries `Root` now and the join is `(id, Root)`. The fixture and the test were green because they spelled the ledger a way no ledger is spelled; both write the real shape now.
2. **The money could be truncated.** Fields were laid left to right and the row trimmed to the frame at the end, so a long name or project cost the row its tail. And `spendProjectCap` reached the measure and not the draw — a cap that moves the table and bounds nothing.
3. **The cut's control vanished on an empty window**, because it lived only on a heading over rows.

Plus the three smaller notes: the head line's floored formatter, the silent fallback on a stale target, and `places.md`.

---

## Checking it

```
make demo-home
```

Then `alt+3`. The fixture carries a heavy stretch — five-figure call counts, billion-token volumes, money in the thousands — so the columns are at full width.

## Notes

- Manual updated in the same change (`internal/manual/chat/models-and-cost.md`, `places.md`), as the manual law requires.
- Deleted: `spendBar`, `spendModelBarCap`, `spendReading.modelCols`, `roleCol`, `spendReading.subjectCols`, `loudestRow`, `loudestRowIn`, `spendDoorWord`, `spendOpensWord`.


AbirAbbas:
Verified against `dev` at the current head (green CI, adversarial read, driven on the real binary). The three-table layout and the new `by topic` / `by model` cut control both work on-screen, but three things block it.

**Blocker — a task row can open another conversation's task.** `spendTaskRecord` (`internal/tui3/place_spend.go:579`) matches `TaskIndexEntry.SessionID` against `SubjectSpend.Session`, but those are different things: the ledger writes the task node's own journal id into `UsageLine.Session` (`internal/session/usage_ledger.go`), while the index's `SessionID` is the conversation that ran it. So the exact-pair arm never fires on real ledger data, every task row falls back to the id-only match, and since task ids restart per conversation, whichever conversation the world walks first wins. The demo fixture and the new test both put the root id in `Session`, which is why they're green. `SubjectSpend` has no `Root` field, so this needs a small plumbing change in `internal/session`.
- Repro: two conversations each holding a task id `7`; `enter` on that spend row can open the other conversation's task `7`.

**Blocker — the money's tail is truncated at ≤120 columns with a long name/project.** `spendProjectCap` (26) is applied on **measure** (`spendMeasured`, `spendplace.go`) but the row is painted uncapped, so a project or subject name of ~30+ cells pushes the row past `width` and the closing `fit(row, width)` cuts the right-hand end — which is the money (`$21.…`). Thresholds: ~38 cells at 80, ~30 at 100, ~36 at 120. This is a regression vs `dev`, where the right side was fitted first, and it contradicts the body's "Never a figure with its tail cut off. Never the money."
- Repro: a project folder named e.g. `agentfield-control-plane-web-ui` at a 100-column width → the row ends `$21.…`.

**Blocker — the cut control disappears on an empty window.** The `←→` control lives only on the heading row, and the heading is only drawn when the selected cut has rows. Step the window (`shift+←`) into a fortnight with no spend and there's no heading, no arrows, and no `←→ by topic` foot — no way back to the other cut except stepping the window forward again.
- Repro: switch to `by model`, then `shift+←` into an empty fortnight.

Smaller: the head line uses the floored money formatter (`$0.01` for a `$0.0068` window) while `/cost` and Settings→Spending keep exact arithmetic; a stale task/conversation still falls back silently to the old tasks/home page with no refusal; `places.md:507` still describes the old three-block page.

PR #1046, opened 2026-09-15
tui3: a home panel stands where its content puts it

ZeroPoint95:
Home's law 2 was **act left, watch right** — the person's three panels on the left, the machine's four on the right, whatever any of them held. On a quiet machine that drew six whispering headings over a screen that was five sixths scaffolding, with the one panel that had rows boxed into a third of the width.

The column is now a reading of what the panel holds.

- A panel **with rows** stands in the **field** — the columns left of the last one, filled from the top left corner down.
- A panel **with nothing in it** stands in the **rail** — the last column, flush with the right edge, as its heading and its whisper.
- `projects` and `spend` are **pinned** to the top of the rail whatever they hold, because their height is the same on every machine on every day. One blank row separates that pair from what is only quiet today.

https://github.com/user-attachments/assets/079a6e61-fa10-4370-a295-9d1998276eb4

### What it costs, taken knowingly

Home's geography moves now, and nobody can learn that `running` is on the right. The **rank** inside each column never moves, so the cost is to one axis and not both. The owner ruled this on 2026-09-15 after reading a real quiet home.

### Two consequences that needed their own call

- **The field fills rather than balances.** Balancing put the cursor's panel in whichever column had room, so `↓` off the tab bar landed on a different panel at 120 cells than at 200 — and the landing is the same line at every width, because a landing that moves with the frame is one nobody can build a habit on. Filling from the left keeps it stable, and never spreads two short columns over a wide screen. The middle column at three is the descriptions': a row's sentence stands beside it there, and a question raised from home stands there as a card; no panel of headings is laid in it.
- **`←→` now step over a column with nothing to stand on.** Without it the rail was out of the arrows' reach entirely on a wide quiet frame — `projects` and `spend` drawn on the screen with no key that walks to them.

### The one person-facing sentence that moved

Home's resting foot gains `ctrl+o open folder`. That clause was already there on any machine whose right column had rows (§6.6 refuses to leave a door invisible when `→` crosses columns instead of opening a row's verbs); what changed is that the rail always has rows, so the condition is always met. The design's four keys still open the sentence and `esc` is still not on it.

### Proof

`make pr-ready` green — the light gate plus the full `internal/tui3` suite (318s). All 399 home tests pass. Six tests were rewritten to pin the new law rather than the old one, and two were added: the rail's group gap, and the descriptions' middle column at three.

`docs/design/home-mission-control/DESIGN.md` carries the amended laws 2, 3 and 6 and says what each of them used to say. The manual's `home.md` is redrawn under the new geography, its opening sketch included.

### After review

The review's three blockers and four should-fixes are in `006fbd6a3`: a `scheduled` row draws its one sentence per kind at every width, as the line under the cursor where there is no middle column; an open fold reads `3 fewer · 40 more` and names no place, because `enter` on it toggles the panel and the heading is the door to the place; `enter` on an order made from home lands standing on that order; the tagged e2e fold test waits for `N more`; and the manual's retired roster, resting foot, fixed-columns claim and the `elsewhere` / `made from home` refusals are rewritten to what the code does. `make pr-ready` green again.


AbirAbbas:
Verified against `dev` at the current head (green CI, adversarial read, driven on the real binary). The field/rail layout engine is largely right and the earlier back-fill bug is fixed, but the manual law and a couple of narrow-frame behaviours block it.

**Blocker — the compiled manual still teaches the retired layout.** The chat corpus is the model's only source of truth about codeaf, and it still describes the home this PR replaced:
- `internal/manual/chat/places.md:417` and `internal/manual/chat/commands.md` still teach the fixed "act left, watch right" roster.
- `internal/manual/chat/home.md:762`, `commands.md:1061` and `keys.md:2067` say the resting foot reads **exactly** the old four-clause sentence (`… enter open · tab next place`) — contradicting the new foot with `ctrl+o open folder` this PR writes elsewhere in `home.md`.
- `home.md:2243` still says the panels' columns are fixed; the whole PR makes a panel's column depend on whether it has rows.

The manual gates pass because they only check a name is *mentioned*, not that the claim around it is true — exactly the failure mode CLAUDE.md warns about.

**Blocker — scheduled rows lose everything but their title below ~170 columns.** The right margin was removed and `nextUpSaidOn` returns empty without the middle column, so at 120 and 80 columns the `scheduled` panel shows only bare titles — no kind, no time, no last result. (Confirmed on real frames at both widths.)

**Blocker — the manual describes refusals the PR removed.** `home.md` still says another-project items refuse with `elsewhere` and home-made items say `made from home — no conversation to open`; on the branch the first opens its transcript and the second opens standing.

Should-fix: the body/`DESIGN.md` still promise an "empty middle column" that now holds descriptions; an overflow fold can read `N fewer · N more · tasks` but `enter` only collapses the panel rather than opening the place it names; the tagged e2e fold test still waits for the retired string `type to find one` (would time out on a built binary); a home-made watch opens the generic standing page rather than the item via `openStandingAt(id)`.

PR #1060, opened 2026-09-15
A launch says a newer codeaf is out, and /update installs it and restarts

AbirAbbas:
The owner's ask: tell people a new version is out, give them the command, and —
better — let them update from inside the chat and come back on the new build.

## What a person sees at launch

A stable build that is behind the newest stable release gets one dim transcript
line, and nothing else changes:

```
codeaf v0.2.0 is out · you have v0.1.0 · /update installs it and restarts · or: curl -fsSL https://raw.githubusercontent.com/Agent-Field/codeaf/main/scripts/install.sh | bash
```

It carries both versions, the command, and the curl line, because the ask was
that the command be given. The look is a command, not an opening read, so the
first frame is drawn before the network is asked anything; three seconds is the
whole budget, and offline, refused, rate-limited or malformed all draw nothing.
The answer is cached for a day beside the profile, though a cached newer release
still shows the line at every launch — the reminder is the point.

## Who is prompted

Only someone who could sensibly be behind. A stable tag when a greater stable
exists; a release candidate once its stable line is published. A `dev-*` or
`staging-*` channel build, a `make build` sha and an unstamped binary make **no
request at all** — a channel build is deliberately newer than stable, and a
source build belongs to whoever built it. `codeaf update --check` still answers
on those, naming what the running build is.

Opt-out: `CODEAF_NO_UPDATE_CHECK=1` silences the launch check only; `/update`
and `codeaf update` still work. Headless doors never check.

## What /update does

Names the tag, then the checksum and where it landed, then restarts:

```
· downloading codeaf v0.2.0 for linux/amd64…
· checksum matched · installed at /path/to/codeaf
· restarting on v0.2.0…
```

Already newest says `you are on the newest codeaf, v0.2.0` and stays. It refuses
while a turn or a task is running, and once started, one shared guard stops a
second `/update` or a new turn from beginning underneath it — the restart must
never quit on top of live work. The restart happens in the door, after the frame
is down and the terminal is back, and re-execs with this launch's own arguments
plus `--session` on this transcript, so the same conversation reopens.

## What `codeaf update` does

`--check` prints one line and exits 0 (newest), 3 (a newer one exists) or 1
(could not ask). Without it, installs stable by default, or `--rc`, `--dev`,
`--staging`, `--version <tag>`. A source build refuses in place and names its
path instead of replacing something a developer is holding. An unwritable target
is left alone and offered the curl line; sudo is never attempted.

## Hand-verified, and how

The repository is private, so `releases/download` cannot be reached anonymously.
The real `v0.2.0` assets and release JSON were mirrored to a local HTTP server
and reached through `CODEAF_GITHUB_API` / `CODEAF_GITHUB_DOWNLOAD`, with a
binary stamped `v0.1.0`. Through tmux, against that mirror: the launch line
appeared; `/update` printed the three notes above; the process restarted; the
restarted build reported `v0.2.0` and resumed the same transcript; `codeaf
version` then printed one line naming `v0.2.0`. Full commands and frames are in
the lane's verification record.

Not verified here: Windows (cross-build only — the `ReplaceFileW` path has a
contract test but has never run on Windows), and a real anonymous download from
a public repository.

## Two things found on the way

The engine kept its twenty-second torn-link grace on a *deliberate* detach, so a
cleanly closed window looked attached long enough to keep a freshly installed
codeaf behind the engine the old one had already left. Fixed here.

`scripts/install.sh` and `internal/update` were reading the same release list by
different rules — the shell accepted any `dev-*`/`staging-*` tag and semver parts
with leading zeroes. They share one grammar now; #1056's publish-time ordering is
untouched, only which candidates are eligible.

## Review

Two adversarial review rounds against the contract. Round one found three
blockers — a `/update` race, `Authorization` surviving a redirect to another port
on the same host, and an ignored Windows rollback error. Round two found two
survived: the guard was missing from the home page's Enter road, and
`ERROR_UNABLE_TO_MOVE_REPLACEMENT_2` is a documented *partial* failure that can
leave no file at the target. Both fixed, each with a test that fails without the
fix.


internal/update/client.go
```diff
@@ -0,0 +1,259 @@
...
+	"strings"
+	"time"
+
+	internalenv "github.com/Agent-Field/codeaf/internal/env"
+)
+
+const (
+	// CurlCommand is the independent installation road shown after a failed or
+	// unavailable in-place update.
+	CurlCommand = "curl -fsSL https://raw.githubusercontent.com/Agent-Field/codeaf/main/scripts/install.sh | bash"
```
santoshkumarradha: should this ne agentfield.ai/get/codeaf ?

AbirAbbas: ill update it to that after repo is public

AbirAbbas: Done in 79230ec66: `CurlCommand` is now `curl -fsSL https://agentfield.ai/get/codeaf | bash`, and the manual’s update section names that line in place of "the curl line below". Checked against the built binary: a refused `codeaf update` ends with the new line. The pre-existing install section on this page is #1076’s to rewrite; the two will need a small merge in `running-from-the-terminal.md` whichever lands second.

PR #1062, opened 2026-09-16
tui3: ctrl+k kills to the end of the line, and the switcher moves to alt+k

ZeroPoint95:
`ctrl+k` is kill-to-end-of-line in every shell on the machine. Here it opened a card. This gives the letter back to the message box and moves the conversation switcher one modifier across.

## What works now

1. **`ctrl+k` kills to the end of the line** — message box, home's box, every filter box. The pair to the `ctrl+u` that has always killed to the start.
2. **`alt+k` opens the switcher** (`opt+k` on a Mac). Card, keys and behaviour otherwise unchanged.
3. **`alt+shift+k` walks it backwards — in every terminal.** `ctrl+shift+k` could not, and was bound on maybe half of them.
4. **Home's rule names it**: `alt+w folder · alt+o model · alt+k chats · / commands`.
5. **`Chats ▾` is deleted.** The `+3` overflow count stays, dim and inert.

## Why the switcher moved and not the edit

Every `ctrl+<letter>` on this surface is spent. Only `h`, `i` and `m` are free, and the terminal already spends those on backspace, tab and enter. So the edit had nowhere to go and the card did.

`alt+` is the modifier the places are already built on. `k` is the letter people have learned.

## What the move bought

`alt+shift+k` is escape-then-`K`, a different byte from `alt+k`'s escape-then-`k`. `ctrl+shift+k` and `ctrl+k` were the **same byte**, so the reverse gesture needed a terminal that had taken the kitty disambiguation flag. Now it needs nothing.

## What it costs

A Mac whose Option key composes accents: `opt+k` types `˚` and no card appears. Same setting every other `alt+` chord here already needs.

Handled rather than left to be found: `˚` is on the dead-key table, and it is the one character there read in a **conversation** as well as on a place — because it is the one chord bound in both. The hint slot says `opt types a letter · turn on option as meta`.

Linux and Windows pay nothing. Alt is already meta there.

## 3 bugs found on the way

1. `ctrl+k` reached the composer but **not home's box** — five boxes hand-roll their own key switch and all five were missing it. Home is where a launch lands, so that was the first box anyone would try it in.
2. Home's rule, the question chip and two key-sheet rows drew `alt+` on a Mac while the rest of the surface drew the Mac spelling. They were built straight out of their constants and never went through `chordSpelling.say`.
3. `config.KeyQuickSwitch`'s doc said the setting governed the switcher's own chord. It has always governed `ctrl+tab`.

## 2 new laws

- **`killpairlaw_test.go`** — fails any key switch answering `ctrl+u` without `ctrl+k`, naming file and line. Two exceptions named with reasons (string fields with no caret).
- **`chordspellinglaw_test.go`** — renders the conversation, key sheet, all seven places and the key map on a Mac's spelling, fails on any `alt+` reaching the frame. A frame test, not a source test: every constant is *supposed* to say `alt+`; what was wrong was the path from constant to paint.

Both verified to bite by reverting a fix and watching them fail.

## Separately: opt+ replaces ⌥

Every modifier on a Mac keycap carries a word **and** a symbol — Control is `control` and `⌃`, Option is `option` and `⌥`. "What the keycap says" never chose between them, and this surface took the word for one (`ctrl+`) and the symbol for another. `opt+` is `ctrl+`'s own move made twice.

Bindings untouched. `opt+` is four cells, exactly as `alt+` is, so every ladder that fits a chord on Linux fits it on a Mac. `⌥` measured as one cell and was not reliably one cell on screen — U+2325 is East-Asian Ambiguous. It was also the one person-visible mark reached as a string literal rather than through `tokens.GlyphSet.Glyph`.

`⌘` is still a glyph and is now the only one.

## Manual

`keys.md`, `screen.md`, `home.md`, `commands.md`, `places.md`, `tasks.md`, `getting-started.md`. New section answering *"why not the option symbol"*, seven new retrieval probes, and the stale claim that `ctrl+k` saved a harness design from inside its room is gone — that row was deleted long ago.

## Proof

`make pr-ready` green, including the full `internal/tui3` suite (315s). Laws, manual gates and the light gate green.


AbirAbbas:
Verified against `dev` at the current head (green CI, adversarial read, driven on the real binary). The move itself works: `ctrl+k` kills to end-of-line in every box (including home, the box it was missing), `alt+k` opens the switcher on home and inside a conversation, `alt+shift+k` reverses, the rule line and help sheet say `alt+k`, and `Chats ▾` is gone. Three things block it.

**Blocker — `/settings` still tells people `ctrl+k` opens the switcher.** The quick-switch description says `… Ctrl+k always opens the list and waits for your choice` (`internal/tui3/settings.go`, and the mirror in `internal/config/settings.go`: `Ctrl+k always browses before opening a chat`). `ctrl+k` now **deletes text**, so someone reading that and pressing it in the filter loses their input. There's also no configurable switcher binding — `ui.quick_switch` only governs `ctrl+tab` — so a person can't restore `ctrl+k` even though the setting implies it.

**Blocker — shipped manual pages still document the retired binding/control.** `internal/manual/chat/keys.md:3172` says `Ctrl+K opens Chats`; `tasks.md` (around :5412 and :5437) still names the top-bar `▾` picker and "browsing Ctrl+k"; `permissions.md:177` still names `Chats` as visible furniture. Same manual-law hole as elsewhere — the gate checks the name is mentioned, not that the claim is true.

**Blocker — deleting `Chats ▾` removed the switcher's only mouse door.** `tabFold` is now inert (`door()` returns false for it in `internal/tui3/chattabs.go`), and on `dev` clicking `Chats ▾` called `hopOpenAll` to raise the card. A mouse-only user can still scroll hidden tabs or go through Home, but can no longer open the switcher card itself. The `+N` count fitting is otherwise correct.

Should-fix: "the line kills travel together" is overstated — five caret editors (`home.go`, `homeexchange.go`, `rewindsheet.go`, `place_tasks.go`, `settings.go`) implement `ctrl+u` as `reset()` (empties the whole field) rather than kill-to-start, so on Home `abcdef` + `left`×3 + `ctrl+u` empties the box instead of keeping `def`. Also worth a note: on a legacy terminal (ESC-prefix, no kitty disambiguation) `esc` then `k` within the ~50 ms window decodes as `alt+k` and opens the switcher — inherent to `alt+` and shared by `alt+w`/`alt+o`, but newly relevant now that the switcher moved off the unambiguous `ctrl+k`.

**Conflict with #1040.** That PR keeps and fixes the switcher on `ctrl+k`; this one moves it to `alt+k`. They conflict in `hop.go`, `keys.md` and `screen.md`, so only one can land.

ZeroPoint95:
Closing. This will be folded into https://github.com/Agent-Field/CodeAF/pull/1040 for simplicity.

Issue #1065, opened 2026-09-16
session, prompt: answers argue with the completion check and hide the deliverable (round 1: initial cleanup)


santoshkumarradha:
> **Round 1 of N — initial cleanup only.** This issue is a first, deliberately narrow pass on how the conversation model shapes its visible answer and how the end-of-turn completion check talks to it. A second round (a different person) will follow with a broader review of `system.md` (ALL-CAPS emphasis, overall length, lean-profile interplay, open-model sensitivity). Do not expand scope here; note anything out of scope as a follow-up for round 2.

## What happened

Session `292ad97cdbcb0c0e` (model `~deepseek/deepseek-pro-latest`, build `dc9e55d34`). The person asked for a per-seat model table. The model delivered it (transcript line 266). Then the end-of-turn reader (`markreader`, glm-5.3) raised "the table was cut off" three times in a row (`[carry on]` at 269, 274, 279). The model reprinted twice, then its final visible answer was:

> The observation is mistaken. The complete side-by-side table … has been delivered **four times** in this thread … I will not re-print it a fifth time to satisfy a loop …

The person never saw the table: a finished turn folds to question + trailing answer (`internal/tui3/workfold.go`), so every earlier table sat inside the `worked · N steps` chip and only the rebuttal stood.

## Root causes

1. **The reader sees 600 bytes of the answer.** `checkpointDigest` clips the last assistant message to `checkpointSaidBytes = checkpointSketchBytes = 600` (`internal/session/checkpoint.go:429`). Any answer longer than ~600 bytes reads as "cut off mid-sentence". The `…` clip marker is too subtle; the reader prompt's "abbreviated content is unknown" did not hold.
2. **The carry-on lead tells the model to explain back.** `checkpointCarryOnLead` (`checkpoint.go:912`): "If the observation is mistaken … explain the evidence briefly, and finish". The explanation becomes the turn's visible answer.
3. **Same false claim carried on 3×.** `checkpointCarryOnCap = 3` re-fires an identical observation class after the model has already rejected it.
4. **The prompt never tells the model what the person sees.** `internal/session/prompts/system.md` has no rule that only the final message survives the fold, and nothing separating codeaf-injected bracketed messages (`[carry on]`, `[taking stock]`, `[silent]`, `[still asked]`) from the person.
[...]

PR #1067, opened 2026-09-16
session, prompt, tui3: a wrong completion check is answered with [no change], and the answer stands (#1065 round 1)

santoshkumarradha:
Round 1 (initial cleanup) of #1065; round 2 follows.

Scope is sections A–D of the issue, plus what the repository rules or review required.

## Taken over

The review fixes are four commits on top of `c8346300`:

- `prompt: the messages-from-codeaf list names every note codeaf writes`
- `session: no change skips judgment only when it answers the reader`
- `test: the model stub replays the no-change fold`
- `prompt: the standing-news tag stays with the event, not on the page`

The third commit adds a deterministic replay through the real binary: eleven `read`
rounds, a 30-row answer, one rejecting completion-reader call, and one exact
`[no change]` reply. The headless and tmux runs both reached that complete path.

## What changed

**A. Harness**
- **The completion check sees more of the answer.** `checkpointSaidBytes` goes from 600 to `8 * 1024` bytes, or less when the person's request already fills most of the digest. `checkpointClipSaid` takes the room that is left. If it still has to cut, the text ends with `[clipped by codeaf: N of M bytes]` instead of a bare `…`, and the marker counts toward the limit.
- **New reply token.** `session.NoChangeReply = "[no change]"`, read in one place by `IsNoChangeReply`.
- **When the token counts.** In `checkpointReopen`, `answeredNoteUnchanged` reads the transcript. The token counts only as the very next reply to the check's own `[carry on]` note; a volatile session note in between is skipped over. It does not count after tool calls, after the person's own message, after a load or ask nudge, or with other words beside it. When it counts, the turn ends with no further check call.
- **After a token reply.** Only a token that answered the reader's note suppresses the route judge and gives memory learning the answer before the token. In every other context, the token is judged and learned from as an ordinary reply, and another completion check can run.
- **Where the token is hidden.** Display remains deliberately context-free. The transcript keeps the token for the model, while `entryRows` gives every tool-less exact-token reply 0 rows and the live feed empties every such reply at its response boundary (`feed.confirmResponse`), whether or not it counted for the completion check. Either way, `deriveWorkfolds` leaves the settled answer before it in view. `workfold.go` is unchanged.
- **Old journals.** The previous carry-on note is kept as `explainingCheckpointCarryOnLead`, so journals that contain it still hide it.

**B. Prompt text.** `checkpointCarryOnLead` uses the issue's four lines. `checkpointRemainsAsk` adds that text marked as clipped is never evidence the person saw a cut-off answer.

**C/D. `internal/session/prompts/system.md`**

`# Tone` and `# Delivery` are replaced by four sections, and the five rules they carried are restored inside them.

- **`# The answer`**
  - Once a turn ends, only its last message stays in view. The plan and the one-sentence status before each tool batch are still shown while the turn runs.
  - The deliverable goes in that last message. Evidence and blocking details stay complete.
  - Restored: match the person's task and vocabulary; state uncertainty at the claim it affects and choose the safe option.
  - "Done" is never a compiling scaffold or a narrowed test, and half-solved work is never handed back.
- **`# Answer or change`**
  - Questions, options, comparisons and tables are answered in the reply itself, not in a file, unless the person asks for a file.
  - No edits until the person asks for a change.
  - Renamed from `# Words or work`, which collided with the existing `## Work or words`.
- **`# When corrected`**, as the issue wrote it.
- **`# Messages from codeaf`**
  - Names the user-role tags codeaf writes into the conversation: `[carry on]`, `[taking stock]`, `[silent]`, `[stuck]`, `[folded …]`, `[context compacted]`. It no longer claims that any other opening is the person's; `[image #N]` is identified only as the person's attachment.
  - Does not name the tool result's `[held]`, the private reader's `[still asked]`, or standing news's `[something you set up fired]`: that line carries its own instruction under the event, and `TestTheSteeringLineReadsAsNewsAndNotAsARequest` refuses a second copy on the page (the prompt diet's WITH THE EVENT class).

Other prompt changes:
- `…act and offer to unwind it` now continues `when the person asked for a change`.
- `NEVER yield while actionable work remains` becomes `Don't end the turn while work the person asked for remains`.
- `promptprofile.go`: all four sections survive the lean profile. `Messages from codeaf` also has an explicit `keeps: true` row.

**Prompt budget raised, on purpose.** Against `origin/dev`, fixed grows from 53,276 to 55,280 bytes and lean from 45,051 to 47,055: 2,004 bytes on both arms. Each arm already had 15 bytes spare, so each waiver rises by 1,989: `fixedPrefixWaiver` 5,291 → 7,280 and `leanPrefixWaiver` 13,566 → 15,555. Both sit exactly on the new measurement, with the original dated note extended by the review measurement. The owner signed off on the raise by taking the PR over; the page's overall length stays round 2.

**Manual.** New section in `internal/manual/chat/starting-codeaf.md`, "It said my answer was cut off when it was not · … · [no change]". It covers the 8 KB limit, the clip marker, when `[no change]` counts, the fact that a lone token is never drawn even when it did not count, and the `--once` caveat.

**Tests**
- **`internal/session/nochange_test.go`:**
  - the table reaches the check whole;
  - a long request still leaves the answer cut with the marker, never a bare `…`;
  - the UTF-8 cut is exact at both byte parities;
  - `[no change]` ends the turn and the table is the last replayed answer;
  - a token without a carry-on does not suppress another check;
  - a token after tool calls is checked again;
  - the memory reflex gets the settled answer for a valid rejection and the token itself when it did not answer the reader;
  - a table of shapes for which replies the token counts on.
- **`internal/session/messagesfromcodeaf_test.go`:** follows the production constants and fails by source name when the prompt omits a session-authored user-role tag or names one of the voices the page must not carry: `[held]`, `[still asked]`, and the standing-news frame.
- **`internal/tui3/nochange_test.go`:** the fold's answer is the table, and text that merely mentions the token is kept.
- The carry-on, clip and gate tests fail against the earlier code. The new provenance test also failed against the reviewed page before that page was corrected.
- **Updated:**
  - `promptprofile_test.go` (kept headings)
  - `prefixbudget_test.go` (waivers and the extended dated measurement)
  - a stale comment in `carryonwait_test.go`
  - `task_ending_e2e_test.go`, which now uses `childWasHanded` so the system page's tag list isn't mistaken for a note sent to the worker

## Test evidence

Original evidence on tested commit `c8346300`, on the Spark:

| Command | Result |
| --- | --- |
| `make pr-ready` | exit 0 (session ok 180.9s, tui3 ok 312.3s) |
| `make test-report PKGS=./internal/session` | 4156 passed, 0 failed, 7 skipped |
| `make test-report PKGS=./internal/tui3` | 5245 passed, 0 failed, 9 skipped |

Takeover verification on the review worktree, all with `GOFLAGS=-buildvcs=false`:

| Command | Result |
| --- | --- |
| `make build` | exit 0 |
| `make test-focus PKGS=./internal/session RUN='NoChange\|ClipSaid\|ReaderSees\|CarryOn\|Prefix\|PromptProfile\|Manual\|MessagesFrom'` | exit 0 |
| `make test-focus PKGS=./internal/tui3 RUN='NoChange\|Manual'` | exit 0 |
| `go test ./internal/manual/ ./internal/e2e/` | exit 0 |
| `make test-quick` | exit 0 |
| `make pr-ready` (laptop, `8ece6760`) | see the comment below: light gate green; touched packages `internal/session`, `internal/tui3`, `internal/config`, `internal/e2e`, `internal/manual`, `cmd/codeaf-demo-home`, `test/remote/stub` |

Nothing red was already on `origin/dev`.

## Live check (current head `c8346300`)

Setup: real TUI in tmux, `deepseek/deepseek-v4.1-flash` only, `--one-model`, throwaway HOME for each run, `make build` at `c8346300`. The fixture is 12 seat files in a git repo. Evidence is in `~/1065-live-evidence/c8346300/` on the Spark (transcripts, screen captures, the check's call).

1. **Table ask, run twice** (read the 12 files one per step, then one side-by-side table of about 2.5 KB, no file edits):
   - 14 tool calls each (`ls` plus 13 `read`), 0 edits.
   - The completion check ran once per run. It saw the whole table (last row present, no clip marker) and answered `NOTHING LEFT TO DO` both times.
   - 0 `[carry on]` messages, so the model never had to reply `[no change]`.
   - The final fold shows `▸ worked … 14 tool calls` with the 12-row table under it as the visible answer.
   - `[no change]` never appeared on screen.
2. **Questions-only ask** ("What are the options for merging them into a single machine-readable config file, and how do they compare in a table? Don't change anything yet."):
   - 13 tool calls (`ls` plus 12 `read`), 0 write, edit or bash calls; `git status` clean.
   - Answered in the reply with an options table and a recommendation.
   - Only two tool rounds, below the 10 needed before the check runs, so no check ran.

The `[no change]` reply itself was last seen with a real model on `44f38508`: one false note, the model replied exactly `[no change]`, no second carry-on, and the table stayed in view.

## Deterministic live replay (takeover worktree)

`test/remote/stub` drove the built binary through eleven distinct reads of `./go.mod`, a 30-row table, one reader rejection, and one exact-token answer. The headless `--once` run logged exactly one `[carry on]`, one `reader 1`, and one `no change` branch; as documented, stdout included the token. Two independent tmux runs of the real TUI (ports 18080 and 18081; 8080 was taken on the laptop) logged the same three counts, kept `| seat 30 |` on screen under `11 tool calls` after the fold, and showed `[no change]` nowhere on screen. The screen capture and stub log from the second run are attached in the comment below.

## Round 2 follow-ups (not done here)
- **A correctly placed `[no change]` can still reject a real gap.** The issue asks for this. Whether a second check should confirm the token is a round 2 decision.
- **The display hides a lone token regardless of context, but whether the token counts depends on context.** A token reply to a person's steer is hidden even though the turn doesn't end on it. The token also shows briefly while streaming, `--once` prints it, and `feed.growSettledAnswer` does not filter it.
- **`DELIVERABLES ARE FILES, born on disk`** in Session facts is only qualified from `# Answer or change`, not rewritten.
- **Overall prompt length**, including the budget room this PR used, and the rest of the issue's round 2 list.
- **The live check showed the check complaining about process** ("one read per step was not followed"). Whether `checkpointRemainsAsk` should judge how the work was done, not just what it produced, is open.

Refs #1065


santoshkumarradha:
## Self-review

I read the diff and the code around it again. Fixes are in `7c9e1c78`, which is also the tested commit.

### Fixed
1. **`[no change]` was accepted after tool calls made in reply to the note.** The check was "carried on at least once this turn". So a model could make calls after the note, then send the token, and the turn ended without another read, leaving an answer those calls may have made stale. The check now reads the transcript (`answeredNoteUnchanged`): the token counts only as the very next reply to the reader's own `[carry on]` lead. Evidence: `TestNoChangeAfterCallsMadeForTheNoteIsReadAgain` fails with the old check ("reader was spent 1 times, want twice") and passes now.
2. **The token was also accepted as a reply to a load or ask nudge**, once any earlier carry-on had happened this turn. Now only the reader's note counts; a person's steer or a load nudge in between does not. Evidence: `TestNoChangeAnswersOnlyTheReadersNote`, a table of six shapes.
3. **Dead `checkpointMeter.declined` flag.** It was only set on the path that ends the turn, so nothing ever read it. Removed; the change entry no longer names it.
4. **A long request pushed the reader's last-words section into the digest's final `…` cut.** A request of about 12–20 KB rides at the top of the digest, so 8 KB of last words then went over the 20 KB limit. The final cut removes text from the end and leaves a bare `…`, the marker this PR set out to replace. `checkpointClipSaid` now takes the room left after the request. Evidence: the new case in `TestTheReaderSeesALongAnswerWholeAndACutNamesItself` fails with the old room ("page ends … why this …") and passes now.
5. **The route judge and the memory reflex read `[no change]` as the turn's answer.** A turn that ends on the token no longer starts the judge, and the reflex gets the answer before it. `checkpointLastSaid` skips the token for any later read. No dedicated test for the judge or reflex paths; the judge only runs on tool-free turns, and a carry-on on such a turn needs a wake.
6. **The prompt's tag list was wrong.**
   - `# Messages from codeaf` listed `[still asked]`, which only the reader ever receives.
   - It left out `[stuck]` and `[held]`, which the conversation model does receive.
   - It read as "any bracketed opening is codeaf's", but a person's message can open with `[image #1]`.

   It now lists the five tags that actually arrive and says any other opening is the person's.
7. **The budget numbers were wrong.** The comment said the text cost 1,705 bytes; 1,705 was only how far over budget it went, because both arms already had 15 bytes spare. After fix 6 the measured cost is 1,779 bytes (fixed 53,276 → 55,055, lean 45,051 → 46,830, measured on `origin/dev` and on this head). The waivers are now 7,055 and 15,330, exactly on the measurement, and the comment gives the real figures.
8. **Manual and change entry overstated things.**
   - They said the token "ends carrying on for this request" without saying when it counts.
   - They said `--once` output can show the token without saying a check has to run there.
   - They named the removed flag.

   All three now describe what the check actually accepts.
9. **Weak tests.**
   - The UTF-8 cut was not guaranteed to land inside a character. It now runs over both byte parities of a two-byte run and checks `utf8.ValidString` and the exact marker.
   - `checkpointCarriedOnNote(nil)[:20]` sliced a sentence. The test now asserts exactly one carry-on notice.

### Not a defect
- **Live feed dropping real text:** `IsNoChangeReply` requires the whole trimmed reply to equal the token. `TestWordsThatOnlyMentionNoChangeStayTheAnswer` keeps a reply that starts with the token and one that quotes it.
- **`checkpointClipSaid` byte math:** the marker's room is measured at its widest (the kept count can never exceed the room). The walk back to a character start cannot split one, and room too small for the marker keeps nothing. Pinned by the tests in fix 9.
- **Replaying old journals:** the previous lead is still hidden (`explainingCheckpointCarryOnLead`), and the token did not exist before this PR.
- **Tasks and rooms:** a task never runs the completion check (`InTask`). Rooms draw through the same feed, and `Transcript()` uses the same shaped entries, so both hide the token.
- **Interrupts:** a reply cut mid-token is never equal to the token, so it is drawn as the cut text it is.
- **`[taking stock]`:** a note that lands between the carry-on and the reply means the token no longer counts, so the turn is read again. That is the safe direction.

### Deferred to round 2
- **A correctly placed `[no change]` can still reject a real gap.** That is what the issue asks for ("end carry-ons for that ask"). Before, the reader would re-read up to the cap. Whether a second read should confirm the token is a round 2 decision.
- **The display hides the token without context, but the check needs context.** A lone `[no change]` reply to a person's steer is hidden even though it does not end the turn. The token also shows briefly while streaming, `--once` prints it, and `feed.growSettledAnswer` (text arriving after the turn ends) does not filter it.
- **Contradictions elsewhere on the page, outside section D's two lines:**
  - `DELIVERABLES ARE FILES, born on disk` (Session facts) pulls against `# The answer` and `# Words or work`, which say tables are answered in words with no edits.
  - The new `# Words or work` has a near-twin, `## Work or words`, whose WORDS includes "anything one edit … finishes".
- **Overall prompt length**, including the budget room this PR used.

### Scope
Beyond sections A–D, this PR only adds what the repository rules or the protocol require: the manual section (manual law), the judge/reflex guard (protocol correctness), and the `childWasHanded` test helper (the new section names `[silent]`).

### Tests (`7c9e1c78`, on the Spark)
| Command | Result |
| --- | --- |
| `make pr-ready` | exit 0 (session ok 182.1s, tui3 ok 311.9s) |
| `make test-report PKGS=./internal/session` | 4156 passed, 0 failed, 7 skipped |
| `make test-report PKGS=./internal/tui3` | 5245 passed, 0 failed, 9 skipped |

santoshkumarradha:
## Owner review fixes

Fixed in `c8346300`.

1. **Dropped rules restored**, inside the new sections: match the person's task and vocabulary; choose the safe option under uncertainty; evidence and blocking details stay complete; "Done" is never a compiling scaffold or a narrowed test; half-solved work is never handed back.
2. **"Only your last message reaches the person" reworded.** It now says that once a turn ends, only its last message stays in view, and everything before it folds into the closed "worked" line. This no longer contradicts `# Planning`.
3. **`# Words or work` renamed to `# Answer or change`** in `system.md`, `promptprofile_test.go`, the `prefixbudget_test.go` note, the change entry and the PR body. The manual never named the heading.
4. **`DELIVERABLES ARE FILES` conflict:** `# Answer or change` now says an answer asked for in chat is written in the reply, not a file, unless the person asks for a file. Session facts is unchanged.

**Waivers, from measurement:** the prompt now costs 2,035 bytes more than `origin/dev` (fixed 53,276 → 55,311, lean 45,051 → 47,086). `fixedPrefixWaiver` is 7,311 and `leanPrefixWaiver` is 15,586. The dated note is updated.

**Tests (`c8346300`, Spark):**
- `make pr-ready`: exit 0 (session 180.9s, tui3 312.3s)
- `make test-report PKGS=./internal/session`: 4156 passed, 0 failed, 7 skipped
- `make test-report PKGS=./internal/tui3`: 5245 passed, 0 failed, 9 skipped

santoshkumarradha:
## Live check on c8346300

Setup: real TUI in tmux, `deepseek/deepseek-v4.1-flash` only, `--one-model`, throwaway HOME for each run. Evidence is in `~/1065-live-evidence/c8346300/` on the Spark.

- **Table ask, run twice** (12 reads, then a table of about 2.5 KB, no file edits):
  - 14 tool calls each, 0 edits.
  - The completion check ran once per run, saw the whole table (no clip marker) and answered `NOTHING LEFT TO DO`.
  - 0 `[carry on]` messages and 0 `[no change]` replies.
  - The final fold is `▸ worked … 14 tool calls` with the 12-row table as the visible answer.
  - `[no change]` never appeared on screen.
- **Questions-only ask** ("options for merging them into one config file… don't change anything yet"):
  - 13 read-only calls, 0 write, edit or bash calls; `git status` clean.
  - Answered in the reply with an options table.
  - Two tool rounds, below the check's 10-round threshold, so no check ran.
- **Not exercised at this sha:** a live `[no change]` reply, because no carry-on happened. It was last seen live on `44f38508`. The stricter rule for when it counts is covered by unit tests only.

santoshkumarradha:
@AbirAbbas this is ready for your review. If it looks good to go, please merge it.

- **Scope:** round 1 of #1065, the initial cleanup. @ZeroPoint95 has round 2.
- **Tests on Spark at `c8346300`:** `make pr-ready` passed; `internal/session` 4156 passed and `internal/tui3` 5245 passed, 0 failed in both.
- **Live check** with `deepseek/deepseek-v4.1-flash` on the same commit: passed. Details are in the PR body.
- **CI:** `light gate` passed; `touched packages` was still running when I posted this.
- **Needs sign-off:** the prompt-size waivers went up by 2,035 bytes on both profiles (to 7,311 and 15,586), because the new `system.md` sections are longer.

AbirAbbas:
Taking this over to get it merged — I'll push the fixes onto this branch and merge it myself. @santoshkumarradha @ZeroPoint95 for visibility.

Reviewed against the code with an independent audit. The protocol works and CI is green on `c8346300`; the tests drive a real `Agent` turn and the carry-on test fails when the guard is removed. What I'm fixing before merge:

1. **`# Messages from codeaf` is not an exact list.** It says the five tags are the whole set and "any other opening … is the person's", but codeaf also writes user-role `[folded …]` (compaction), `[context compacted]` (older journals) and `[something you set up fired]` (standing news), and `[held]` arrives as a tool result. As written, the rule tells the model that standing news is the person's request. Rewriting the section so it is true, remeasuring the waivers, and correcting the change entry and PR body that repeat the claim.
2. **The judge and memory skip is context-free.** `withdrawn := IsNoChangeReply(answer)` in `loop.go` fires on every exact token, not only one that answered the reader's note (`answeredNoteUnchanged`). Making it use the same check. The display side (replay and live feed hide any lone token) stays as you described it, as round 2.
3. Noting in the change entry that the waivers now sit exactly on the measurement.

Not changing: the clip, the lead, the four sections and the restored rules all check out; the waiver figures reproduce (fixed 55,311, lean 47,086).

Will run `make pr-ready` here plus a live tmux check on the real binary before merging, and post the evidence.

AbirAbbas:
Pushed the takeover fixes: four commits on top of `c8346300`, head `8ece6760`. The PR body is rewritten to match.

**What changed since your head**

1. `# Messages from codeaf` no longer claims five tags are the whole set or that any other opening is the person's. It names the compaction tags too (`[folded …]`, `[context compacted]`), drops `[held]` (a tool result), and a structural test (`messagesfromcodeaf_test.go`) follows the production constants so the list cannot drift again. It failed against your page text before the edit, naming the three missing tags.
2. `[something you set up fired]` is **not** on the page: `TestTheSteeringLineReadsAsNewsAndNotAsARequest` refuses it, because the standing-news line carries its own instruction under the event and a second copy on the page is one law bought on every request (prompt diet, WITH THE EVENT). The structural test files it with the voices the page must not name.
3. `withdrawn` in `loop.go` now requires `answeredNoteUnchanged`, so the judge and memory skip only when the token answered the reader's note. Both real-Agent tests now also assert what the memory reflex was handed (the table in the honoured case, the token in the after-calls case). Display stays context-free, as your round-2 list says, and the manual says so in one sentence.
4. `test/remote/stub` gained `PROBE-NOCHANGE`, a deterministic replay of #1065 against the real binary, with the tmux recipe in `test/remote/stub/PROBE-NOCHANGE.md`.

**Budget.** Fixed 55,280 and lean 47,055, 2,004 bytes over `origin/dev` on both arms; waivers 7,280 / 15,555, exactly on the measurement.

**Gates on `8ece6760`, this laptop, `make pr-ready`:** light gate green; touched packages green: `internal/session` 297s, `internal/tui3` 360s, plus `internal/config`, `internal/e2e`, `internal/exec`, `internal/manual`, `cmd/codeaf-demo-home`. (A first run had `TestTheCeilingCarriesTheLastSketchIntoTheBrief` hit its 60 s deadline while tui3 and a tmux replay ran beside it; alone it passes in 0.06 s, which is the shape its own message describes.)

**Live replay in the real TUI** (stub on 127.0.0.1:18081, fresh `CODEAF_HOME`, `--model stub/scripted --one-model`, typed `PROBE-NOCHANGE ./go.mod`):

```
carry-on requests: 1
no-change branches: 1
reader branches: 1
[no change] on screen: 0
```

Screen after the turn settled, bottom of the pane:

```
  seat 25  stub/scripted  deterministic replay row 25                                                         
  seat 26  stub/scripted  deterministic replay row 26                                                         
  seat 27  stub/scripted  deterministic replay row 27                                                         
  seat 28  stub/scripted  deterministic replay row 28                                                         
  seat 29  stub/scripted  deterministic replay row 29                                                         
  seat 30  stub/scripted  deterministic replay row 30                                                         
                                                                                                              
    · the ask is not finished · carrying on rather than stopping here                                       
                                                                                                  · 17:46 · 
```

@santoshkumarradha @ZeroPoint95 — ready to merge once CI is green on this head.

AbirAbbas (approved):

PR #1073, opened 2026-09-16
crew: open/all model-family toggle with an all-models preset table

santoshkumarradha:
Adds a model-family toggle to the crew: the same three preset words (`frugal`, `balanced`, `max`) can now resolve against either the open-weight table this build ships, or the whole catalog with closed/frontier models in it.

## What changes

- **`models.crew.source`** (new setting row, `model family`, right under the crew row). Values `open` (default) and `all`. Profile-only, like the tier rows.
- **`crewAllModels`** in `internal/config/crew.go`: the same three presets answered from the whole catalog. The open table is byte-for-byte untouched, and a profile that never answers the row resolves exactly what it did before.
- **`CrewModelsForSource(source, preset)`** plus `CrewSourceAt`; `CrewModels(preset)` keeps its signature and delegates with `open`, so no existing caller rippled. `CrewAt` and `ApplyCrew` resolve through the source row, so the crew word and the five ids it summarizes can never be drawn from different families.
- Manual paragraph under the crew section.

## The two families

**open** (default, unchanged):

| seat | frugal | balanced | max |
|---|---|---|---|
| reflex | `mistralai/mistral-nemo` | `mistralai/mistral-nemo` | `mistralai/mistral-nemo` |
| small work | `deepseek/deepseek-v4-flash-0731` | same | same |
| worker | `deepseek/deepseek-v4-flash-0731` | `z-ai/glm-5.3-flash` | `z-ai/glm-5.3` |
| careful | `z-ai/glm-5.3-flash` | `moonshotai/kimi-k3` | `moonshotai/kimi-k3` |
| mastermind | `z-ai/glm-5.3` | `moonshotai/kimi-k3` | `moonshotai/kimi-k3:high` |

**all** (new):

| seat | frugal | balanced | max |
|---|---|---|---|
| reflex | `google/gemini-2.5-flash` | `google/gemini-2.5-flash` | `google/gemini-2.5-flash` |
| small work | `deepseek/deepseek-v4-flash-0731` | same | same |
| worker | `openai/gpt-5.6-sol` | `openai/gpt-5.6-sol` | `anthropic/claude-fable-5.1` |
| careful | `google/gemini-3.8-flash` | `anthropic/claude-opus-5` | `openai/gpt-6-astra` |
| mastermind | `anthropic/claude-opus-5` | `anthropic/claude-fable-5.1` | `anthropic/claude-fable-5.1` |

Both tables keep the crew's existing laws: the worker column is the spend dial, and the careful seat is a different vendor from the worker in every preset (asserted by a test over both families).

## Design notes

- The toggle changes what the preset words **mean**, not the five ids on disk. Flip the family and re-pick the crew; until then the crew word reads `custom` against the new family, which is true.
- Ids picked off the catalog's own published Artificial Analysis scores (intelligence / coding / agentic) against blended price, seat by seat by the work each seat does.

## Not in this PR yet

- The `/crew` picker's open/all selector (sibling slice, in flight).
- The `docs/changes/unreleased/` entry needs the real PR number: `make changelog-new PR=<n> KIND=added SLUG=crew-model-family`.

## Verification

`go build ./...` clean. `go test ./internal/config/...` green, including: the careful-seat-different-vendor invariant over both families; the all family pinned to the locked ids and passing the tier gate; an untouched profile resolving the open family byte-for-byte and an unknown source word reading open silently; flipping to `all` and re-applying a preset writing the all ids and reading `balanced`.

—

Drafted with [agentfield ai](https://agentfield.ai/github?utm_source=github&utm_medium=pull_request&utm_campaign=drafted_with) · reviewed and owned by the author

## Scope note: what "law 1" means here

The review loop was briefed that a profile which never answered resolves
byte-identically to before. That is true of the **family mechanism**: with no
`models.crew.source` row, nothing about resolving a preset changed.

It is **not** true of the **roster**: the open table's frugal mastermind and its
balanced careful and mastermind seats moved in this same PR, and so did the
shipped default crew (which is exactly the balanced row). A never-answered
profile now gets `kimi-k3` where it used to get `qwen3.8-27b` and `glm-5.3`.

That move is deliberate and disclosed: it is in the change entry, and the manual's
open table shows the new ids. It is kept in this PR rather than split out because
the roster decision and the toggle were made as one piece of work. If a reviewer
would rather see them apart, say so and the roster change comes out into its own
PR with its own entry.


santoshkumarradha:
Merged into the `santos/dev` accumulation branch (not into `dev` directly); the work is preserved there and will reach `dev` in a single PR later. Closing the PR here.

PR #1092, opened 2026-09-17
connect, settings: named, multiple, switchable custom connections (#1089)

### LINKED ISSUE #1089: Custom connections: named, multiple, switchable — in /connect and /settings
Today `/connect` → **Something else** can hold exactly one custom connection, and the experience has three gaps that make custom connections hard to use.

**1. No name step.** The connect flow prompts for an address and a key, then auto-derives the display name from the URL hostname (`modelServiceSlug`). For an IP-literal host there is no usable label, and the slug can collapse to a meaningless name, so the connection and all its models show up under a garbage prefix. There is no way to type your own name.

**2. Only one custom connection.** The custom path is hardcoded to a single `"custom"` id, and `persistConnectedSource` upserts by id — so connecting a second custom service **replaces** the first. There is no way to keep several custom connections side by side.

**3. No easy switching.** A model row is bound to its source, and there is no first-class "active connection" to flip. Moving between OpenRouter and a custom connection means re-navigating the model picker rather than switching a connection.

## Proposal

**Custom connections — named and multi-instance.** "Something else" becomes a reusable connection kind:
- add a **name step** to the connect flow (default = slugged hostname, editable, validated against existing names),
- mint a **per-instance id** instead of the single literal `"custom"`, so several custom connections coexist,
- make `modelServiceSlug` handle IP-literal hosts gracefully instead of producing an unusable name.

**The same add/manage path exposed in `/settings`, driven by the same backend.** The Providers tab (the `models` category, which already leads with the model + lane rows and the service keys) gains the ability to add and edit a custom connection inline — name, address, key — by calling the *same* functions `/connect` uses (`ConnectService`, the probe, the name derivation, `Collides`). One code path, two doors; the interactive sheet and the settings rows both feed `ConnectService`, never a parallel reimplementation. This means lifting a reusable "validate + persist a custom connection" out of `internal/tui3/modelservices.go` into `internal/config` / `internal/connect` so both surfaces call it.

**A connection switcher.** A first-class "active connection" concept: flip the active connection from `/connect`, as a row in `/settings` → Providers, and see connections grouped in `/model`. Switching between OpenRouter and a custom connection becomes one action, not a re-setup.

**Smooth end-to-end for any custom connection.** When a custom connection fronts a model with a different wire contract, recover cleanly (the single `provider`-object-off retry) instead of burning the relaxation ladder and surfacing "the request could not be sent as it was."

## Naming

The public word stays **connection** (matches `/connect` and "your connected accounts"); internally these are services. We deliberately do not call them "providers" — that word already means the OpenRouter lane/routing axis, and a custom connection is a different axis (the whole account) from a provider inside OpenRouter.

## Where each piece lands

- `/connect` — interactive add/validate (name, address, key), rename, switch active.
- `/settings` → Providers tab — inline add/edit (same backend), rename row, active-connection switcher row.
- `/model` — models grouped by connection.
- `internal/provider/endpoints.go` — the wire-recovery fix for custom connections.

## Acceptance

- Connect a custom service and be prompted for a name; it shows up under that name.
- Connect a second custom service; both coexist, neither replaces the other.
- An IP-literal address (`http://127.0.0.1:PORT/...`) produces a sensible name, not a garbage one.
- Add, rename, and switch the active connection from `/settings` → Providers without opening `/connect`.
- A custom connection fronting a non-OpenAI-dialect model completes a turn instead of failing with "the request could not be sent as it was".

santoshkumarradha:
Closes #1089.

`/connect` → **Something else** used to hold exactly one custom connection: the id was the
literal `custom`, so connecting a second custom service replaced the first, the display name
was auto-derived with no way to type your own, and an IP-literal host produced a name nobody
could read. This makes a custom connection a named, repeatable thing, and makes the one the
conversation is talking to something you can flip from either door.

## What changed

**Ids are the persistence vocabulary; the written name is the routing prefix.**
`internal/modelsource` mints `custom` for the first connection and `custom-<slug>` for each
one after, so several coexist and `persistConnectedSource` no longer upserts them onto each
other. `SourceSlug` reads a host whole instead of taking its last dot-separated label, so
`127.0.0.1` answers `127-0-0-1` (not `0`), IPv6 literals keep their groups in order
(`fe80::1` → `fe80-1`, `::1` → `ipv6-1`), a zone id is cut before the host is read, and a
literal that reduces to nothing falls back to the plain custom id.

**One backend, two doors.** The validate-and-persist step lifted out of
`internal/tui3/modelservices.go` into `internal/config/custom.go` (`PrepareCustomSource`,
`ActiveConnectionFor`, `RenameConnectionModels`), so `/connect` and `/settings` → Providers
both call it rather than each carrying its own copy. The add row is offered on both surfaces
whenever a connection can be added — including on an empty profile — and carries an id no
real connection can mint, so it can never collide with one.

**The active connection is derived, never stored.** It is read from the conversation's live
model, not from a slot in a shared profile file, so two conversations on different
connections each see their own. Enter on the active-connection row walks a ring that includes
the default (non-custom) connection, so switching back is the same single action as switching
away. A switch that arrives while a turn is working is deferred and spent when the turn
settles; the panels refresh only when a settle actually moved the conversation, so an idle
settle no longer sends an open panel's cursor back to the top or clears a standing disconnect
confirmation.

**A rename carries everything it named.** Role models persist in the crew tier rows, written
with the connection's name as their prefix. `RenameConnectionModels` re-prefixes every stored
row and writes them in one atomic profile write, validating each new value first, so a rename
cannot half-apply and leave some roles pointing at a name that no longer exists.

## Acceptance (issue #1089)

| Acceptance bullet | Where |
|---|---|
| Connect a custom service and be prompted for a name; it shows up under that name | name step in the connect flow, resolved through `Collides`; `connectionswitcher_test.go` |
| Connect a second custom service; both coexist | per-instance ids in `internal/modelsource`; `custom_test.go` |
| An IP-literal address produces a sensible name | `SourceSlug`; `modelsource_test.go` table |
| Add, rename and switch the active connection from `/settings` → Providers without opening `/connect` | `internal/tui3/settings.go` Providers rows on the shared `internal/config` door; `addrow_test.go` |
| A custom connection fronting a non-OpenAI-dialect model completes a turn | the `provider`-object-off retry already reaches a custom base; pinned here by `internal/provider/refusal_test.go` so it cannot regress |

**`/model` groups models by connection** — the last of the issue's "Where each piece lands"
items. The picker already grouped by connected service (`modelsFor` in `internal/tui3/palette.go`);
what this PR adds is the coverage that keeps it true for *several* custom connections:
`TestTwoCustomConnectionsGroupApartInThePicker`, `TestRenamingACustomConnectionMovesItsPickerHeading`
and `TestACustomConnectionWithNoListKeepsItsPlaceInThePicker`.

**Switching onto a keyless default service is refused**, the way an empty model list already is,
so the switcher can no longer hand the conversation a state whose first send opens a key door
nobody asked for. One helper, `defaultServiceHasKey`, now answers that question for both the
switcher and the key prerequisite.

**A minted custom id carries only plain characters** (`modelsource.IDWord`): lowercase, runs of
anything else become one dash, and a name that reduces to nothing answers `connection`. The
written name and every id already persisted are untouched.

## A mid-session connect now reaches the conversation that is open

Interactive `codeaf chat` runs its conversation on an `engine --daemon` process. That door
registered nothing it built, so a profile broadcast reached no conversation and an agent kept
the account set it was born with. A service connected mid-conversation therefore matched no
member of that set, and its qualified model id fell through to the default service with the
prefix still on it — a 400 about a model no router publishes, read by the person as *the
request could not be sent as it was*.

| in one `codeaf chat` process | before | after |
|---|---|---|
| connect, then a turn in the same conversation | 400 | works |
| connect, `/new`, then a turn | works | works |
| connect, restart the binary, then a turn | works | works |

**This is pre-existing and not introduced by #1089.** Any mid-session `/connect` of any service
on the hosted road behaved this way; custom connections are simply the first case where a person
routinely connects mid-session and is switched onto it in the same breath. The fix is a paired
register/forget tied to the engine session's lifecycle: the engine registers every conversation
it builds — boot, `/new` and `/resume` — and releases it through a new `Engine.Closed` door,
told on both roads a conversation ends. Registering without releasing would grow the held list
for the life of a daemon that outlives every conversation in it, which its own test pins across
eight open/close cycles.

**No billing moved.** The 400 names no upstream, so it classifies as a malformed request
(`ActionReshape`) and ends the turn. The deepseek and openrouter rows beside it in `calls.jsonl`
are the separate title and router lanes, which run alongside every turn and end with
`context canceled` when the turn dies. The turn itself never retried onto another model.

**What the evidence is worth.** One test fails on the unfixed tree: a structural law that reads
the engine door with `go/ast` and counts the conversations it builds against the ones it
registers (`bootEngine builds 3 conversations and tracks 0`). The others pin pieces that did not
exist before the fix, so they cannot fail first by construction. That law is brittle — it counts identifiers inside a function named `bootEngine`, so a refactor that moves
the build behind a helper would make it pass while the defect returned. Its failure messages name
both numbers so a reader can see what it actually checked. A behaviour-level equivalent would
need a live wire harness and would pass before the fix anyway, since the held list is only ever
empty on the engine road.

## Evidence

- `make check` green on this exact head, run on the Spark cluster over ssh from a fresh clone of
  `191cf579a70cbd70ee1db80efdf2caa46ecaec5d`: **99 packages ok, 0 failures**, binary
  53,149,961 bytes (under the 54,600,000 size budget).
- The branch carries `santos/dev` merged in at `3b97f50c`, so that run covers exactly what would
  land rather than a stale base.
- The engine fix was additionally proved on its own head before it was merged here: `cmd/codeaf`,
  `internal/remote` and `internal/session` each green from an isolated clone, plus
  `scripts/laws.sh` over 117 files and 27 packages.
- ~3,000 added lines, of which more than half are tests: `custom_test.go`,
  `connectionswitcher_test.go`, `addrow_test.go`, `modelservices_test.go`,
  `modelsource_test.go`, `refusal_test.go`, `chatv3_track_test.go`, `closeddoor_test.go`.

## What the review passes changed

The branch went through an architecture review, a correctness review, and a final read-only
review of the whole diff, each followed by a fix pass for the findings that were verified
against the code. Three of those are worth calling out, because each is a behaviour a reader
would otherwise have to trust:

- **An open panel re-ranked on every settled turn**, sending the cursor back to the top and
  clearing a standing disconnect confirmation. The refresh is now gated on a settle that
  actually moved the conversation — `TestASettleWithNothingToSpendLeavesTheConnectPanelWhereItWas`.
- **A rename during a working turn could silently eat a pending switch.** The person presses
  enter on the switcher row, is told the move waits for the turn to finish, then renames a
  connection — and the promised move disappeared. The re-spelling of the frozen live pick is
  bookkeeping and may now claim the deferred slot only when it is empty; a pending move always
  wins, carried under the new name if the rename was its own. Three tests cover the three
  cases (pending on another connection, pending on the renamed one, nothing pending).
- **Nothing tested that a second add mints a second id.** If the mint had regressed to
  answering `custom` again, `persistConnectedSource` would have replaced the first connection
  by id and the suite would have stayed green while the "both coexist" acceptance bullet
  broke. `TestPrepareCustomSourceMintsAFreeIdForEveryAdd` now walks `custom`,
  `custom-<name>`, and the numeric tiebreak twice.


santoshkumarradha:

PR #1094, opened 2026-09-17
a node retry avoids the upstream provider that just failed

santoshkumarradha:
## What failed

A node call that failed on the upstream machine OpenRouter routed it to was
retried up to three times as the same request. Each fresh attempt was built
with fresh knobs and — under the simple routing row — no `provider` object on
the wire at all, so the router's own default routing answered every attempt
the same way, all three landed on the same upstream, and the node ended with
`after 3 node call attempts: provider ended the response with
finish_reason=error` having produced no work.

## What the retry now does

An attempt that failed on a NAMED upstream — a relayed 5xx (`APIError.Provider`,
which is where a `finish_reason=error` reply is normalized to), or a cut stream
(`StreamCut.Provider`) — records that machine and the next attempt asks the
router not to use it. The names accumulate across the attempts of one call, so
the third attempt avoids the machines that failed the first two.

- The list rides the call's context (`provider.WithRetryAvoid`) the way every
  other per-call fact does, and is read once into `callKnobs`. It is built in
  the node's own retry loop, handed over per attempt, and gone when the call
  ends — nothing reaches a later call or a stored setting.
- On the wire it is `provider.ignore`, merged into whatever preference object
  the request already composed rather than replacing it: the sort word, a
  ranking and a demand all survive beside it. `only`, `order` and `ignore` are
  narrowed through one shared field law (`providerPrefs.strike`), which the
  per-call refusal list already followed, so two lists cannot carry two answers
  about how the three fields fit together.
- A machine a demand still names is never also written into `ignore`, and a
  demand the veto emptied stops being a demand.
- A person's pin for the model wins outright: nothing is ignored.
- A base that has not shown it carries a preference object, and a routing row
  of `off` — including the direct, one-road services that resolve to it — are
  left exactly as they were, so those requests are byte-for-byte unchanged.
- A failure that names no machine changes nothing; the retry goes where it
  always went.
- When every attempt fails, the final error names the machines that were tried.

Every request that does not fail this way carries an empty list and is
unchanged.

## Tests

New tests beside both touched packages: the retry carries the list and
accumulates it; a cut stream's name is carried; a failure naming nobody avoids
nothing; a relayed 4xx names no machine; the list does not outlive its call, at
the node loop and at the request body alike; the veto goes out on the next body,
rides beside a ranked road, stays off a road that asked for no preference
object, and never empties a demand it joined.

## How it was tested

Spark job `20260917-063523-000531`, revision `e1d0e75248273cbb1eddfac983313fcc7ff871b0`:
`go build ./...`, `go vet` and `go test -count=1` over `./internal/exec/...`,
`./internal/provider/...` and `./cmd/codeaf/...` — all green.

```
ok  internal/exec        65.5s
ok  internal/exec/bare    6.8s
ok  internal/provider    71.7s
ok  internal/provider/pool 0.5s
ok  cmd/codeaf           71.7s
```

https://claude.ai/code/session_01Sb7knukoGMPXo8kK876MsU


santoshkumarradha:
Pre-merge run on the exact head e7d65a51784abf460835d99ac4f0d7d4de21bc38: Spark job 20260917-064054-000535, go build ./... ok, go vet ok, go test -count=1 ok for internal/exec, internal/exec/bare, internal/provider, internal/provider/pool, cmd/codeaf.

PR #1095, opened 2026-09-17
telemetry: anonymous usage counts, off with CODEAF_TELEMETRY=off

santoshkumarradha:
Anonymous usage counts for codeaf: on by default, a notice printed once, off with `CODEAF_TELEMETRY=off`. The contract is `docs/TELEMETRY.md`; this PR is the whole pipe from the binary to the relay's door.

## What is sent

Four events, all coarse: `first_run` (once per install), `session_started`, `session_ended` (banded counts of turns, model calls, tool calls, failures, cost, duration, stop reason, exit code) and `fault` (a 16-hex fingerprint of codeaf function names at the crash site, never the message). Common properties are version, channel, OS, arch, usage context (local, CI, container) and install method. The distinct id is a hash of a random install id; no person profile is created and geolocation is off.

Never sent, and tests prove it: prompts, code, file names, paths, repo names, keys, email, IP, machine name, model slugs, raw error text. The relay strips unknown keys as a second wall.

## How it works

- `internal/telemetry`: the events, an allowlist, a local spool (`~/.codeaf/telemetry/spool.jsonl`, 50 per POST, 7-day age, 1000-line cap), one deadline-bounded flush at exit, and the opt-out ladder. Session counters are three atomic adds on the hot paths and count whether or not telemetry is on; only sending asks the ladder.
- `cmd/codeaf`: lifecycle in `execute` for chat, resume, do, exec, run and `plan run`; `codeaf telemetry status|show|on|off`; fault events from `reportFault` (main) and from `guard.Recover` through a hook (goroutine).
- `internal/config`: the `telemetry` setting row. A project file can only turn it off, never on.
- `scripts/install.sh`: records how the copy was installed and prints the notice; sends nothing.
- `scripts/release_downloads.py` and a daily workflow: public GitHub release download counters as one gauge event per asset per day.

## Off switches, in the order checked

`CODEAF_TELEMETRY=off|0|false`, `DO_NOT_TRACK=1`, `telemetry = off` in the project file, `codeaf telemetry off`, an empty `CODEAF_TELEMETRY_ENDPOINT`. Unstamped or dirty builds and test binaries never send. Nothing is sent before the notice has been shown.

## Verified

- Full `make check changelog-check test-laws` on Spark on this exact head (job id and result in the last comment). One failure is pre-existing on `dev` and untouched here: `internal/release` `TestPublishedPinExamplesGiveVersionToBash`, because `README.md` on `dev` no longer carries a `| VERSION=` pin example. It fails the same way on a clean `dev` checkout.
- A stamped build of this head ran two real task sessions against the production relay: a clean one arrived as `done` with exit 0 and the model-call band filled; an interrupted one arrived as `interrupted` with exit 3. The relay's own checks (schema, size, enum, unknown-key stripping) were exercised end to end.
- `internal/telemetry` and `internal/guard` pass under the race detector. Binary size stays under budget.

## Known limits

- `turns` is counted where the chat loop seals a turn, so task sessions report the 0 band for it.
- The notice is not printed under `--json` or when stderr is not a terminal in chat mode, and until it has been printed nothing is sent.


santoshkumarradha:
Verification evidence for head `c828d0c5f`:

- Full suite on Spark (aarch64, go1.26.5), fleet job `20260917-101803-000555`: `make vet fmt-check test` on the exact head in its own clone. 96 packages ok. The one failure is `internal/release` `TestPublishedPinExamplesGiveVersionToBash`, which fails identically on a clean checkout of `dev` at `09299019b` (README.md has no `| VERSION=` pin example since the README rewrite); it is not touched by this PR.
- Remaining `make check` targets, fleet job `20260917-102358-000556`: `test-packed-manual` ok, `size` ok (`bin/codeaf` 53,018,889 bytes under the 54,600,000 budget), `changelog-check` ok, `test-laws` ok.
- `internal/telemetry` and `internal/guard` pass under `-race`; `cmd/codeaf` passes with `-count=2`.
- Live: a stamped build of this branch ran two real task sessions against the production relay. The clean one arrived in PostHog as `session_ended` with `stop_reason=done`, `exit_code=0`, `model_calls=2-5`; the interrupted one as `stop_reason=interrupted`, `exit_code=3`. `first_run` and `session_started` arrived once each per install. The events were personless, carried only allowlisted keys, and the spool was empty after exit.

santoshkumarradha:

AbirAbbas (approved):

PR #1099, opened 2026-09-17
session: no git command runs without a directory, and the checkout guard reads a linked worktree's own share

santoshkumarradha:
## What

Three holes around git commands that could act on a checkout the run does not own.

- `worktreeDirtIn` (`internal/session/task_run.go`) and `UnsavedEditsNote` (`internal/session/taskpreflight.go`) built `git -C <dir>` themselves; `git -C ""` is a no-op, so an empty directory ran the command where the process stood, past the refusal in `gitWith`. Both now go through it (`gitContext` is `gitWith` under a caller's context; the refusal and the pinned environment live there once). `TestStatusReadsWithNoDirectoryNeverRuns` proves neither runs git with no directory.
- `taskTree.releaseLanded` (`internal/session/groundladder.go`) removed a worktree and deleted its branch with no empty-root guard and without the root lock, unlike its mirror `releaseKept`. It has both now; `TestReleaseLandedLeavesAnEmptyTreeAlone` and `TestReleaseLandedRemovesItsWorktreeAndBranch` pin each side.
- The package's checkout guard (`hermetic_checkout_test.go`) read branches and worktree registrations for the whole shared clone, so from a linked worktree a sibling's ordinary task/* churn was reported as this run's damage, and its live-codeaf probe walked `/proc`, which does not exist on darwin. From a linked worktree the branch reading is skipped and the worktree reading keeps only registrations under the checkout; the probe reads the process table. `TestTheCheckoutGuardDoesNotReportASiblingWorktreesBranch` runs the shared-clone case for real and shows the same churn is still reported from the main checkout.

## Verification

Spark job on the final head: posted below when it completes. Not merged; awaiting the owner's look.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #392, opened 2026-09-02
session: the checkpoint tests script one queue, so the namer #333 runs ahead steals the turn's steps


santoshkumarradha:
## Problem

Three tests in `internal/session/checkpoint_test.go` fail about one run in five, alone
and under load alike, since PR #333 landed on `dev`:

- `TestAContinuationSayingNothingIsLeftDropsTheCeilingHandover`
- `TestTheCeilingIsDroppedWhenTheReaderAgreesNothingRemains`
- `TestADowryOfMachineMarkupIsRefusedAndNeverBecomesTheName`

Two signatures:

```
the turn ended as a assistant saying "(unscripted)"
the task was announced as "this looked like work, so task 1 started: Working through the", want the person's own words
```

Bisected: clean at 2183719a (#279), 0 of 100 on each; failing at 69029c4e (#333), 8, 8
and 21 of 100. Not in known-red, so it reads as a fresh regression to everyone who hits
it, and it has already cost several verification rounds tonight.

## Mechanism

#333 names a task before it is announced. `handOverRunningTurn` now starts the namer
first, on its own goroutine, before the handover's own brief and hand-off calls, and
`launchRouteTask` adopts the name if it has landed by then. That is the design and the
code does it correctly.

The tests' scripted completer is a single positional queue: every call, whoever makes
it, takes the next step. It does not recognise the namer's ask. So the namer's concurrent
call takes a step scripted for the turn. When it lands on the slot holding the final
answer, the turn runs off the end of the script and ends on "(unscripted)". When it lands
on a grinding step, the namer is answered with that round's tool call, the name cleaner
trims it to "Working through the", and the task is announced under it. Instrumented
traces show both, with the namer's call recognisable by its own system line.

The Dowry test's own comment says it reads the announcement "before the namer has had
its second at it", which is the assumption #333 removed on purpose.

`taskname_test.go` met the same race earlier and already routes the namer's call by the
shape of its messages rather than by position. The checkpoint tests never got the same
treatment.

## The law

A scripted completer answers a concurrent errand by the shape of its request, never by
its place in the queue. A test asserts on a sequence it controls, not on which goroutine
the scheduler ran first.

## Fix shape

Test-side only, no code change:

- `scriptedCompleter` gains an optional aside: a function consulted before the positional
  step that may answer a request without spending a step.
- `checkpointAgent` installs an aside that recognises the namer's request (the existing
[...]

PR #1104, opened 2026-09-17
session: the carry-ladder fixtures wait for the mark's drawing, not race the scheduler

santoshkumarradha:
Turns `dev` green. `TestASuccessfulHandoffJournalsTheRungThatSuppliedTheBrief` (`internal/session/carryladder_test.go:118`) failed in a full `make check` on `dev` at `da6e7a0e`: the worker's brief was the running model's draft instead of the mastermind's written handoff. It passed when run alone. Test-only change; no product code moves.

## Cause

The suspect was an errand (namer, title) stealing a scripted step, as in #392. That is not what happens. The mark's drawing is a sidecar (`sidecar.go`): it starts at the boundary that crosses the net and the turn goes straight on. How many rounds pass before the drawing lands is up to the scheduler, and a scripted round costs nothing. On a busy machine the turn ran 41 to 54 rounds against a 53-step script, the writer's ask landed past the end, the scripted completer answered with its past-the-end line `(unscripted)`, the handoff rung read as degenerate, and the carry ladder fell back to the draft.

## Fix

`checkpointAgent` now puts the existing `watchReadings` watch on the turn, and the checkpoint and carry-ladder tests submit through `watchedContext`. `answerWhenQuiet` holds each request that carries the belt until the readings beside it have landed, the order a real turn has, where a real model spends seconds per step. No assertions in existing tests change.

Regression test: `TestAHandoverWaitsForTheDrawingBesideItInsteadOfRacingIt` holds the drawing back by 200 ms and requires that the turn does not spend its script racing it and that the writer's rung still supplies the brief.

## Evidence (Spark, over ssh)

- Negative control: with the `watchReadings` line removed, the new test fails 3/3 (`the turn made 54 rounds while the drawing was in flight`). With it, 20/20 pass.
- `TestASuccessfulHandoffJournalsTheRungThatSuppliedTheBrief` 200/200 under 40 CPU burners with the fix.
- Full `go test -count=1 ./internal/session/`, **twice**, each from a fresh clone of `8f7fb007afe9a4a5e06cf1a005c4603b4a80cdee`: both `ok` (178.8 s, 179.5 s).
- The later commit on this branch adds only the changelog entry.

## Overlap

#1099 (`fix/session-git-holes` → `dev`) touches the same package. Whichever lands second should rerun `internal/session`.


santoshkumarradha:

PR #1107, opened 2026-09-17
chat: the model picker is a table with headed columns

ZeroPoint95:
`/model` lists 600 models. You can't scan it. This makes it scannable.

## Before

```
inference-net/schematron-v2-turbo        $0.03/$0.15 per M · 128k
~openai/gpt-astra-latest                       $10/$50 per M · 1M · sees
inclusionai/ling-3.0-flash-vl   $0.06/$0.18 per M · 131k · sees · watches
nex-agi/nex-n2.5-mini:free                                 262k · sees
```

Every figure correct. No two rows put the same fact in the same place.

## After

Real renders, owner's live catalog, 108 cols. Opens alphabetical:

```
 › /model filter by name
  model ↓                                   in/M   out/M  window   elo  inputs
  deepseek/deepseek-chat-v3.1              $0.25   $0.95    163k  1135
  deepseek/deepseek-r1                      $0.7    $2.5     64k
  deepseek/deepseek-v3.2                   $0.27    $0.4    163k  1187
› deepseek/deepseek-v4-flash               $0.07   $0.14      1M  1220
```

**`alt+s` sorts by the next column, `alt+shift+s` walks back.** Every column is two
presses — its own direction, then reversed — so one key reaches every order:

```
  model                                     in/M  out/M ↓  window   elo  inputs
  ibm-granite/granite-4.0-h-micro         $0.017    $0.11    131k
  deepseek/deepseek-v4-flash-0731          $0.06    $0.12      1M  1252
  openai/gpt-oss-20b                       $0.03    $0.13    131k   947
```

**`→` opens a model's providers — two levels, their own table, their own sort:**

```
  deepseek/deepseek-v4-flash  coreweave   0.4s      1M   24
    auto                                           auto-route based on /settings — coreweave now
    openrouter
      provider    first ↓  t/s    $/M    up  note
›     coreweave      0.4s   24  $0.28   99%  tail 3s
      cloudflare     0.8s   58   $1.3  100%  no tools
      deepinfra      1.2s   27  $0.18   99%  out ≤ 65k
      default
```

Narrow frames fall back to the old tail — it's better per row, only worse down the
page.

## What else moved

**Columns, not verbs.** `sees` / `hears` / `watches` / `draws` / `films` / `speaks`
were six invented words each carrying a side *and* a thing. Now the catalog's own
nouns under `inputs` / `outputs`. Blank = text in, text out.

**Enter no longer closes the list.** It applies and stays open, so you can compare
two models and change back. `esc` closes.

**`auto` and `openrouter` sit together** — they're the two things you're choosing
*between*, and under `simple` routing they send the same thing. Dropped the bullets
(they claimed a distinction the wire doesn't make) and the dim paragraph that
re-printed each row's own numbers in prose. `enter` on `openrouter` opens it and
lands on `default`, which reads as "these exist, and you've chosen not to pick".

**The filter box searches names. Only names.** It used to also take `@cloudflare`,
`<1s`, `>50t/s`, `$<0.3`, `fp8`, `tools`, `fast`, `cheap` — a query language you had
to read a manual page to know existed. Those were questions about facts, and the
facts are columns now, so you can just look. Type `$<0.3` and you get an empty list,
not a silent reinterpretation.

**`→`/`←` stop fighting the caret.** They're the tree's keys unless you're mid-typing
(within 600ms of the last change to the box). Leaving an open fold used to cost one
press per character typed. Any edit takes the caret back, and `ctrl+b`/`ctrl+f` do it
without changing your query.

**One order for the foot, at every level** — grouped by what each key *moves*:

```
↑↓ pick · ← back · → providers · alt+s sort · enter <verb> · ctrl+t effort · esc
```

**Four things were plain broken.** `→ providers` and `ctrl+t` did nothing on home's
model list while the UI advertised both. The provider heading was drawn but not
counted, so the fold could overflow. `←` on `default` closed two levels at once.
And `openrouter` didn't name `→ providers` when nothing was measured, though the key
worked.

## Two bugs worth calling out

**The sort read the wrong lane.** `via`, `first` and `t/s` cells come from the lane a
row *names*; the sort asked `bestLane` instead. So rows with **blank** cells carried
real numbers into the sort and the blanks didn't group. One reading now serves both.

**An alphabetical default buried the best match.** `gpt` returned
`anthropic/claude-gpt-echo` above `gpt-5-classic` — the search itself going wrong.
The name column now orders by relevance first, alphabetically only inside a tier.

Both found by trying it against the real 521-row catalog rather than a fixture.

## Proof

`make pr-ready` + `make test-laws` green. 51 new tests, 30 new manual probes.
Manual updated, including a page that says the filter grammar is gone and points at
the columns instead — the old keystrokes are in its heading so they still retrieve.

Also fixes two `cmd/codeaf` update tests that failed on every macOS laptop and
passed in CI (`t.TempDir` gives `/var/...`, which the update door resolves to
`/private/var/...` on purpose). Product untouched.

Still red on clean `origin/dev`, not mine: `TestTheOpeningHintNamesBothDoors`,
`internal/release`'s ledger test (#1060).


AbirAbbas:
Drove this in a real terminal against a throwaway home with a live catalog, at 140 and 80 columns, and everything the description claims holds up: the headed table, `alt+s` / `alt+shift+s` cycling through every column both ways, the provider fold two levels deep with `←` on `default` closing one level, enter applying and staying open, esc closing from any depth, `$<0.3` giving an empty list, `gpt` leading the search, and the narrow fallback. A missing ELO draws blank. Laws, the untagged e2e needle test, manual, cmd/codeaf and all 49 new picker tests are green here. The filter-grammar regression test is a nice touch.

Three things I would like fixed before it merges, all small:

1. **The task composer's model list draws a sort arrow it cannot move.** `composerPickKey` hands every key straight to `navigate` (`internal/tui3/composerlayer.go:558`) and never goes through `foldKey`, where `alt+s` lives. The other three doors do. So the composer draws the headed table with `↓` on `model` and the chord is swallowed. Your own comment in `palette.go` says an order changeable from one door and not the others "would be four lists again"; this is that.
2. **`internal/manual/chat/commands.md:978` says a plain text chat model shows `text` under both columns.** The renderer skips text (`internal/tui3/models.go:1000`), and live rows like `deepseek/deepseek-r1` draw an empty `inputs` cell. `models-and-cost.md` gets it right; the change entry repeats the wrong claim. This is the page the in-product model answers "what do these columns mean" from, so it will tell people to expect a value in a cell that is always blank.
3. **`internal/manual/chat/models-and-cost.md:136` still says esc "leaves the picker and changes nothing" and that "the model in use" comes back as it was.** That was true before this PR. Now enter commits at once and esc only closes, which the same page says correctly further down. Sections are retrieved alone, so the stale one will be served by itself. The comment at the top of `palette.go` (line 7) has the same stale claim.

Not blocking: the settings sheet's foot never names `alt+s sort` though the key works there; the sort-cycle code block at `models-and-cost.md:3178` skips `out/M` (the prose under it has it); `docs/MULTIMODAL.md` still promises `sees · draws` tails.

Also: dev moved under you this morning (#1108 landed), and this branch now conflicts in `internal/tui3/palette.go`, `lanes_test.go`, `palette_test.go`, `internal/manual/chat/commands.md` and `models-and-cost.md`. Needs a rebase before anything else.

ZeroPoint95:
recheck

AbirAbbas (approved):
Re-verified 2026-09-21 after the review items were addressed: each item checked in the code and on the built binary, gates green locally and in CI at this head. Details in the review ledger.

PR #1112, opened 2026-09-17
README: telemetry section moves before Docs and says why to keep it on

oktaygoktas:
## Summary
- Moves the Telemetry section from after Install to just before Docs, so the pitch flows straight from Install into the product and the disclosure sits with the reference material at the end.
- Adds a two-sentence clarification as prose under the notice block: CodeAF is open source, collects nothing sensitive, and the counts are how it gets better.
- The quoted notice block is unchanged and still verbatim, so it keeps matching the binary, the installer, and docs/TELEMETRY.md.

## Test
`sh test/installer-telemetry.sh`: 35 passed, 0 failed.


santoshkumarradha:
Closing this one: the CLA is unsigned, and #1115 lands the same move in a shorter form (one collapsed block under Docs, no Telemetry heading). Thanks for the push to get it out of Install.

PR #1093, opened 2026-09-17
Pareto crewing: the crew picker, the Model Pool with its judge and relay, and the paper

santoshkumarradha:
One pull request to `santos/dev`: **Pareto crewing** (the crew picker), the **Model Pool** with its judge, relay and mirror, the **paper**, and the fixes folded from #1097–#1102. `santos/dev` (through #1105) and `dev` (through #1106) are merged in, so this diffs cleanly against its base.

**Status:** ready for review on head `fd29d79a5`: full `make check` green on Spark at `cf4ea354c` (107 packages, EXIT 0), and `fd29d79a5` adds one runbook sentence on top (docs only). `santos/dev` `825ef2c41` (#1092) is merged in at `dc180fdfb`. Not to be merged until the owner has looked.

## What this brings

| Piece | What it does | Where |
|---|---|---|
| Pareto crewing | The crew is picked on the cost-quality front, per role and per task, from evidence rather than a fixed table: `picked from` = `table` \| `catalog` \| `learn` | `internal/crewpick`, `internal/config/auto.go`, `/crew`, settings sheet |
| The judge | After a chat task lands, a model outside the crew rates each seat's work; the numbers feed the next pick and are billed to the `judge` seat | `internal/pool/judge`, `cmd/codeaf/poolrecord.go`, `session.Config.TaskLanded` |
| The Model Pool client | Signed, versioned index read daily with a cache and an embedded seed; a text-free outbox pushed under a per-install nonce; `model_pool` = `on` \| `read` \| `off`; `codeaf pool show \| status \| verify` | `internal/pool/{index,pull,poolcfg,outbox,record,tally}`, `cmd/codeaf/pool*.go` |
| The relay | One edge function, one store of per-install daily count/sum/sum-of-squares triples (rows folded in, the join for copies between stores), one hourly Ed25519 signing job, a per-install quota; a GitHub Actions mirror of the signed index every 6 h to the `model-pool` branch | `relay/`, `.github/workflows/pool-mirror.yml`, `docs/design/model-pool/RUNBOOK.md` |
| The paper | *Pareto Crewing*: objective, front and knee, the pool's statistics, the estimator and its constants, the Thompson picker with role-level feedback, the deployment; two referee reports and the response beside it | `docs/design/model-pool/pareto-crewing.pdf` |
| Seat on every usage row | `seat` (reflex, low, worker, high, mastermind, talk, judge) derived at the bank | `internal/session/usage_seat.go` |
[...]

PR #1116, opened 2026-09-17
ci: Model Pool mirror workflow (signed index onto the model-pool branch)

santoshkumarradha:
Adds the scheduled workflow that mirrors the Model Pool's signed index document onto the `model-pool` orphan data branch, plus the signature verifier it runs. Nothing else: no product code, no relay Worker (that ships with the Pareto crewing PR into `santos/dev`, #1093).

## What it does

Every six hours, and on manual dispatch, the job fetches `https://codeaf.agentfield.ai/pool/index.json` and its detached Ed25519 signature, verifies the signature against the public key (a literal, no secret needed), and commits both files unchanged under `pool/` on the `model-pool` branch. The branch holds the signed document and its signature only: no count, no leaderboard, no extracted statistic.

## Why it goes to `dev` on its own

GitHub only runs a workflow that exists on the default branch, so the mirror cannot start until this file is on `dev`. The relay is already live and the `model-pool` branch exists (created by hand today, refreshed hourly from a stand-in job until this workflow takes over).

## Evidence

- Both files are byte-identical to their copies on `feat/auto-crew` (#1093), where the relay's 58 node tests and the verifier run.
- Workflow requests `contents: write` itself; the repository's default token permission is read-only.


santoshkumarradha:

PR #1123, opened 2026-09-18
pareto crewing v2: one dial in dollars, a grade no model can move, and the paper behind it

santoshkumarradha:
## What this is

The crew picker's one dial is now a **stake in dollars** — what a failed task costs you — and the thing it learns from is a **model-free grade** of what the task left in the tree. The paper (`docs/design/model-pool/pareto-crewing.tex`, PDF committed) is the design; the code lands the decision rule, the grade, one scoreboard, learned seat shapes, the relay side, and the manual pages.

## Code

- `crewpick`: `AtStake(front, stake, tokens)` picks the crew with the least `Loss = bill·tokens/1e6 + stake·(1 − quality/100)`. `frugal` = \$1, `balanced` = \$10 over `TypicalTaskTokens` (4.4M, the median of 133 measured runs); `max` = top of the front. `knee` is deleted. `ShapesFrom` learns each seat's volume and in/out from real tasks, shrunk N/(N+30). `PriorFromCells` folds cells of one seat and model by count.
- `internal/pool/grade` (new): gofmt over the files the task wrote, `go build ./...`, `go vet` and `go test` on the touched packages under 2-/3-minute budgets → `Pass`, `Source` (`codeaf/grader` when tests ran, `codeaf/grader-build` when they did not finish), stage and detail. No `go.mod` → no grade, nothing recorded.
- `internal/pool/record`: metric `acceptable` (100/0 per held seat, `source` dim) is what the picker reads; the judge's 0–100 stays under `role_quality` as evidence. `seed.json` re-keyed to `acceptable`/`reviewer`. `index` holds the floor on `installs` when a cell carries it. `internal/pool/shape` (new) reads the usage ledger into per-task seat spend.
- `config`: `AutoShapes` seam beside `AutoOwnCells`; `PoolQualityMetric = acceptable` (bernoulli); catalog prices scaled to \$/1M at the one seam — a dial in dollars makes the scale matter.
- `cmd/codeaf`: the landing hook grades first, then asks the judge, records both, writes `pool/grade-last.json`; shapes are learned once at start-up.
- `relay/`: accepts both metrics, keys `acceptable` under its own name so stored triples never move, publishes graded cells per source, declares both metrics. **Deploy the relay before this client** — an older relay refuses a batch holding an `acceptable` row.

## Evidence

- Fixture catalog (`crewpick` tests): all-family `balanced` moves from **\$2.60/task at Q 95.4** (glm-flash / fable / fable — the crew whose checker ate 75–80% of real bills) to **\$0.65/task at Q 89.2** (glm-flash / qwen-max / qwen-max). \$50 of stake buys the old crew back. `frugal` and open-family `balanced` are unchanged.
- 133 measured runs: high seat mean token share 0.126 vs the default 0.08; median dollar share 0.45 — most of a dear checker's bill is price, the rest is the tail the learned shape now sees.
- Paper: proxies measured on 202 runs (a run's own `settled` claim ρ = 0.02 of a grade; an off-crew rating 0.42); calibrated simulation: Thompson-in-J does not repay exploration where every model has a catalog index — exploration belongs in a capped, separately-vehicled budget (design only here).
- Tests: `crewpick`, `config`, `pool/*` (incl. the new hook test that grades a real module both ways), relay (35), manual gates, `make test-quick` green locally. Touched-package suite running on Spark from `~/src/af-pareto-v2` at `167329d06` — result to follow, plus a live `/task` smoke under `learn balanced` with the branch binary.

## Base

Temporarily based on `fix/judge-unverified-restart` (b1's judge-restart fix, `cc405486a`) at their and the manager's request, so the grader sits on their hook shape; when that merges to `santos/dev` this PR is rebased and their commits drop out of the diff. Relay half already merged as #1129 and deployed.

## Still design only

Price bands, cascades, the capped exploration budget, the switch guard, a custom stake setting.


santoshkumarradha:
Superseded by the per-task Pareto crewing rewrite on `feat/pareto-crew` (PR to follow), which replaces the preset crew and the model-pool paper.

PR #1143, opened 2026-09-18
manual: the status row's ledger is the tree's, not the conversation's

santoshkumarradha:
The chat screen manual says the status row's ledger, meter and job counts all stay the
conversation's. That has been false for the ledger since `c7142ca4` (2026-08-31) made the
bill tree-inclusive, and the same page already contradicts it two screens earlier.

## What is wrong

`internal/manual/chat/screen.md` describes the bill as **what this conversation and its
tasks have spent** where it introduces the groups, and then, in the paragraph about a task
room, says:

> The ledger, the meter and the job counts on the same row stay the conversation's, because
> those are measurements of a session.

The code says the first one. `internal/tui3`'s `spendShown` draws the larger of the subtree
receipt (`session.UsageTree`) and the conversation's own books, and every call a task
subtree makes writes one ledger line rooted at the chat conversation's session id, so a
running node's spend is in the bill whether or not its room is open. `/cost` is where the
two halves are taken apart.

Nothing catches the stale sentence: the Manual CI step checks commands and aliases only
(`manual_test.go`), so a prose claim about a figure can go out of date silently.

## What changed

One paragraph in `internal/manual/chat/screen.md`. The meter and the job counts keep the
sentence they had — those really are the conversation's, and the meter is the only one of
the three with a defined window, which is the reading the compaction decision uses. No code,
no behaviour, no test.

## Where this came from

A read-only diagnosis of what the status row draws while a task subtree is running, which
found the manual claim while establishing that the bill is already correct. The same
diagnosis recommends one behaviour change that is **not** in this PR and is waiting on a
decision: while the chat door is at rest and its subtree is working, the state word says
`idle`, and it could say what the subtree is doing instead — everything needed is already in
hand at draw time, and the cost is about thirteen `idle` assertions in `internal/tui3` plus
the vocabulary table on this same manual page.


santoshkumarradha:
Manager's Spark evidence, fresh clone of the head: go build, go test ./internal/manual/... ./internal/guard ./internal/namelaw ./cmd/codeaf-changes: ok, EXIT 0, log spark:~/src/trees/check-555036d2.log

PR #1144, opened 2026-09-18
resident: the wall refuses a first replan round it cannot hold

santoshkumarradha:
## Why

`internal/resident/grow.go` reads how near the wall is from the job's own measured round pace, and `jobPace` returns 0 for a job with no measured round yet. Zero pace meant "not near", so a FIRST replan round was admitted however close the wall was. It then ran long and the clock killed it mid-round: the job was released with its root unlanded, no delivery gate was ever cut, and broken work shipped. ds1 bought a first round at 40m22s of a 45m wall; awilix ended the same way with a `src/awilix.ts` a build would have caught. (Root cause from the do-loop diagnosis.)

## What changed

- When `jobPace` returns 0, `wallNear` is now read against a conservative floor, `roundFloor(node)`: the elapsed runtime of the leaf that just overran (it has run that long without finishing, and a round replanning the same remainder can hardly cost less), or `noPaceRoundFloor` (4 min, the benchmark's median round, biased low) where no start was stamped. A first round the wall cannot hold is handed over via the existing `RefusedOutOfWall` path — which closes the job out so a gate precedes the wall — instead of being killed mid-flight. The measured-pace branch and every other refusal (rounds, ceiling, rail) are unchanged, and the floor leans toward finishing: the wall refuses a round only when it cannot hold even that much.

## Tests

`internal/resident/grow_test.go`: a first overrun round with no measured pace is refused (`CauseOutOfWall`) when the wall cannot hold the floor, and admitted when it can. `go build ./...` and `go test ./internal/resident/` green.


santoshkumarradha:
Manager's canonical scoped check, fresh clone of c8f9e365 on Spark (login shell, nothing exported): go build, go vet ./internal/resident ./cmd/codeaf, go test ./internal/resident ./cmd/codeaf ./cmd/codeaf-changes ./internal/guard ./internal/namelaw: all ok, EXIT 0. Log: spark:~/src/trees/check-c8f9e365.log. The earlier head eb1d8ed3 failed eleven do-door tests in the same environment (check-eb1d8ed3.log) while trunk passed (check-ctl-a06b53a91.log); fixed by the roundFloor change.

PR #1145, opened 2026-09-18
codeaf: a test's crash fixture stays in a profile the test owns

santoshkumarradha:
A test in `cmd/codeaf` wrote a fake crash — `runtime error: slice bounds out of range [:-1]` —
into a live profile's `chat.log`, where it read as a genuine crash. The test had moved `HOME`
and `CODEAF_HOME` to directories of its own. That is not enough, and the gap is the whole point
of this change.

## What was wrong

`reportFault` appends through `config.ProfilePath(config.ProfileDir(), "chat.log")`, and it has
to: the surface's running log and the crash's append are one file, so a profile that moved
the first must move the second. **`CODEAF_PROFILE_DIR` therefore decides where a fault test's
fixture lands, and it outranks both variables a test naturally reaches for** — an exported
profile wins over `CODEAF_HOME`, which only answers where the state root is when no profile
was named. A harness that exports a profile points every such test at somebody else's log.

## What changed

Tests only. Four tests and helpers now name every variable that decides where their writes go:

| file | what followed the inherited profile |
|---|---|
| `nopanic_test.go` | the fault fixture itself, into `chat.log` |
| `telemetry_test.go` (`telemetryHome`) | the profile's telemetry row through `config.WriteTelemetry`, and the spool read back |
| `chatv3_belt_test.go` | the model catalog (`catalog.LoadLazy`) and the pool's index refresh, both seated by `openV3Process` |
| `vocabulary_test.go` | `pool/install` and `pool/outbox.jsonl`, minted by the launch `do` and `exec` seat |

**Reading found three of the four.** The fourth was found by running the roads instead: every
`cmd/codeaf` test that seats a launch or writes a profile path was run with
`CODEAF_PROFILE_DIR` pointed at an empty directory, and the directory was then read. The other
launch-seating rows in `vocabulary_test.go` stopped at a missing key before any errand started
and wrote nothing across three runs each.

## Failing test first

`faultprofile_test.go` holds the fault case down **from the outside**, because a test that pins
the variable itself cannot fail on the tree that leaked. It runs the fault test as a child
process with `CODEAF_PROFILE_DIR` at an empty stand-in directory and then reads the stand-in.
The assertion is on *where* the file landed and never on a byte of its text, and it reads only
directories it created, so no real profile is touched — not even to check it.

Verified by hand on the unfixed tree: `faultprofile_test.go` alone, dropped onto `6c98ddca`
with none of the fixes,

```
--- FAIL: TestAFaultTestKeepsItsFixtureOutOfAProfileItInherited (0.03s)
    faultprofile_test.go:65: the fault test wrote into the profile the environment named
    instead of one of its own: [chat.log]
```

## What would stop the class, proposed and not done

`cmd/codeaf`'s `TestMain` already gives the package a floor — `isolateTestEnvironment` in
`testenv_test.go` — and that floor pins `HOME` and `CODEAF_HOME` but not
`CODEAF_PROFILE_DIR`; `pinTestEnv` also leaves alone any variable the process was deliberately
started with, which is exactly the harness case. Clearing `CODEAF_PROFILE_DIR` unconditionally
there, the way `internal/tui3`'s `TestMain` already does, would cover all 170 test files in the
package at once. **It is deliberately not in this PR**: it changes what every test in the
package inherits, which is a wider claim than this leak needs, and it wants its own review.

## Evidence

- `go test ./cmd/codeaf -count=1` — **ok, 81.2s, exit 0** — on the Spark cluster over ssh from a
  fresh clone of this PR's head `44791a3da4023680a02bbfe110706348d51eaa41`, rebased onto
  `santos/dev` at `49760fb4`.
- The failing-first run above, on `6c98ddca` with only the new test file added.

## One inconsistency left alone

`vocabulary_test.go` writes the literal `"CODEAF_PROFILE_DIR"` while the other three use
`config.ProfileDirEnv`. It matches the literal `"CODEAF_HOME"` on the line above it in that
file, so it was left as the local idiom rather than made consistent across files.


santoshkumarradha:
Manager's Spark evidence, fresh clone of the head: go build, go vet ./cmd/codeaf, go test ./cmd/codeaf ./internal/guard ./internal/namelaw: ok, EXIT 0, log spark:~/src/trees/check-99ff68c3.log

PR #1147, opened 2026-09-18
test: pin the delivery gate over a grown job root

santoshkumarradha:
## Why

Do-loop diagnosis cause (b), "gate-after-grown", reduced to a proof test. The concern was that a job which grew earlier (an overrun spliced a continuation) might not be gated when it finally lands its top-level root — the shape awilix landed in. Investigation showed the gate is already correct once pacing (#1144) lets the grown job reach its root landing: the delivery gate reads the root's OWN landing, and the growing flag is set only for the leaf landing that grows, never for the top-level job node that later gathers the pieces. So the root's landing arrives with `continuing=false` and is gated. The awilix miss was the wall cutting the job off before that landing, which #1144 fixes.

This PR pins that law so a future refactor can't silently drop it.

## What

`cmd/codeaf/gate_after_grown_test.go` (new, test-only): table test over `shouldGate`. A top-level job node (`Parent==store.RootID`, non-reflex, non-nil outcome, not continuing) is gated; the growing-leaf landing (`continuing=true`), reflex group, non-root child, and nil-outcome shapes are exempt.

Note: at this revision the growing flag reaches `shouldGate` as its `continuing` argument (3-arg signature; there is no `carriedOn` parameter in this tree).

## Tests

`go test ./cmd/codeaf/ -run ShouldGate` PASS. `go build ./...`, `go vet ./cmd/codeaf/` green. Test-only; no production code changed.


santoshkumarradha:
Manager's Spark evidence, fresh clone of edb04cab5: go build, go vet ./cmd/codeaf, go test ./cmd/codeaf ./internal/guard ./internal/namelaw: ok, EXIT 0. Log: spark:~/src/trees/check-edb04cab.log

PR #1206, opened 2026-09-18
review conclusions require an executed Checks: command

santoshkumarradha:
## What changed

A `check:` task on the worker harness could conclude `holds:` or `does not hold:` after reading alone, and a worker could add a leaf without a `Checks:` command, so the review round could reach a verdict without executing anything that bears on the claim.

Three coordinated changes close that:

1. `internal/plandb/store.go` (`Done`): a `RoleCheck` task whose result begins `holds:` or `does not hold:` is refused unless its `trajectory.jsonl` records a step whose command matches one of the task's declared `Checks:`. The store hand-parses the trajectory (it sits below `internal/run`, so it cannot call `run.Trajectory` without an import cycle); the `{kind:"step", command}` shape matches `internal/run/trajectory.go`.
2. `internal/run/run.go` (`addReviewCheck`): when a worker leaf declared no `Checks:`, the review check task inherits the commands the leaf actually ran (from its trajectory), so the checker receives an executable contract instead of an empty one.
3. `internal/session/prompts/bashworker.md`: the worker page now requires at least one `--check` on every delegated task.

The check seat stays read-only: `verifyOnlyBash` and `refuseOutsideDoor` are untouched.

## Tests

Failing-first tests at the seam: `internal/plandb/store_test.go`, `internal/run/review_test.go`, `internal/run/supervisor_test.go`, `internal/session/bashbelt_worker_test.go`.

Verified on Spark against tree `9ba1abb3`:
- `go build ./...` ok
- `go test ./internal/plandb/... ./internal/run/... ./internal/session/...` ok (plandb 13.6s, run 11.7s, session 199.3s)
- `go run ./cmd/codeaf-changes check` ok (9 entries well formed)
- `go test ./internal/guard ./internal/namelaw` ok


santoshkumarradha:
Review (changes wanted before the fold), against the real cell c249 (store and trajectories on Spark, `~/src/doe/peer/c249/profile/v3/projects/-home-santosh-src-bl-c249/16fec8d6b8bb313d/`). The store rule is the right seam. Facts:

**1. The backfill admits every command, and on a real leaf that is 37 commands and 26,312 characters.** Leaf `jcrhzh` ran 37 steps; with no declared checks its check task would carry all 37 in `Checks:` and in its description: `plandb task note …` (step 1), `sed -n` and `git grep` reads, every patch and heredoc write, and the leaf's own `plandb done --agent jcrhzh …` (step 37, 1,404 chars).

**2. The rule is satisfied by any one of them.** `ranDeclaredCheck` returns true on the first exact match, so a check that re-runs the leaf's `git status` line may conclude `holds:`. That is the check c249 ran (one `cat`), now permitted by the store.

**3. Exact string match against how checkers type.** Every command in c249's trajectories is prefixed `cd /home/santosh/src/bl-c249 && …`, by workers and checkers alike. For a DECLARED check such as `go test ./internal/tui3`, the checker's `cd <tree> && go test ./internal/tui3` does not match, the store refuses the conclusion, and the check has no other ending: it runs to the step cap at the plan seat's price. On c248 and c250 the review round was already $4.7 of a $5 cap.

**4. Cost.** `checkSection` tells a check to run every declared check first. With the backfill that is the leaf's whole command list again, per leaf, per fix, on the plan seat. c248: 10 work + 10 check + 2 fix tasks, $4.83 of $5.02 on gpt-5.6-sol. One real small cell's bill before and after, please.

**5. The hand-parsed format.** The new tests write `trajectory.jsonl` by hand in the shape the store parses, so they pass if `internal/run/trajectory.go`'s writer changes. One test in `internal/run` that appends through `appendTrajectory` and then asks `store.Done` for a `holds:` conclusion fails the day the shape moves.

Open questions, in order: what is the smallest rule under which a backfilled contract holds only commands that bear on the claim (the shipped engine has this law already: `invocableChecks`/`runnableChecks`/`repeatableChecks` in `internal/session`, and `internal/run` already imports `internal/session`); what counts as "executed" when the checker's line wraps the declared command; what a check does when the contract is empty after the filter, so it can still end.

Not folding into `harness/work-tree` until 1 to 3 are answered; 4 and 5 can ride the same push.

santoshkumarradha:
Revision folded in (c261), addressing the three review blockers on real c249 data:

- **Backfill breadth:** `internal/run/addReviewCheck` now runs recorded commands through `session.InvocableChecks`, which reuses the shipped audit law (`commandLike` / `approval.Vouchable` / `auditAllowed.CheckBash` / `runnableChecks` / `invocableChecks`) to keep only commands that bear on the claim and could run where the checker stands, dropping writes and the leaf's own `plandb done`.
- **Wrap match:** both `store.ranDeclaredCheck` and `InvocableChecks` strip a leading `cd <dir> &&` so a checker's `cd <tree> && <cmd>` satisfies a declared `<cmd>`.
- **An ending for an unrunnable contract:** the worker check page now offers `check: <one sentence>` when no declared command can run; the store gate only fires on `holds:` / `does not hold:`, so that answer ends the check rather than burning to the cap.
- **Format drift:** new `internal/run/trajectory_store_test.go` appends through `appendTrajectory` against a real c249 `jcrhzh` fixture, so the store's hand-parsed shape cannot drift from the writer.

Note on provenance: the cell capped at $5.01 (partly the pre-#1207 plan-seat billing artifact) but committed the complete fix. I independently verified and fixed one raw-string `\n\n` typo it left unpolished (commit 3785d8ea5). Verified on Spark, tree 24b2f44d: `go build ./...`, `go test ./internal/plandb/... ./internal/run/... ./internal/session/...` (session 201s), `codeaf-changes check` (9 entries), `guard`+`namelaw` all green. The 16 em dashes in the diff are all inside the captured c249 fixture data, not prose.

santoshkumarradha:
Re-review at 3785d8ea5. Still held. The `cd <tree> &&` strip and the writer/reader format test are right and stay. Three facts from running the branch, each reproducible with a ten-line test in `internal/run`:

**1. On the real c249 data the backfill is empty, and the test asserts that as the goal.** `TestC249BackfillKeepsOnlyInvocableChecks` wants `len(got) == 0`. It is empty with the tree present and with `/nonexistent/tree` alike, so it proves nothing about the filter. The reason: every recorded command is a chain, and after the leading `cd` is stripped what remains still holds `&&`, so the whole line is dropped. The fixture does contain the leaf's real proof, fourteen times, as the last segment of a chain: steps 28, 30, 31, 32 end in `go test ./internal/tui3 -run 'TestWorkTab' -count=1`, and step 36 runs `go build ./...` and `go vet ./internal/tui3`. Workers never pass `--check` (the c249 finding this pull request exists for), so on real runs the contract is empty almost always.

**2. An empty contract makes the check a paid no-op, and removes the one power it had.** With no `Checks`, `ranDeclaredCheck` is false, so the store refuses `holds:` and `does not hold:` both; the only ending is `check: <sentence>`. `recordCheckFinding` returns on anything that is not `does not hold:`, so a `check:` answer writes no note, opens no fix task and reads downstream exactly as `holds:` did. A checker that finds a real defect by reading the diff can no longer say so. That is a model call at the check seat's price, per leaf, that can conclude nothing.

**3. The filter still admits what blocker 1 named, and one match still satisfies the gate.** `session.InvocableChecks(tree, …)` on seven probe commands keeps `plandb done t-x --result ok`, `cat go.mod` and `git status --short` beside the three real ones. With `ranDeclaredCheck` returning true on the first match, a check that runs `git status --short` may answer `holds:`.

What would clear it:

- Split a recorded chain on `&&` and `;`, drop a trailing `2>&1`, `| tail`, `| head`, and judge each segment. Keep only commands whose exit status bears on the work: test, build, vet and lint runners and repository scripts. `plandb …`, `cat`, `ls`, `grep`, `git status|log|diff|show`, `sed`, `printf` are never checks. Turn the fixture test round: it must yield `go test ./internal/tui3 -run 'TestWorkTab' -count=1`, `go build ./...`, `go vet ./internal/tui3`, deduplicated, and the probe list above must yield exactly its three real ones.
- `holds:` requires every declared check to appear in the trajectory, not any one. `does not hold:` requires none: a finding costs one fix task, once, and the page already asks for the command that showed it.
- `check:` must mean something. `recordCheckFinding` writes it as a note on the checked leaf, so the root's wake and the task page read that this leaf was not checked, in the surface's existing words. With an empty contract the check may answer `does not hold:` from reading, or `check:`; never `holds:`.
- The before and after cost line I asked for last round is still missing: one small real cell on this branch, the check tasks' count, their spend, and how each ended.

`check:` as the answer verb is fine with me.

santoshkumarradha:
Superseded by #1220, which lands the review check-contract work as one small PR onto santos/dev (no long-lived wave branch). #1220 replaces the trajectory-guessing backfill with the worker-declared-checks shape. Closing this in favor of #1220.

PR #1207, opened 2026-09-18
a belt check runs on the crew checking seat, not the thinking seat

santoshkumarradha:
## What changed

Under `CODEAF_TASK_BELT=bash`, a review check resolved to the mastermind (thinking) tier, so the crew's checker model was billed for but never asked to check. `SeatFor` now maps `RoleCheck` to the careful work tier (`config.ModelTierHigh`), where the crew seats the checker, so a belt check runs on the checker model. The work, fix, plan, and probe seats are unchanged, and the check seat stays read-only (`verifyOnlyBash` / `refuseOutsideDoor` untouched).

## Files

- `internal/run/crew.go`: `SeatFor` returns `ModelTierHigh` for `RoleCheck`; the factory comments updated to match (a check rides the careful work seat, read from the profile).
- `internal/run/crew_test.go`: new `TestSeatForMapsBeltRolesToCrewSeats` pins Work to Worker, Check to High, fix to Worker, Plan to Mastermind; the three factory tests now expect the careful work seat for a check.

## Verification (Spark, tree 090a12c8)

- `go build ./...` ok
- `go test ./internal/run/... ./internal/config/` ok (run 11.4s, config 3.1s)
- `go run ./cmd/codeaf-changes check` ok (8 entries well formed)
- `go test ./internal/guard ./internal/namelaw` ok


santoshkumarradha:
Extension folded in (c264): the check seat now has a door of its own, so a two-model run never draws a third untyped model.

- **`config.CheckSeat(flagCheck, flagPlan)`** resolves the check seat at the door: `--check-model` if typed, else the `--plan-model` flag the person typed (a pinned run rides its two models, the check on the plan seat), else empty.
- **`run.Seats` gains `Check`** and `CrewFactory` seats a `ModelTierHigh` check on `seats.Check`; an empty Check falls to the profile's careful row, the crew's checker, which is what the manual promises the chat door and any unpinned run.
- **`cmd/codeaf/do.go`** adds `--check-model` and wires `seats.Check`; `main.go` documents the ladder (`flag › --plan-model › crew careful`).

So `codeaf do --model A --plan-model B` now runs the check on B, not a third model; `--model A` alone and the chat door get the crew's checker. `--check-model` reads the flag, not the resolved plan seat, on purpose, so the second rung is only "the person typed it".

Verified on Spark, tree `ab4686f8` (then folded to `1b9d1b71e`): `go build ./...`, `go test ./internal/config/ ./internal/run/... ./internal/manual/...`, `codeaf-changes check`, `guard`+`namelaw` all green; the new door tests pass.

Note on the cmd/codeaf package: the full `./cmd/codeaf/...` run shows one red from the `TestDoOnTheRunEngine*` family, but that is a pre-existing in-package flake, not from this PR. Evidence: on santos/dev `bd54a50a3` the same family fails in-package 2 of 2 runs at low load (a different member each time), and every one of these tests passes when run alone. c264 delivered tip-pass at $1.08 on a cheap crew.

santoshkumarradha:
Looked at 1b9d1b71e. The shape is the one I recommended and I would take it once the owner says yes to moving the check off the plan tier. One gap, two notes.

**Gap: a plan seat pinned by the environment is not "the person named it".** `resolveSeat` honours the seat's environment variable as `SeatEnv`, one rung under the flag. `CheckSeat(flagCheck, flagPlan)` reads the two flags only, so a person who pinned the plan seat through the environment gets their checks on the profile's careful row, a third model they never named. That is the case the ladder was written to prevent. Suggest `CheckSeat(flagCheck string, plan Seat)`: the flag first, then `plan.Model` when `plan.Source` is `SeatFlag` or `SeatEnv`, else empty; and give `SeatCheck` its own environment spelling through `seat.Env()` so the three seats pin the same two ways. A test per rung.

**Note: the cell driver's checker cap goes live.** `task-run.sh` stops a run when the checker model's spend passes `CHECKER_CAP_USD` (2). Today the checker model is never called, so it has never fired. With this change the review round bills that model, and every leaf gets a check, every finding a fix and every fix a check, so a cell can now end on `checker-cap` with its work half landed. Worth one line in the change entry's invalidates, since it changes what a cell's `stop` word can be.

**Note: the `TestDoOnTheRunEngine*` flake on trunk.** That is the envelope race cell c256 fixed (the root's worker return landing after the supervisor stops absorbing, so `rootResult` is empty and the door prints the landing line). The fix is on `harness/work-tree` (#1204), not yet on `santos/dev`. It should stop once #1204 lands; if a member still fails after that, it is a new bug.

santoshkumarradha:
Gap closed (c267): a plan seat pinned by ENVIRONMENT now seats the check too, and the check seat has an environment rung of its own.

- `config.CheckSeat(flagCheck string, plan Seat)` now takes the resolved plan seat and rides it when `plan.Source` is `SeatFlag` **or** `SeatEnv`; before it read the plan flag string only, so a plan pinned by `CODEAF_PLAN_MODEL` sent checks to the profile's careful row, a third model on a pinned run.
- New `CODEAF_CHECK_MODEL` env rung (`CheckModelEnv`, registered read-only in `OperatorEnvPins`). Ladder: `--check-model` flag, else `CODEAF_CHECK_MODEL`, else the plan seat when it is flag or env pinned, else the profile's careful row.
- `do.go` passes the resolved `seats.Plan`. Per-rung tests in `internal/config/seats_test.go` and `cmd/codeaf/seats_wiring_test.go`.
- The change note now states `CHECKER_CAP_USD=2` is live, so a cell can end on the checker cap.

Verified on Spark, tree `c0db7f3b`: build, `internal/config` + `internal/run` + `internal/manual`, the `internal/session` prefix-budget + frontier tests (manual edits stay under budget), `codeaf-changes check` (8 entries), `guard`+`namelaw` all green.

The one red in `./cmd/codeaf/...` is the pre-existing `TestDoOnTheRunEngine*` envelope-race flake (c256's fix, on harness/work-tree #1204, not on santos/dev yet). It fails ~1 in 12 runs ALONE on this branch AND on the base identically, so it is not from this PR and will clear when #1204 merges down. c267 delivered complete; it capped at $5.01 because the checker now genuinely runs its round, which is the behavior this PR restores.

santoshkumarradha:
Final look at the head after c267: the gap is closed. `CheckSeat(flagCheck, plan Seat)` takes the flag, then `CODEAF_CHECK_MODEL`, then the resolved plan seat when the person pinned it by flag or by environment, and only then leaves the seat empty for the profile's careful row. A run with `--model A --plan-model B`, or with the plan pinned by environment, stays a two-model run; an unpinned run and the chat door get the crew's checker. The live checker cap is stated in the entry. Nothing further from me. It still moves the check off the plan tier, a choice I made in a cell brief on 2026-09-18 and the owner never ruled on, so it waits on his yes.

santoshkumarradha:
Correction to my earlier comment on this PR: I wrote that c267 "capped because the checker now genuinely runs." That is wrong, and I want the record straight.

The dealer pulled c267's usage.jsonl per model: openai/gpt-5.6-sol 466 calls / $5.014, qwen/qwen3.8-max 0 calls / $0. The checker never ran. c267's review round still executed on the thinker/plan seat and hit the cap there, the same plan-seat billing artifact this PR fixes, not a real checker-cap.

Why: the cell ran on the belt binary (66b5a65f, pre-#1207). A cell's own checks are seated by the RUNNING binary, not by the code in its worktree, so #1207's SeatFor/CheckSeat fix in the tree does not change how that cell's checks are billed. A belt cell will only route its checks to the checker once the belt binary is itself rebuilt from a tree that contains #1207 (i.e. after this merges, or an explicit belt build off #1207's head).

What is unchanged: the fix code in this PR is correct and complete, verified green in the tree (build, config, run, manual, the session prefix-budget tests, changes, guard, namelaw). Only my characterization of why c267 capped was wrong.

santoshkumarradha:
Landed on santos/dev as ae0eef3b6 (squash of this branch at 9536c0659, tree 5d33e2b10). Full make check on that exact commit on Spark: EXIT 0, 114 ok (spark:~/src/trees/check-sdev-ae0eef3b6.log). The first run's one red was internal/run TestStartAnswersTheRootsDoneResult, a load-dependent race that also lives on the base (80 of 80 green on both trees at normal load); kept as check-sdev-ae0eef3b6.run1-flake.log.

PR #1208, opened 2026-09-18
installer: the PATH line is the last thing printed, bare and bold green

ZeroPoint95:
Install prints three lines and ends on the `export PATH` line. `codeaf telemetry` gains `info`; `show` becomes JSON. Four groups of change, 14 commits.

## 1. The install one-liner prints three things and nothing else

Real capture, Debian arm64 container:

```text
$ curl -fsSL https://agentfield.ai/get/codeaf | bash
installed codeaf v0.3.0 built 2026-09-18 18:16 · go1.26.5 linux/arm64

codeaf shares anonymous performance data with AgentField
codeaf does NOT share your prompts, code, files, or any private information
see what is shared: codeaf telemetry info · turn off: CODEAF_TELEMETRY=off

export PATH="/root/.codeaf/bin:$PATH"

```

- Last line is **bold green** on a terminal, plain when piped or with `NO_COLOR`.
- Gone from a normal run: `codeaf: stable v0.3.0 for …`, `codeaf: installed <path>`, and the `codeaf: add it to this shell with:` prefix. `--verbose` still prints the first two, on stderr.
- Unchanged: when the line prints (folder not on `PATH`, never on Windows), `--no-modify-path`, the shell-file edit, checksums, channels.

## 2. `codeaf telemetry` has five verbs, was four

- **`info`** (new) opens on `codeaf does NOT collect or share your chat`, then `1. Usage Counts` and `2. Model Pool`: this machine's live values, one example row per event from the contract's own band constants, and an indented example pool row. No `never` lines inside, no band lists.
- **`show`** is JSON only, indented by two: `{"usage": {destination, off?, waiting[]}, "model_pool": {…}}`. `off` appears only when nothing is sent there. Empty queue is `[]`, never `null`.
- The binary's notice, README and `docs/TELEMETRY.md` now say `What is collected:        codeaf telemetry info`. Tests hold all four copies to the same bytes.

<details>
<summary><code>codeaf telemetry info</code> on a fresh stable install</summary>

```text
codeaf does NOT collect or share your chat. No prompts, replies, code, file names,
paths, repo names, keys, email, IP or machine name leave for AgentField. Only the
fields below do, as this machine would fill them.

1. Usage Counts (https://agentfield.ai/api/oss/codeaf/telemetry)
  every event, as this machine would send it now
    codeaf_version       v0.3.0
    channel              stable
    os                   darwin
    arch                 arm64
    usage_context        local
    install_method       script
    install_id_hash      sha256 of a random id, minted on the first send
    session_id_hash      sha256 of the run id, one per session; absent on first_run
    event_id             16 random bytes as hex, one per event
    event_time           2026-09-19T18:11:40Z

  what each event adds, for example
    first_run            nothing; sent once per install
    session_started      mode=chat  resumed=false
    session_ended        mode=chat  duration=5-30m  turns=6-20  model_calls=21-100  model_calls_failed=0
                         tool_calls=6-20  tool_calls_failed=0  cost_usd=0.1-1  stop_reason=done  exit_code=0
    fault                mode=chat  scope=main  fingerprint=3fa9c1e2b7d04e85
    stop_reason          one of done · error · incomplete · budget · turn-cap · deadline · price · question · interrupted · unknown

2. Model Pool (https://codeaf.agentfield.ai/pool/v1/rows)
  one row per judged seat, after a task lands, for example
    {
      "schema": 1,
      "metric": "role_quality",
      "role": "worker",
      "model": "<the seat's model slug>",
      "score": 81,
      "judge": "<the judge's model slug>",
      "door": "task",
      "size": "M",
      "day": "2026-09-20"
    }
    nonce                16 random bytes as hex, one per row, so a resend is not a double count
    X-Codeaf-Install     a header: a random per-install id, minted on the first send; not the usage counts' id
```

</details>

<details>
<summary><code>codeaf telemetry show</code> with one event waiting</summary>

```json
{
  "usage": {
    "destination": "https://agentfield.ai/api/oss/codeaf/telemetry",
    "waiting": [
      {
        "event_name": "session_started",
        "event_id": "8c1f4bd3bf17bad6e701e0ece115288f",
        "install_id_hash": "cd95afc6…",
        "session_id_hash": "d6f189b2…",
        "event_time": "2026-09-19T18:11:32Z",
        "props": {
          "arch": "arm64",
          "channel": "stable",
          "codeaf_version": "v0.3.0",
          "install_method": "script",
          "mode": "chat",
          "os": "darwin",
          "resumed": false,
          "usage_context": "local"
        }
      }
    ]
  },
  "model_pool": {
    "destination": "https://codeaf.agentfield.ai/pool/v1/rows",
    "off": "model_pool read (capped by the telemetry off switch)",
    "waiting": []
  }
}
```

</details>

## 3. Merged from `claude/codeaf-telemetry-disclaimer-5af1dd`

- `codeaf telemetry show` covers the Model Pool stream, and every telemetry-off switch caps the pool at `read`.
- One follow-up fix here: the config reader law names `ModelPoolAt` for the `model_pool` row, the reader that branch left it with.

## 4. After review

- `internal/release`'s five installer tests asked a normal run for the channel, tag and platform. They ask a `--verbose` run now, where those lines live, and read the receipt on a normal run. The channel-picking tests still see which release was chosen.
- A missing release names its tag in the failure line itself: `no codeaf-linux-arm64 in release v9.9.9; check the tag on the Releases page`.
- An empty `CODEAF_TELEMETRY_ENDPOINT` caps the Model Pool at `read`, as `docs/TELEMETRY.md` already claimed. `poolcfg` reads it as the third environment rung, with tests.
- The comment over `telemetryRowsOff` says why the former `AFORGE_TELEMETRY` spelling reaches the pool (`ProjectBoolAt` falls through to `TelemetryAt`, which reads the pin through `internal/env`), so a tidy-up cannot drop it.

- Merged `dev` after #1333 (devaf). The installer keeps its `--name` / `CODEAF_INSTALL_NAME` and the quiet run: the receipt names the installed file's `version`, the notice follows, the `export PATH` line is last. #1333's new `TestV1InstallerName` reads the receipt and the last line on a normal run and the install path on a `--verbose` run. The guide and manual say the same.

## 5. Proof

- `test/installer-telemetry.sh`: 46 checks. Notice bytes against `docs/TELEMETRY.md`, the hint's blank lines, no escapes off a terminal, `NO_COLOR`, the hint call is the script's last line.
- New Go tests: example values held to the allowlist, the pool example parsed back into `record.Row`, `show` re-encodes to the same bytes, `info` opens on the chat sentence and has no `never` line.
- Ran the installer for real against v0.3.0 in a Debian arm64 container under a pty; section 1 is that capture.
- `go test ./internal/release/` passes the five installer tests. `make pr-ready`: green except three macOS-only reds that fail identically on clean `origin/dev` and pass on Linux, where `dev`'s Full check is green: `TestC9SourceBuildRefusesBeforeTheNetwork` and `TestC9AnInstallEndsWithTheNewBinarysVersionLine` (`/var/folders` vs `/private/var/folders`), and `TestTheReleaseSurfaceTestSurvivesEveryLedgerState` (bash 3.2 and an empty array under `set -u`). Not from this branch.

## Try it

Installs into a throwaway folder; touches nothing under `~/.codeaf` or your shell files.

```sh
T=$(mktemp -d) && CODEAF_HOME=$T/home CODEAF_INSTALL_DIR=$T/bin bash scripts/install.sh --no-modify-path
```

Note: `agentfield.ai/get/codeaf` serves `scripts/install.sh` from `main`, so the public one-liner changes on the next promotion, not on merge to `dev`.


AbirAbbas:
Reviewed this on Linux: a real install against the published v0.3.0 (pty, piped, `NO_COLOR`, `--verbose`), the built binary's `telemetry info` / `show` on a throwaway home, and a read of the diff. The installer output, the colour gating and the notice bytes all match what the description says, `show` is valid JSON with `[]` never `null`, and `info` prints nothing private. The pool cap holds for the env switches, the disk rows and the legacy spelling.

One thing blocks merge:

**`go test ./internal/release/` fails five tests on this head and passes on dev.** `TestInstallerGetsLatestStableAndFinishesWithVersion`, `TestInstallerPicksTheNewestChannelBuildWhateverTheListOrder`, `TestInstallerPinsAReleaseAndNamesAMissingOne`, `TestDocumentedVersionPinReachesThePipedInstaller` and `TestInstallerUsesWgetWhenCurlIsAbsent`. They assert a normal run names the channel, tag and platform (`internal/release/install_test.go:266`, `:576`, `:622`) and that a missing release names `v9.9.9` (`:590`, `:826`), and all of that moved behind `--verbose` in `scripts/install.sh`. CI did not catch it because the touched-packages gate derives its package set from changed Go files, so a scripts-only change runs no suite at all. That is a CI gap we will fix separately, but this PR needs the tests and the script to agree before it lands. Either direction is fine by me, restore the strings or move those assertions to a `--verbose` run, as long as the channel-picking test can still tell which release was chosen.

Two smaller things, not blocking:

- `docs/TELEMETRY.md:99` says an empty `CODEAF_TELEMETRY_ENDPOINT` is one of the switches that also stops the Model Pool from sending, and `codeaf --help` says the same. It does not: `internal/pool/poolcfg/poolcfg.go` reads only `CODEAF_TELEMETRY` and `DO_NOT_TRACK`, and `telemetryRowsOff` in `internal/config/settings.go:3411` reads only the two disk rows. With an empty endpoint the usage counts stop and the pool keeps sending. Either cap it or drop it from the list.
- The comment above `telemetryRowsOff` (`internal/config/settings.go:3396`) says "disk only, never the process environment", but the reader it calls goes through `TelemetryAt`, which does read the environment, and that is exactly why the legacy `AFORGE_TELEMETRY=off` caps the pool. Worth making the comment say so before someone "fixes" it and drops the legacy spelling.

Noting for promotion rather than for this PR: a fresh stable install today rejects `codeaf telemetry info`, and the public one-liner serves the script from `main`, so the notice will point at a verb the installed binary does not have until the next stable carries it.

AbirAbbas (approved):
Re-verified 2026-09-21 after the review items were addressed: each item checked in the code and on the built binary, gates green locally and in CI at this head. Details in the review ledger.

PR #1210, opened 2026-09-18
run: a run's result is read from the store when the root's worker returns after the run ended

santoshkumarradha:
## What

`run.Start` builds its summary's result from the root worker's return alone. The root's worker ends its own task with `plandb done`, so the store reads done before the worker has come home. A pass on the supervisor's 300 millisecond clock that lands in that gap ends the run, and the summary's result is empty while the store holds it.

The fix is six lines in `internal/run/run.go`: when the absorbed result is empty and the store's root reads done, the summary carries the stored result. The same commit is already on `harness/work-tree` (#1204); it is sent here on its own because the race is on the branch everything else builds from and has failed two full gates.

## Evidence (Spark, base bd54a50a3)

| | with the fix | without the fix |
| --- | --- | --- |
| new `TestStartAnswersTheRootsDoneResultWhenItsWorkerReturnsLate`, `-count=3` | 3 of 3 pass | 3 of 3 fail: `summary result = ""` |
| existing `TestStartAnswersTheRootsDoneResult`, `-count=150 -cpu 1,4`, load about 5 | 300 of 300 pass | 300 of 300 pass |
| existing test inside a full `make check` at load 12 | not yet run | failed twice (`wait_test.go:208`) |

The existing test only loses the race under load, which is why the new test holds the worker's return back for longer than a pass: the gap is every run.

`go test -count=1 ./internal/run`: ok. `gofmt -l ./cmd ./internal`: clean. `make changelog-check`: well formed.


santoshkumarradha:
Landed on santos/dev as the squash 43e42cdfc, carried by the integration commit 164548f9e. Evidence: full make check on Spark on the exact commit 164548f9e (tree 9cb4ae561), EXIT 0, 114 packages ok, log ~/src/trees/check-sdev-164548f9e.log. A first run failed on an unrelated temp-dir cleanup flake in cmd/codeaf (TestResumeOpensTheConversationThePersonSpokeInLast, passes 20 of 20 alone); that log is kept as check-sdev-164548f9e.run1-flake.log and the flake class has its own diagnosis cell.

PR #1211, opened 2026-09-18
test: join the synthetic usage writer before teardown in the waiting-line test

santoshkumarradha:
Null result on the once-in-six "TempDir RemoveAll cleanup: directory not empty" failure of TestTheWaitingLineSurvivesAJournalThatIsMovingElsewhere.

The late writer was the test's own synthetic usage-noise goroutine, not product code. The test spawned a goroutine calling graph.RecordUsage in a loop and cancelled it with context.WithCancel plus defer stop(), but stop() only signalled cancellation and never waited, so a RecordUsage could still write under t.TempDir() after the test body returned, which is what the cleanup RemoveAll then hit.

The fix is test only: startUsageNoise returns a stop that cancels the context AND joins the goroutine before returning, so teardown cannot run while a write is in flight. TestUsageNoiseStopWaitsForTheWriter forces the ordering so the guarantee is deterministic rather than one-in-six.

Product code is untouched: the real codeaf do journal and usage writers are not implicated; this was a test teardown race no user sees.

Verified on Spark at the rebased head: go build ./... clean, gofmt -l clean, go test ./cmd/codeaf ok (89.7s), the two tests pass 30 of 30 under -race, cmd/codeaf-changes check clean, guard and namelaw ok.


santoshkumarradha:
Landed on santos/dev as the squash 7cda67c9a (tree 53b7129aa). Evidence: full make check on Spark on that exact commit, EXIT 0, 114 packages ok, log ~/src/trees/check-sdev-7cda67c9a.log.

PR #1212, opened 2026-09-18
session: a landing's report is evidence to the completion check, and a check refused for its cd says the form that passes

santoshkumarradha:
## What

Two defects seen on the real binary with the belt on, in a throwaway home and a two-file repository.

- **The completion check did not count a landing's report.** A hand-off landed, the conversation answered the question it owed, and the check then raised: "The delegated task's report never came back in evidence". The turn carried on once more, the model answered `[no change]`, and the person saw `the ask is not finished` under a finished answer. The landing's document is the person's question, a blank line and the run's outcome line; the check read all of it as the ask. Now the ask is the question alone, and the outcome line reaches the check's reader as something that came back. A turn no landing woke reads exactly as before.
- **A check that leads with a directory change says what would pass.** It is still refused: dropping the step is meaning-preserving only when the directory is the ground itself, which this door does not know, and a check that changes into any other folder would be kept as a command run where its files are not. The refusal now ends with the form that passes: leave the directory change out and name each file by its path. An absolute path as an argument is legal. The schema sentence says the shape and names no tool.

## Not here

The wider contract problems the owner hit in a live conversation (one command with quoted metacharacters refused, problems arriving one per round, the ground asked late, parallel siblings counted as repeats) are their own cells and follow as small pull requests.

## Proof (Spark, on 164548f9e)

`gofmt -l ./cmd ./internal` clean, `go build ./...`, `go vet ./internal/session`, `go test -count=1 ./internal/session -run 'Checks|Refusal|Owed|Landing|Checkpoint|Prefix|Frontier|Propose|Completion|Schema|Belt'` ok with `CODEAF_TASK_BELT` unset, `go test ./internal/manual/ ./internal/namelaw/` ok, `make changelog-check` well formed. The fixed prefix stays under its budget.


santoshkumarradha:
Landed on santos/dev as the squash afdf015e2, carried by b94df33da. Evidence: full make check on Spark on the exact integration commit b94df33da (tree a57201094), EXIT 0, 114 packages ok, no failures, log ~/src/trees/check-sdev-b94df33da.log. Earlier runs of this work failed on single tests outside the change that pass alone; those logs are kept beside it as run1 to run4 flake logs and have their own diagnosis cells.

PR #1213, opened 2026-09-18
session: a reopened conversation keeps its tasks: a plan-born node reads back, and a refused checkpoint is never overwritten

santoshkumarradha:
## What the owner saw

"i seem to lose done tasks from chat, i close and open again, not sure why". A conversation that had run twenty tasks reopened with an empty rail.

## Cause, from his own files

The engine's log at the reopen: `session: ignoring corrupt task checkpoint …/tasks.json: node 5 has no acceptance`. `tasks.json` had been 42 KB; the next save made it 1.8 KB with `"nodes": null`, and the id counter restarted, so new rows answered to numbers the transcript already used. Transcripts, working copies and the plan store were all still on disk.

Three defects made that one event:

1. **The reader refused what the writer writes.** A task out of a run's plan is admitted to the tree with the store's id and no acceptance of its own (`plandb_plan.go`, the dispatch), and `decodeTasks` refuses any node without one, whole file. A plan-born node now reads back; an ordinary node without an acceptance is refused as before.
2. **A refused checkpoint was overwritten.** The graph opened empty and its first save landed on the same path. The file is now moved beside itself as `tasks.json.refused-<seconds>` before anything is saved.
3. **Ids were reused.** The counter lived only in that file. It is now raised past every task that left a transcript or a working copy in the conversation's folder.

And one more found reproducing on 7cda67c9a (throwaway home, real binary, a run with two parts): **a finished belt run reopened as running**, with a spinner, because its ending reached the screen and never the checkpoint. The ending is saved with the row, and a run's row keeps when it started and ended.

## Not here

His lost sheet cannot be restored from `tasks.json` (it was overwritten before this fix existed). Rebuilding a conversation's sheet from the transcripts, copies and plan store that are still on disk is its own cell, together with the chat's `tasks` tool answering from the run's store.

## Proof (Spark, on 7cda67c9a)

`gofmt -l ./cmd ./internal` clean, `go build ./...`, `go vet ./internal/session`; `go test -count=1 ./internal/session -run 'StartTaskBashBelt|Belt|Landing|Owed|Checkpoint|RefusedCheckpoint|PlanBorn|RunRow|Job|TaskStore|Recover'` ok with `CODEAF_TASK_BELT` unset; `go test ./internal/manual/` ok; `make changelog-check` well formed. New tests: a plan-born node reads back, a refused checkpoint is kept byte for byte beside itself and the next id is past the highest journal on disk, a reopened conversation's row for a landed run is done with its start and end.


santoshkumarradha:
Landed on santos/dev as the squash 52a73b677 (head 9bad3579b, with the decodeTasks ratchet fix), carried by b94df33da. Evidence: full make check on Spark on the exact integration commit b94df33da (tree a57201094), EXIT 0, 114 packages ok, no failures, log ~/src/trees/check-sdev-b94df33da.log. Earlier runs of this work failed on single tests outside the change that pass alone; those logs are kept beside it as run1 to run4 flake logs and have their own diagnosis cells.

PR #1214, opened 2026-09-18
session: a proposal learns every problem in one refusal, and a task stands where its brief plainly says

santoshkumarradha:
## What the owner saw

Three tasks proposed in one message; nine calls refused over three rounds before three started. Rounds 1 and 2 were about the shape of a check (its own pull request follows). Round 3 refused all three for a rule nobody had mentioned: "this conversation has been working in two places … Ask the person which", although every brief named the repository it was about. Round 4 set `ground` to that third place and everything started.

## In this branch

- **Every problem in one refusal.** The contract's optional halves (what it expects, its checks) are all read before anything is returned, and when the rest of the contract is whole the stand is resolved too, so a refusal about a check also says the question about where the task stands. Nothing refused before starts passing.
- **The brief gets a rung of its own**, between `ground` and the conversation's places. By property, naming no tool or kind of file: the existing absolute folders a task's own contract names, exactly one folder containing them all, and that is where it stands; the receipt says `brief` decided. Several places inside one folder are one answer; an aside naming an unrelated repository makes the brief ambiguous and the ladder climbs on as before.
- **A folder outside every repository is where output goes.** It does not vote when another named folder is inside a repository, and a stand read from the brief is not then refused for the output folder the same brief names. Without this the rung answered his first round (the scratch folder did not exist yet) and fell silent for the second (it did).
- **A said third place** is accepted as before and is visible at once: the receipt shows the folder with `said` provenance. A test pins it.

## Proof

Laptop, focused only, because Spark is held quiet for the trunk's gate: `gofmt -l` clean, `make test-focus PKGS=./internal/session RUN='Ground|Stand|Propose|Refus|Places|Prefix|Frontier|…'` ok with `CODEAF_TASK_BELT` unset (38 s), `go test ./internal/manual/ ./internal/namelaw/` ok, `make changelog-check` well formed. The full check is yours on Spark.


santoshkumarradha:
Landed on santos/dev as the squash b94df33da (head 4bc193400). Evidence: full make check on Spark on the exact integration commit b94df33da (tree a57201094), EXIT 0, 114 packages ok, no failures, log ~/src/trees/check-sdev-b94df33da.log. Earlier runs of this work failed on single tests outside the change that pass alone; those logs are kept beside it as run1 to run4 flake logs and have their own diagnosis cells.

PR #1215, opened 2026-09-18
session: a check is one command as the shell reads its quotes, its refusal says what passes, and side-by-side calls are one attempt

santoshkumarradha:
Stacked on #1212, #1213 and #1214 (the branch is cut from the stacked candidate b94df33da); its own diff is the last six commits.

## What was true

The owner proposed three tasks in one message on 2026-09-18 and nine calls were refused over three rounds. Every refusal was about the shape of a check.

- A check holding a bar inside a single-quoted search pattern was refused as a pipe. The shape test read the raw text and never the quotes.
- The refusal ended `… is not`. It named neither the character nor a form that would pass, so the next call repeated the shape.
- After the third round a `[stuck]` note said the call had been repeated three times and failed the same way. Those three were side-by-side calls of one reply, and the failure had changed every round.
- The shape test also rebuilt the command from its fields, which flattens spacing inside a quoted argument.

## What is true now

- A check is one command as the shell that runs it would read it. Inside single quotes every character is text. Inside double quotes a dollar, a backtick and a backslash still refuse the check, because the shell still expands them there. An unclosed quote is refused. Nothing the shape law stopped before starts running: only what the shell passes to the one program as an argument is newly allowed, and it reaches the checker byte for byte.
- The refusal names the one character and says what passes. A check that leads with a directory change keeps the sentence #1212 gave it. It never offers a rewritten command.
- The schema sentence says both rules inside the fixed prefix budget.
- Calls made side by side in one reply count as one attempt, and one failure shared by them counts once.

## Proof

Focused, on the laptop, with the belt variable unset: the check, refusal, audit, loop, prefix, frontier, proposal and complexity ratchet tests pass in 23 s. `gofmt -l ./cmd ./internal`, `go build ./...`, the manual and name laws are clean. The full suite runs on Spark at the gate.


santoshkumarradha:
Landed on santos/dev as the squash d5f2700c4 (head 3446a861a), carried by 39ffbc548. Evidence: full make check on Spark on the exact integration commit 39ffbc548 (tree bb5d10af5), EXIT 0, 114 packages ok, no failures, log ~/src/trees/check-sdev-39ffbc548.log. Two earlier runs failed for reasons outside the change, with logs kept: one under heavy machine load, one because another process on the machine held port 8765.

PR #1220, opened 2026-09-19
run: declared checks are the review contract, judged by observable properties, with the verdict basis recorded

santoshkumarradha:
The review round's contract becomes the command or commands the worker declares as the proof of its work, judged only by properties any project shares: the checker ran the command in this check task, its exit was recorded, and it passed the shipped read-only audit law. A holds conclusion requires every declared command to have a recorded zero-exit audited run; a does not hold conclusion is never gated, so a reading checker can still report a defect on an empty or partial contract; a leaf that declared nothing is judged by reading, and reading can hold. Each verdict records its basis on the leaf, by run with the commands and their exit codes, or by reading, so a later reader and a grid see how a verdict was earned. The trajectory-guessing backfill and its runner set are removed, and the check seat stays read-only.

This closes #1206. Its check-contract line of work lands here as one small PR onto santos/dev rather than a long-lived wave branch.

Provenance, plainly: the product commit on this branch was written with a second agent present in the worktree during the run. It was accepted on review and on tests, not on the cell's own say-so. The seven test-first commits specify the behavior; the change passes them, the internal/plandb, internal/run and whole internal/session suites, the complexity ratchet, and the prompt prefix budgets; and its store.go was read line by line for anything beyond scope. The learning crew's fit row for the originating run is flagged two-agent and excluded from crew fit and the arbiter.

Verified on Spark at tree b61129dde: go build clean, gofmt -l clean, go test ./internal/plandb/... ./internal/run/... ok, the whole internal/session suite ok through one-suite.sh (259s, including the ratchet and prefix-budget tests), go run ./cmd/codeaf-changes check clean, and go test ./internal/guard ./internal/namelaw ok.

## Update: one quote-aware reader, shared with the store (folds in #1215)

santos/dev 39ffbc548 (with #1215) is merged in. The check door's quote-aware reader now lives in one place, internal/approval: FirstCompositionOutsideQuotes and FirstBarOutsideQuotes, with the composition character set as approval.ShellComposition. Session's three files (task_audit.go, task_checks.go, task_divide_scope.go) and the store's auditableDeclaredCheck all read through it. The store's duplicate declaredCheckComposition const and its non-quote-aware strings.Fields composition pass are gone, so a check the proposal door admitted with a quoted bar can no longer be refused a holds verdict by the store.

A store-side acceptance test (internal/plandb, through checkVerdictBasis) reuses the proposal door's own eight-form never-run table: a quoted-bar check is admitted, run, its exit 0 recorded, and holds is allowed; every form that is more than one command is refused by the store even with a zero exit on file.

Verified on Spark at tree a8a1ddf12: go build clean; whole internal/session suite ok through one-suite.sh (217s), including the complexity ratchet, both prefix-budget tests, and the door's quoted-bar contract test; internal/run, internal/plandb, internal/approval ok; gofmt -l ./cmd ./internal clean; go run ./cmd/codeaf-changes check clean; go test ./internal/guard ./internal/namelaw ok. Combined exit 0.

## Recording the command's real exit (round 2)

The verdict now rests on the command's own exit, recorded by the belt as it runs. A bash tool that ended is exit zero; one that failed carries the non-zero code the belt appended to its output (read back, or a plain non-zero when the output was cut); a harness refusal and a non-bash tool record no exit. The exit is a pointer on the step, so a step written before this field, or one whose exit is unknown, decodes to nil and is never read as a zero a real exit could equal. A holds verdict rests only on a recorded zero exit for every declared command.

A build that records exits stamps an opening line, before any step, and the ending line with exits_recorded, so a trajectory cut off mid-run is still known to be from a new build. On a new-build record a declared check with no recorded run refuses the holds verdict: a checker that ran nothing has not earned it. A record from before exits were recorded cannot be judged that way, so it holds as a reading and names, in VerdictBasis.Unobserved, the declared checks it never observed, so no reader mistakes it for a holds earned by running them.

Acceptance drives a real command through the real belt recorder into the store, with no hand-written exit line: a passing command yields holds, a failing one is refused.


santoshkumarradha:
Held at 353f434e2. The shared reader move is right. Under it, the verdict gate reads an exit code from each recorded step, and nothing in the product writes that field, so on the real binary a declared check that ran and failed would read as exit 0 and earn holds. Found in review of this head; the tests were green because only a test helper wrote the field. The fix records the command's own status, reads an absent status as unknown, and lets unknown earn nothing, accepted through the real worker and recorder with a passing and a failing command. Gated again on the new head.

santoshkumarradha:
Landed in santos/dev as squash 2daa32ef2. Gated as a five-PR stack (#1220, #1237, #1232, #1235, #1240) on the stack top 608c459f2, tree e0413ce99: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Run 1 of the same commit failed once in cmd/codeaf, a package this stack does not touch (TestALaunchReusesOneEmptySessionAndReapsTheRest, TempDir cleanup: directory not empty); the log is kept and that class has its own diagnosis cell. Logs: spark:~/src/trees/check-sdev-608c459f2.log and .run1-flake.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1221, opened 2026-09-19
test(lane): gate router movement on request counts so the verdict does not depend on host load

santoshkumarradha:
## What changed

The S2 router-movement gate (TestS2TheDefaultGoesSlowAndTheRouterMoves) now scores each request on the SCRIPTED first token of the lane that answered, looked up from the sheet, and reinstates the design's own ship gate on that figure. The verdict no longer reads measured wall time, and it is no longer the weaker request-count check.

## Why

Two problems, in order:

1. The original gate read p90 first token off measured wall time. On this hundredfold clock a healthy answer carries about 2.5 s of loopback overhead around 1.2 s of script, and that overhead sits in the numerator AND the denominator of the improvement ratio where it does not cancel. Under load one scheduler delay on the eighteenth of twenty answers dragged a healthy 45% improvement to 28.7% and failed the 30% gate. That is the flake.

2. The interim fix replaced the gate with a request-count check, routedVictim >= pinnedVictim. The pin sends 20 of 20 to the slow default, so a router that moved a single request of twenty passed. That lowered the bar the design set.

## The fix

Each request's first token is scored as the scripted median of the lane that answered it (e2eSheet), with a lane that was rescripted slow scored at its slow figure. This value is fixed by the script, so the same routing decisions give the same verdict on a quiet box and a loaded one. On that figure the test restores both clauses of the design gate:

- a p90 first-token improvement of at least 30 per cent (provider-routing.md, Part III, C3), and
- the moving arm's p90 below the pinned arm's median.

The measured wall figures stay in the test as logged diagnostics.

## Load independence and the bar

Stressed on a box at load about 80, five runs of S2 in a row gave routed scripted p90 of 768 to 844 ms against the pin's 4 s every time, a 79 to 81 per cent improvement, no flake. The earlier measured-wall attempt had failed one of the first three runs on the same box at 29.3%.

## The deliberate negative

A router configured not to move (the routed arm pinned to the victim) scores the pin's own scripted p90 and fails by construction:

    e2e_test.go:449: victim Baidu scripted p90 first token routed 4s, pinned 4s (pinned p50 4s); ...
    e2e_test.go:468: the router's scripted p90 first token is 4s against the pin's 4s, 0.0% better, and the gate is 30%
    --- FAIL: TestS2TheDefaultGoesSlowAndTheRouterMoves

## Verified on Spark

Tree ed368fcc5. go build ./... clean; go test ./internal/lane/... ok; gofmt -l ./cmd ./internal clean; go run ./cmd/codeaf-changes check clean; go test ./internal/guard ./internal/namelaw ok. Combined exit 0. Test-only change, no changelog entry.


santoshkumarradha:
Landed on santos/dev as the squash 39ffbc548 (head b45315976). Evidence: full make check on Spark on the exact integration commit 39ffbc548 (tree bb5d10af5), EXIT 0, 114 packages ok, no failures, log ~/src/trees/check-sdev-39ffbc548.log. Two earlier runs failed for reasons outside the change, with logs kept: one under heavy machine load, one because another process on the machine held port 8765.

PR #1222, opened 2026-09-19
session: the conversation's tasks tool reads a run's tasks from the run's own record

santoshkumarradha:
Cut from the candidate that carries #1215, so until that lands this diff shows #1215 too. Its own change is the last three commits.

## What was true

On the real binary, 2026-09-18, belt on: a hand-off ran, split in two, was checked and landed. One turn later the conversation called `tasks {"id":"1"}` and was told `No task "1" in this project`, then `tasks {}` and was told `No tasks have run in this project yet`, beside a rail that showed `#1 done`. Told no task existed, it set out to check the work by running the suite itself and spent five minutes on an approval that timed out.

## What is true now

- Under the belt, `tasks` lists the run's tasks first: the name a person sees, the title, the state and the first line of what came back. Tasks from earlier sittings and other windows still follow.
- A hand-off is named by the number its card and the rail show, `#2`. A part the run made is named by its place under it, `#2.1`. A store id never reaches an answer.
- Reading one task answers from the run's record: what it was asked, what came back in full, what the run's checks found, and its last steps, bounded like the engine's other reads of a run.
- A name the run does not hold is answered by the shipped reader exactly as before, and with no run nothing changes.
- The tool's description and the conversation's page say a finished task is asked about with `tasks` and never redone by hand, inside the fixed prefix budget.
- The manual section on asking what a task did described an ask box that no screen opens. It now says to ask the conversation, how tasks are named, and that the box does not exist yet.

## Proof

Focused on the laptop with the belt variable unset, every exit code read directly: the tool, plan, prefix, frontier, schema, conversation page and complexity ratchet tests pass. `gofmt -l ./cmd ./internal`, `go build ./...`, the manual and name laws and `make test-laws` are clean. The fixture is the store shape a real run left on 2026-09-19. The full suite runs on Spark at the gate.

Known limit: this works in the engine, so it works in a hosted conversation. The rail's own run rows do not yet reach a hosted conversation; that is a separate change.


santoshkumarradha:
Full check on Spark, squash a26a36b8f (tree 334f93d46, identical to head c609a6151): EXIT 2. Two deterministic failures in internal/session: TestEveryLawIsStatedOnceAndInItsOwnPlace (lawregistry_test.go:334) and TestTheHandoffReceiptSaysTheSessionIsToldWhenTheWorkLands (tasklook_test.go:109). Both require the sentence "Never to WAIT for handed-off work" in the tasks description; the rewrite reads "Never WAIT for handed-off work". Held until the description and the law registry agree; then it is gated again.

santoshkumarradha:
Landed in santos/dev as squash d026bc6a7. Gated as part of a seven-PR stack (#1223, #1222, #1225, #1226, #1228, #1227, #1229) on the stack top fe9eb62af, tree 0a5b96a35: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-fe9eb62af.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1223, opened 2026-09-19
config: report unread profile config keys once at load

santoshkumarradha:
## What was wrong
Profile config.json is read as a FLAT map of dotted keys (readProfileConfig, internal/config/budget.go; per-key lookup persistedValue, settings.go). A key at the wrong depth (a nested {"models":{...}} where "models.tiers.reflex" was meant) or a mistyped dotted key is silently ignored: it is not the dotted key any reader looks up, so every affected seat falls to the shipped defaults with NO error. A fresh profile that set model.talk and all five models.tiers but NESTED sent 16 requests in one chat --once turn, none the configured model (8 google/gemini-2.5-flash reflex, 4 ~deepseek/deepseek-v4-flash-latest talk, 4 deepseek/deepseek-v4-flash-0731 low). A misshapen config was silently the same as no config.

## The change (internal/config/config.go)
At load, warn ONCE per profile (sync.Map guard) naming every top-level config.json key that no reader consumes (the settings registry Rows() plus the known non-setting fields). By property, any key the reader does not know is named, at any depth, including a nested object used where a dotted key was meant. It is a log line, never an error that stops a session, and it does not change how a correctly shaped flat config resolves.

## What a person sees
- 12 nested file {"models":{"tiers":{"reflex":"..."}},"model.talk":"..."}: `codeaf: <profile>/config.json has unread top-level config key(s): models`
- one mistyped dotted key {"model.tark":"..."}: `codeaf: <profile>/config.json has unread top-level config key(s): model.tark`
- a correct flat config (api_key, model.talk, models.tiers.*, effort, model_sources, tools.approval, tools.approvalMode): NO output.

## Scope
Surfaced through the process logger, once per load. Whether it should also appear on the chat surface at turn time is a separate product decision (chat-time attribution), out of scope here.

## Tests and checks (Spark, focused, rebased onto 39ffbc548)
- internal/config/config_test.go TestLoadWarnsOnceForEveryUnreadTopLevelProfileKey: nested + typo named, flat keys NOT named, resolution unchanged, warned once across two loads.
- go build ./..., go vet ./internal/config/..., go test ./internal/config ./internal/guard ./internal/namelaw (ok), go test ./internal/session for both prefix-budget tests and the three ratchet tests (all PASS), gofmt -l ./cmd ./internal empty, no em dashes.


santoshkumarradha:
Landed in santos/dev as squash 927d1e1a3. Gated as part of a seven-PR stack (#1223, #1222, #1225, #1226, #1228, #1227, #1229) on the stack top fe9eb62af, tree 0a5b96a35: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-fe9eb62af.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1224, opened 2026-09-19
design: closing the window never ends work, and interrupted work continues with one key (settled)

santoshkumarradha:
A design page only. No code changes. The owner answered both of its questions on 2026-09-20, so it is settled rather than proposed and three cells follow it.

## The rule

Closing a window never ends work and never loses it. Work is always in one of three conditions the screen names: running, interrupted, or settled. While the engine is alive a run keeps running. When nothing is driving it, it is interrupted, never failed, and everything it did is kept. On the next launch the person is asked once and `enter` continues. Nothing restarts silently, and the five-minute tick never continues a run.

## What it records as true today

- A live run does not count as work, so the engine retires its conversation at 30 minutes, mid-run.
- A run's context belongs to the turn that proposed it, not to the conversation.
- On reopen an unsettled run row is stamped failed and stopped, over a store that holds every row.
- The supervisor can already release dead claims and carry on from a store. Nothing calls it on reopen.

## The owner's answers

Both yes, in his own words, quoted at the foot of the page:

> 1. `interrupted` is the state word. Yes.
> 2. Continuing always asks first, and the five-minute tick never continues a run on its own. Yes.
>
> The design is settled on those two points. Build it as written.

So `interrupted` joins running, finishing, done, incomplete and your call, and means only "nothing is driving this". The fourth row in the order of work, the tick continuing a run on its own, is refused by that second answer. It is written down as refused rather than merely unbuilt.

## The three cells

| # | Cell | Size | New words |
| --- | --- | --- | --- |
| a | a live run counts as work; the run's context is the conversation's; the engine stays while a run lives | S | none |
| b | on reopen, an unsettled run reads `interrupted` from the store's own state, never failed and stopped | S | `interrupted` |
| c | the card, and continue through the existing supervisor on the existing store | M | none beyond b |

The order is not arbitrary. Cell b reads a condition that cell a is what makes truthful, and cell c draws the word cell b introduces.

Two acceptance notes are now on the page beside that table, because they are the parts most likely to be dropped under time pressure. A cell that has to end an engine ends it by the process id recorded when it was launched, confirmed by reading that process's own executable, never by matching a pattern against a command line and never by signalling a group. And cell b carries the manual in both directions: the new word has to reach the pages, and the page that today tells a person their run ended when codeaf closed has to be found and corrected in the same change, because a page that names a word while denying the thing around it satisfies every gate and still lies.

## Proof

Docs only, so the light gate is the level. On Spark, plain ssh, a detached tree of head `c97770f35`. Exit codes read directly.

| Step | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `make test-laws` | exit 0 |


santoshkumarradha:
Both questions answered.

1. `interrupted` is the state word. Yes.
2. Continuing always asks first, and the five-minute tick never continues a run on its own. Yes.

The design is settled on those two points. Build it as written.

santoshkumarradha:
Closed in favour of #1318, which carries the settled design in one place rather than on a branch.

Cells a and b are already landed (`3d95e7072`, `3d3b7542b`); cell c is in flight; cell d stays deliberately unbuilt.

PR #1225, opened 2026-09-19
fix(resident): a dispatch pass does not reclaim the leaf it just released

santoshkumarradha:
## What

A single resident dispatch pass no longer reclaims a leaf that the same pass just released. claimNext records each leaf it dispatches in a new pass.dispatched set and skips those leaves for the rest of that pass; a later pass picks them up normally.

## Why

Under the busy gate a leaf that ran out (StopBudget) and was released by its worker mid-pass could be reclaimed and re-executed by the same Tick. What a person running background self work would have seen: the same leaf executed twice in one pass, spending its budget twice, before the pass ended. The ordering is forced deterministically, not by load, with an afterDispatch test seam that holds the pass until the worker lands (nil in production). The new test asserts one pass executes the leaf exactly once and leaves it Pending for a later pass.

Change entry: docs/changes/unreleased/1225-resident-dispatch-no-double-run.md.

## Verification

Runs in the manager's stacked gate after the Spark quiet window: go build, go test ./internal/resident/..., gofmt -l ./cmd ./internal, codeaf-changes check, guard and namelaw.


santoshkumarradha:
Landed in santos/dev as squash ac84ab183. Gated as part of a seven-PR stack (#1223, #1222, #1225, #1226, #1228, #1227, #1229) on the stack top fe9eb62af, tree 0a5b96a35: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-fe9eb62af.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1226, opened 2026-09-19
test(session): wait helpers check the condition before the clock

santoshkumarradha:
## What

Two load-robustness fixes in session test helpers. No assertion is weakened.

- waitFor checks its condition before trusting the clock: it evaluates the condition at the top of each iteration and declares a timeout only after a final check, so a condition that becomes true at the deadline is not reported as a timeout.
- TestASlowListenerNeverHoldsTheTurnThatIsTellingIt awaits phaseDesk.settled() before asserting the listener was told, so the async phase post has completed before the count is read.

## Why

Both failed under box load by racing the clock rather than the behavior. The paced-turn budget assertion and the listener-was-told assertion are unchanged; only the wait is made robust. This also covers the previously unowned flake TestASlowListenerNeverHoldsTheTurnThatIsTellingIt.

## Verification

Runs in the manager's stacked gate after the Spark quiet window: whole internal/session suite through scripts/one-suite.sh, plus gofmt, codeaf-changes, guard and namelaw.


santoshkumarradha:
Landed in santos/dev as squash 3a45c3f20. Gated as part of a seven-PR stack (#1223, #1222, #1225, #1226, #1228, #1227, #1229) on the stack top fe9eb62af, tree 0a5b96a35: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-fe9eb62af.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1227, opened 2026-09-19
fix(chatv3): seal the place sweep note at close instead of joining the walk

santoshkumarradha:
## What

The place sweep note is now sealed at process close instead of joined. `startPlaceSweep` runs the walk fire-and-forget; `noteSweep` writes under `sweepMu`, and `v3Process.closeAll` calls `sealPlaceSweep()`, which takes that lock and marks the sweep sealed. A note already mid-write finishes; no note writes after the seal. Close does not wait for the walk.

## Why

The sweep was started and forgotten. On an error it resolves the current home and appends `sweep.log`; a process or test that closed quickly could see that write land after close, under a directory being torn down (the cleanup race that leaves "directory not empty"). Joining the whole walk at close would fix the race but move a pause that grows with the number of conversation folders the machine has held from launch onto quit. Sealing the note gives the same guarantee (no write after close) in O(1): quit waits only for one line already in flight, never for the walk.

## Verification

Focused, no heavy lock:

```
go build ./...
go test ./cmd/codeaf -run TestClosingSealsThePlaceSweepWithoutWaitingForTheWalk -count=2
```

The test blocks the sweep as an unbounded walk, asserts `closeAll` returns within a small fixed bound while the walk is still blocked, and asserts a note attempted after the seal writes no `sweep.log`. Without the seal the late note recreates `sweep.log` after close, so the test fails by construction. Whole `cmd/codeaf` and the rest run in the stacked gate.

## Attribution of the gate cleanup failure

The property test proves a note attempted after close writes `sweep.log` without the seal and nothing with it. The sweep has one writer and one file (`noteSweep`, `sweep.log` under the process home), and `closeAll` seals before a v3-process test's `t.TempDir` RemoveAll runs, so a gate run's "directory not empty" cleanup failure in such a test is consistent with this writer. It was not reproduced in 200 quiet runs on either the sealed or unsealed tree; a raw loop cannot force the interleaving that fails, only the property test's blocked walk can.


santoshkumarradha:
Landed in santos/dev as squash 78e8146cc. Gated as part of a seven-PR stack (#1223, #1222, #1225, #1226, #1228, #1227, #1229) on the stack top fe9eb62af, tree 0a5b96a35: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-fe9eb62af.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1228, opened 2026-09-19
test(session): the pin-the-wire test joins the usage writer before cleanup

santoshkumarradha:
## What

The pin-the-wire test now proves FlushUsage joins the product-owned usage writer before the test body returns. A nil-in-production beforeWrite seam on usageWriter lets the test hold the writer mid-write, so the test can Close the agent and FlushUsage while a write is in flight and assert the ledger file exists afterward.

## Why

The test failed under load with "TempDir RemoveAll cleanup: directory not empty": the usage writer created the ledger after the test body returned. FlushUsage already joins the writer; the test simply was not calling it before cleanup. The seam forces the writer-still-going ordering deterministically, no load, and the assertion pins that flush waits for the product writer. Product behavior is unchanged, the seam is nil in production, so this carries no changelog entry.

## Verification

Manager's stacked gate after the quiet window: whole internal/session suite, gofmt, codeaf-changes, guard and namelaw.


santoshkumarradha:
Landed in santos/dev as squash e88d353b4. Gated as part of a seven-PR stack (#1223, #1222, #1225, #1226, #1228, #1227, #1229) on the stack top fe9eb62af, tree 0a5b96a35: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-fe9eb62af.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1229, opened 2026-09-19
test(provider): order the late-first-token rescue by a signal, not elapsed slack

santoshkumarradha:
## What

Orders the late-first-token rescue test by a transport signal instead of elapsed-time slack. Lane A now carries a `FirstTokenUntil` channel closed only at cleanup, so A cannot emit its first token until B has completed and the rescue has committed a winner. The winner, loser, and cancellation assertions are unchanged.

## Why

The test asserts B wins, A is unnamed, and A is cancelled, but it ordered A after the rescue only by giving A a 300ms first token against B's 5ms: an elapsed bound (`took < 300ms`) with about 80ms of slack. Under gate load that slack could be eaten and A could complete first; the product then correctly commits A and cancels B, and the fixed B-wins expectation flips.

This is a test-side timing expectation, not a product defect. The rescue commit path names the first completed arm the winner under the race mutex and cancels every other registered arm, so a real request cannot overwrite a winner or leave an uncancelled loser (decide/commit at `internal/provider/hedge.go:1521-1564`, settlement at `hedge.go:2007-2065`). Forcing B to complete before A's first token makes the winner follow the transport ordering the product actually guarantees, at any speed.

## Verification

Focused, no heavy lock:

```
go build ./...
go test ./internal/provider -run '^TestALateFirstTokenIsRescuedByTheAlternativeAndTheLoserIsCancelled$' -count=20
```

20/20 forced-order runs pass. `internal/provider/pool` untouched. The whole `internal/provider` suite runs in the gate.

Deliberate negative: with B forced before A there is no wrong product outcome to encode; making A win and still demanding B would only assert an invalid expectation. No product change is supported by this seam.


santoshkumarradha:
Landed in santos/dev as squash fe9eb62af. Gated as part of a seven-PR stack (#1223, #1222, #1225, #1226, #1228, #1227, #1229) on the stack top fe9eb62af, tree 0a5b96a35: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-fe9eb62af.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1230, opened 2026-09-19
remote, tui3: a hosted conversation reads and steers its run

santoshkumarradha:
## What was wrong

`codeaf chat` hosts its engine in another process by default. The surface reads a run through one optional method set: the run's rows, one task's page, six steering verbs and the run summary. The wire carried none of them, so in every hosted conversation the rail drew one row and nothing under it, a task's page could not open, and no verb reached the store. All of it worked under `--no-host` and in tests with a scripted agent.

## What changed

- `session.PlanAgent` is the one method set. The surface names it, and the local engine and the remote client both implement it.
- `internal/remote` carries all ten methods, and holds this line so a method added later fails the build until it is carried:
  `var _ session.PlanAgent = (*Agent)(nil)`
- The surface-door ledger shrinks from 23 to 22 and its law follows an interface alias.
- Plan reads and verbs leave the update loop through the ordered door line. The page's follow read is one at a time, and an answer is folded only into the page that asked.
- The run's plan is re-read on a three-second beat of its own, and only while the reading holds a row that is queued or running. It used to ride on the reading of other windows' work, which a hosted conversation does not have.
- Something does paint at rest: the re-read runs from every frame that draws the rail or the tasks place, and from that place's tick. Before the condition below, that was twenty plan reads a minute over the wire for ever in a hosted window nobody was touching.
- At rest the count is zero. A hosted conversation always has a plan reader, so the beat asks a second question: can any held row still move by itself. The read that finds every row settled is the last. A row of this window's own graph moving starts it again, and a hand-off publishes its row after its store is seeded.
- The run summary's refresh carries how long the caller will wait, never the instant it stops waiting. An engine reached with `--host` runs under another machine's clock, and an instant read against a clock one minute ahead is a refresh cut before it starts.

Built with `make build`, a throwaway home and a two-file repository, `codeaf chat` with no flags and the belt on, driven in tmux at 170 by 50.

1. The same two-part task was handed off hosted and unhosted side by side, and the rail was captured every ten seconds. Before the beat fix the unhosted rail drew the two parts at 70 s and the checks at 80 s, and the hosted rail drew one row until the run ended. After it the hosted rail drew the parts at 10 s, a third part at 20 s, and each check as the run added it.
2. The tasks place listed the conversation with its seven pieces of work, their state, steps and cost.
3. Enter on a part opened its page over the wire: the state line, the description, and four steps with their commands and first output lines.
4. A note typed on that page was drawn under `notes` on the same page a moment later.

## Proof

- `internal/remote` and `internal/enginehost`, whole packages, green. The enginehost test seeds a real store, serves a real agent over a real pipe and reads the rows back through `remote.Dial`.
- Focused `internal/tui3` set, `make test-laws`, `gofmt -l ./cmd ./internal`, `go build ./...`, `go vet`: green.
- `TestTheRunsPlanIsReadAgainOnItsOwnBeat` fails with the beat removed and passes with it.
- Two tests go through a real store, a real session agent, the real remote server and the real client, with the client's plan reads counted. `TestAHostedConversationAtRestReadsNoPlanOverTheWire`: twenty beats at rest cross the wire no times after the first read. It fails with twenty extra reads when the condition is removed. `TestAHostedRailLearnsOfAnAddedPartWithinOneBeat`: a part added to the store by another handle is on the rail one beat later, at the cost of one read.
- Rebased onto santos/dev fe9eb62af; every proof above was re-run on the rebased head 42bbdc38e, and the hosted rail drive was repeated with the conditional beat: parts at 10 s, checks as added.
- No prompt, schema or tool description changes.

## Seen on the drive and not fixed here

- Once a run has ended the rail draws its parts and not the run's own row. It does the same unhosted.
- A click on a rail row does not open its page yet. That is the next pull request.
- A step whose command is a multi-line document spills over several rows of the page.


santoshkumarradha:
Landed in santos/dev as squash ccfef489a. Gated as a three-PR stack (#1230, #1231, #1233) on the stack top 4e176a59a, tree ecf703c56: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-4e176a59a.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1231, opened 2026-09-19
tui3: a task's page opens with the task, and a run's rows on the rail open their pages

santoshkumarradha:
Stacked on #1230. It needs that pull request's wire to be seen at all in a hosted conversation, so read only the commits after `docs: the #1230 change entry`.

## What was wrong, on the owner's screen and on mine

- A run task's page opened with `t-qax3xr is your task in the plan.`, a line written for the worker, and never showed what the task was asked. The owner's words: "task details is not shown like actual task it just says like internal name task4".
- Every step row spent its width on a change into the task's own folder, and the command was cut off.
- With a run on the rail, the run had no row of its own, only its parts. A click on a row did nothing. A run's rows were text with no door.

## What changed

- The page opens with the title, the brief folded with the rest one key away, the declared checks, and the folder said once.
- One function draws a command on the page, the rail and the tree. It drops only a leading change into the folder the head names, and draws a command of many lines as its first line and a mark. The record of what ran is unchanged, and a test pins that byte for byte.
- The rail keeps the run's own row and hangs the run's rows under it. Two rules each hid the other's row: the reading hides the run's plan row because a node row wears its title, and the rail hid that node row because the store held its title.
- The frame and the pointer read one drawn view. The press used to be answered from the rows before the run's were spliced in.
- Every row of a run opens that task's page over the conversation, through the single off-loop page door from #1230. A row with no stored page opens its room as before.

## Hosted drive on the real binary, 2026-09-19

`make build`, throwaway home, two-file repository, `codeaf chat` with no flags, belt on, tmux at 170 by 50.

1. A two-part task ran. The rail drew `two independent parts,… #1` with three parts and three checks under it.
2. I typed a draft, then clicked the last part's row. Its page opened: title, `done · 4 steps`, brief, folder, four steps. The step that wrote a file read `cat > muldiv.go <<'EOF' …` on one row, and the note box and the foot were on screen.
3. `esc` returned to the conversation with `my draft stays` still in the box.
4. A click on the run's own row opened the run's page with its brief, folder and notes.

## Proof

- Focused `internal/tui3` set over rail, plan, page, crumb, room and job tests: green. `make test-laws`, `go test ./internal/e2e/ ./internal/iconlaw/ ./internal/manual/ ./internal/namelaw/`, `gofmt -l`, `go vet`: green.
- `TestARunIsItsOwnRowWithItsPartsUnderItAndEachOpensItsPage` covers the row order, the line's task, and the press.
- No prompt, schema or tool description changes.

[...]

PR #1232, opened 2026-09-19
session: a run works in a copy of its own, and its work comes home when it ends

santoshkumarradha:
## What was wrong

With the bash belt on, a hand-off's workers typed in the person's own folder. The run's landing committed straight onto the branch the person had checked out. The receipt said `in a copy of its own`, which was false. Two hand-offs on different folders would have shared one tree.

## What changed

- A run works in a copy cut by the shipped ground ladder from the folder the task is about, as that folder stands. Every part of one run works in that one copy.
- A second hand-off joins the run underway only when it is about the same folder, and its answer says so. A proposed task about another folder is refused with both folders named and what to do next. That is the one refusal that stays at the run's door. Every other failure still falls through to the shipped engine, which cuts its own copy from the same stand.
- When the run ends its work comes home through `taskTree.comeHome`, the function every shipped task lands through. The copy's work is committed, merged into the folder it was cut from, and the copy is given back. The run is then cleared, as on trunk, so the next hand-off starts a run of its own.
- Work that will not go in keeps its branch in the person's repository, and the outcome note says so once.
- The card says `merged` when the work is in the folder. It used to say `branch kept` for every run that named a branch.
- A hand-off that joined a run ends with it. Its row used to stay `running` after the run had ended.

## What I did not take from the cell

The cell made landing a separate step that a person triggers. That had three faults. A later hand-off joined a run whose workers had stopped and never ran. The waiting door lived only in memory and was lost when the window closed. The conversation was told the work was done while the folder held none of it. The cell also rewrote the shipped road's receipt and two manual pages to say `until you land it`; those are restored, and the worker page gains the sections that are now true.

Built from this branch merged with #1231, because the run is not visible on a hosted screen without #1230. `codeaf chat` with no flags, belt on, a two-file repository with an unsaved file of my own in it. Two hand-offs typed six seconds apart.

1. While the run worked, `git status` in my folder showed only my unsaved file. One copy existed, under the conversation's `trees`. The second hand-off's row hung under the first on the rail.
2. When the run ended the folder held one new commit with the three files the run wrote. My unsaved file was untouched. The copy was gone and no task branch was left behind.
3. The card read `done · 30s · 3 files · merged`. The joined row had settled under the run's row.

On the first drive the card read `branch kept · work` and the joined row was still spinning. Both are fixed in the second commit and pinned by tests.

## Proof

- Whole `internal/session` on Spark under the box lock: see the comment below for the tree and result.
- New tests: a real file comes home committed and the copy is given back; a run that only read changes nothing; a person's commit during the run keeps the branch and never says merged; a hand-off after an ended run is a fresh run in a fresh copy that holds the first run's work; the joined row settles.
- Every belt test now waits for its run to end before cleanup. That removes the intermittent `directory not empty` failure at its cause.
- Focused set with the ratchet, schema and prefix tests, `make test-laws`, manual and name gates, `gofmt -l`, `go vet`: green. No prompt, schema or tool description changes.

## Owed after this and #1231 both land

The task page's folder line reads the conversation's folder. While a run is live it should name the run's copy, and the step rows should drop a leading change into that copy. It is a small follow-up that needs both changes on trunk.


santoshkumarradha:
Whole `internal/session` on Spark under the box lock, belt unset, `make test PKGS=./internal/session`.

| Tested commit | Tree on Spark | Result | Ended |
| --- | --- | --- | --- |
| b73487e25 | `~/src/trees/c293-b73487e25` | `ok  internal/session 241.233s`, exit 0 | 2026-09-19T02:10:21Z |
| 833128cbe | `~/src/trees/c293-833128cbe` | `ok  internal/session 229.526s`, exit 0 | 2026-09-19T02:22:51Z |

The head d6977312b adds only the change entry and three manual sentences on top of 833128cbe. Log: `session-suite.log` in each tree.

santoshkumarradha:
Rebased onto santos/dev fe9eb62af. Whole `internal/session` on Spark under the box lock, belt unset, at the rebased head:

| Tested commit | Tree on Spark | Result | Ended |
| --- | --- | --- | --- |
| 2ff956cb3 | `~/src/trees/c293-2ff956cb3` | `ok  internal/session 243.494s`, exit 0 | 2026-09-19T02:40:16Z |

This branch does not touch `internal/run`.

santoshkumarradha:
Rebased onto santos/dev 4e176a59a, head ef2c5f52b. No conflicts, no code change from 2ff956cb3.

Whole `internal/session` on Spark under the suite lock at ef2c5f52b: `ok github.com/Agent-Field/codeaf/internal/session 237.707s`, exit 0, 2026-09-19T03:06:34Z.

Laptop, same head: build, vet, gofmt clean; the focused session set with the road ratchet, Schema and Prefix tests green; `make test-laws`, the manual and name gates and the change entry check green.

santoshkumarradha:
Landed in santos/dev as squash 98cc2547c. Gated as a five-PR stack (#1220, #1237, #1232, #1235, #1240) on the stack top 608c459f2, tree e0413ce99: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Run 1 of the same commit failed once in cmd/codeaf, a package this stack does not touch (TestALaunchReusesOneEmptySessionAndReapsTheRest, TempDir cleanup: directory not empty); the log is kept and that class has its own diagnosis cell. Logs: spark:~/src/trees/check-sdev-608c459f2.log and .run1-flake.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1233, opened 2026-09-19
fix(run): a self-finished codeaf do run seats its review before it ends

santoshkumarradha:
## What

A `codeaf do` run whose root finishes itself now seats its review before the run ends. When the root's status is terminal Done but its worker goroutine is still in flight (the root task is still in the supervisor's cancel set), the pass no longer accepts the stored ending; it waits for the worker return, which is what seats the review. A root whose return has already been absorbed, or whose terminal status is not Done, proceeds as before, so a completed tree that left an unrelated worker behind still drains and the stored-result path from earlier work is untouched.

## Why

A root worker can write its Done to the store before its goroutine returns. A supervisor pass that observed that stored Done first accepted it and ended the run, so the review round was never seated and the run exited with no check task (do_engine_test.go:551, "check tasks in store = 0, want exactly one"): a real `codeaf do` run whose root finishes itself could skip its review. The deferral is narrowed by the root task's identity in the supervisor's cancel set, so it holds back only the self-finished root return and not a worker the completed tree left behind.

## Verification

The forced-ordering regression `TestSupervisorChecksASelfFinishedRootBeforeAcceptingItsStoredEnding` delays the root worker return across several supervisor passes so a pass deterministically observes Done with no return available; on the base tree the run ends with no check, with the fix it waits and seats one, without machine load. Focused `TestDoOnTheRunEngineChecksASelfFinishedRootAndExitsZeroWhenItHolds` passed 30/30. Full `internal/run` is green, including the completed-tree left-behind-worker law and the root review laws. `go build ./...`, `gofmt -l ./cmd ./internal`, `go run ./cmd/codeaf-changes check`, and `go test ./internal/guard ./internal/namelaw` all pass.

## What bounds the wait

The deferral holds only while the root task is still in the supervisor's cancel set, and it is bounded by exactly the limit that already bounds every in-flight worker: the run's own context. The root worker is launched with that context (`context.WithCancel(ctx)`). If it is killed or errors, its goroutine returns on the run's `finished` channel and is absorbed, which records the root failure and ends the run (`OutcomeIncomplete`) and removes it from the cancel set, so the next pass no longer defers. If it hangs, the caller's wall fires `ctx.Done` in the run loop, which cancels that same context, drains the in-flight workers so their endings land, and returns `OutcomeIncomplete`. The change adds no new wait: a self-finished root whose goroutine had not yet returned was already an in-flight worker keeping the run alive under that same bound; the pass simply stops accepting its stored Done before the return. Covered by `TestRunAnswersIncompleteWhenTheRootWasCancelledAfterItRan` and `TestSupervisorNeverCancelsAWorkerThatWroteItsOwnDone`.


santoshkumarradha:
Landed in santos/dev as squash 4e176a59a. Gated as a three-PR stack (#1230, #1231, #1233) on the stack top 4e176a59a, tree ecf703c56: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-4e176a59a.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1234, opened 2026-09-19
session: every run a conversation has made stays readable

santoshkumarradha:
## What was wrong

A second hand-off in one conversation renames the first run's store to `plandb.db.1` and seeds a fresh one. Every read opened only the live store. The first run's row stayed on the rail, but its page, its parts and its notes could no longer be opened, and the `tasks` tool could no longer answer for it.

## What changed

- Reads take every ended store oldest first, then the live one. The listing carries each run's rows, and a task of any run opens by the id the rail shows. A root's id is the rail's number and is never reused; a part's id is random.
- Only the live run takes the six steering verbs. A task of an ended run answers `that task's run has ended`, and a test asserts the ended store's bytes are unchanged after every verb.
- An ended store never changes, so it is opened once and kept for the life of the conversation. The handles are given back when the conversation closes. The live store is still opened per read so another process's writes are seen.
- One function names the archives for the door that makes them and the reads that find them.
- With no archive the listing is byte for byte what it was, pinned by a test.

Built from this branch merged with #1231 and #1232, because a rail row is not a door without #1231. `codeaf chat` with no flags, belt on. Two hand-offs, the second typed after the first had ended.

1. The session folder held `plandb.db` and `plandb.db.1`. The rail drew `#1` and `#2`.
2. A click on `#1` opened the first run's page: brief, folder, its three notes, its steps.
3. A note typed on that page was answered `that task's run has ended` on the page's own line, with the words left in the box.

## A defect this PR had, and how it was found

An earlier head of this PR froze a conversation's engine for good. It passed the cell's checker and the whole `internal/session` suite, and was found only by a real-binary drive in hosted mode.

- **What happened.** The reader that opens every run's store takes the plan's lock and returns a closer that releases it. With a plan armed and no store to open yet, it returned no stores, and all three callers returned on that before the line that deferred the closer. Nothing released the lock.
- **What a person would have seen.** A run whose rows stop moving, a third hand-off that never ends, and an engine that does not leave when the terminal is closed. It is a race on the first hand-off, so one drive in six hit it.
- **How it was found.** The engine from a failed drive was still alive an hour later and ignored a terminate signal. A goroutine dump showed `Agent.Close`, the plan listing over the wire, the run's driver and the turn loop all waiting on the same mutex.
- **The fix.** The helper unlocks before it answers that there is nothing to read, as the shipped single-store opener does, and each caller defers the closer on the line after the call. The failing test came first.
- **The law.** `internal/guard/lockcloser_test.go` covers every lock handed out through a returned closer, in two halves. The helper never answers nil while locked. Its caller writes no return before the closer is deferred or called, except on the first result being nil. Pointed at the leaking head it names the three exact lines. Two older helpers of this shape are on trunk and neither is exposed.

This is the argument for #1241, which puts that drive in the repo.

## Proof

- Whole `internal/session` on Spark under the box lock at 87dac1550: `ok github.com/Agent-Field/codeaf/internal/session 238.001s`, exit 0, 2026-09-19T04:04:37Z.
- Hosted, real binary, this head: `scripts/hosted-drive.sh` three times in a row, 15 of 15 each time.
- Laptop: build, vet, gofmt clean; the focused session set with the road ratchet, Schema, Prefix, Landing and Stand tests; `make test-laws` including the new law; manual and name gates; change entries. No prompt, schema or tool description changes.

## Order

Head 87dac1550 sits on santos/dev 1cbcb4b7d, which carries #1231, #1232 and #1240.


santoshkumarradha:
Whole `internal/session` on Spark under the box lock, belt unset, `make test PKGS=./internal/session`.

| Tested commit | Tree on Spark | Result | Ended |
| --- | --- | --- | --- |
| 71b939a96 | `~/src/trees/c299-71b939a96` | `ok  internal/session 236.943s`, exit 0 | 2026-09-19T02:47:10Z |

santoshkumarradha:
Rebased onto santos/dev 608c459f2, which carries #1232. Head cdc04bc0f.

What the rebase changed: the archive test's run start gains the stand #1232 added, both PRs' appended tests are kept, the manual section sits beside #1231's rail section with the one sentence this change makes false corrected, and the archive-free comparison builds the row's folder the way the listing does.

Whole `internal/session` on Spark under the suite lock at cdc04bc0f: `ok github.com/Agent-Field/codeaf/internal/session 239.928s`, exit 0, 2026-09-19T03:41:46Z.

Laptop, same head: build, vet, gofmt clean; the focused session set with the road ratchet, Schema, Prefix, Landing and Stand tests green; `make test-laws`, the manual and name gates and the change entry check green.

Hosted, real binary: `scripts/hosted-drive.sh` from #1241 passes 15 of 15 on the merge of this branch with #1232 and #1240, including the earlier run's page opening from the side list and a note on it answered `that task's run has ended`.

santoshkumarradha:
Held at cdc04bc0f by its author before landing; the manager's full check on squash 13c738250 was stopped, not failed. Reason: a lock leak found on the real binary in a hosted drive. The new plan-read helper takes the plan mutex and returns a closer; when the plan is armed but no store can be opened yet and there is no archive, both callers return on zero stores without calling the closer, so the mutex is never released and the conversation's engine stops answering plan reads and cannot close. It is a race on the first hand-off, which is why the session suite never met it. Fix with a failing test first, then it is gated again.

santoshkumarradha:
Landed in santos/dev as squash 72bbd3475. Gated as a four-PR stack (#1242, #1243, #1234, #1241) on the stack top 12840eeb5, tree 0d2a25941: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-12840eeb5.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1235, opened 2026-09-19
tui3: a wait whose stream has stopped arriving paints at the spinner cadence, and the harness that measured it

santoshkumarradha:
## What this changes

A wait whose stream has stopped arriving paints at the spinner's cadence instead of thirty
frames a second. `internal/tui3/app.go` gains `streamFresh()`, which reports whether a
delta has landed within `streamFreshFor` (250 ms), and `paint()` multiplies the frame
interval by `spinnerStep` when the only live term is the wait on the model and the stream
has gone quiet. Animations that count in paints land exactly where they would have at full
cadence, one stride at a time. Any other liveness term, or a stream still arriving, keeps
the full cadence. 31 lines.

It also adds `bench/idle-surface/`, the harness and the measurements that found it.

## The claim, and its exact scope

**This changes the WAIT path only: a turn in flight with no delta arriving.** It does not
touch idle at rest, and the numbers below are not interchangeable with each other.

Three idle figures exist and they differ, because they measure different states on
different profiles:

| state | figure | profile | mode |
| --- | --- | --- | --- |
| settled, nothing in flight | 0.8 to 1.1 percent of a core, flat at K=5, 50, 200 | pinned FRESH | suppressed host |
| turn in flight, no delta arriving | 3 to 6 percent of a core | pinned FRESH | suppressed host |
| settled, on the operator's own live profile | 1.92 percent of a core, and the resident set never settled | operator's live profile | hosted, 2 processes and a daemon |

**Idle at rest on the operator's live profile is open and unexplained.** It is not what
this changes, and cell c300 is investigating it separately. Nothing here should be read as
addressing it.

### Before and after on the wait path

| point | before, percent of one core | after | saving | ranges |
| `5:stream`, both windows held open by a slow stream | 1.639 (pairs 1.600 to 1.667) | 1.033 (1.033 to 1.034) | 0.61 pp | disjoint |
| `5:nodelta`, the held wait, window 1 | 2.589 (2.467 to 2.700) | 2.011 (1.967 to 2.067) | 0.58 pp | disjoint |

**The absolute saving is the same 0.6 percentage points of a core in both cases**, which
says more than the two percentages do: what comes off is a fixed cost, not a proportion.

**Why window 1 alone for `5:nodelta`.** Its second window is not a wait. The stub holds the
first delta for 90 s, the hold expires mid-window, and w2 reads 1.0 to 1.17 on BOTH sides
because by then the surface is simply settled, a state this change does not touch.
Averaging it in dilutes the effect. What makes that choice safe rather than convenient is
that **the cell put the same asterisk on its own nodelta rows before any after-figure
existed**, so it was not made by looking at the result. On the two-window mean the same
point reads 1.845 to 1.500, a saving of 0.34 pp, also with disjoint ranges: **the
conclusion holds under either reading.**

Method for those rows: three pairs per point, the ORDER FLIPPED between pairs
(before-after, after-before, before-after), so a monotone load drift pushes both sides the
same way in one pair and opposite ways in the next. Both binaries built with `make build`
in the same tree with only `internal/tui3/app.go` swapped; they are byte-for-byte the same
size, 52,429,065, so the guard is the **sha256 of the file `/proc/<pid>/exe` points at**,
printed per row. A row is VOID rather than absent if the hash is wrong, if the stub saw no
POST, or if no table row appeared. Runs refuse to start above a printed load limit. Load
recorded at both ends of every run.

**Profile for every row above: the pinned FRESH profile plus one key, `daily_budget_usd`.**
See the reproducibility note below for why that key is there.

## Acceptance, and what is not yet done

Measured here, suppressed-host mode: the wait-path figures above. **Not yet measured in
hosted mode**, which is the mode a person runs; a narrow follow-up cell is briefed for it,
because the harness launches `chat --yolo --no-host --max-cost 5`
(`bench/idle-surface/ns-wait.sh:47`) and every figure in `attribution.md` and `results.md`
therefore describes one process with no engine daemon. Its own `results.md` says the daemon
column reads "none" for every point, which is that flag showing through rather than a
property of the product.

`streamFresh()` decides on `lastDelta`: the stream counts as arriving while a delta landed
within the last **250 ms**. The window matters in both directions and is worth reviewing on
that basis: too long and the clock stays slow into resumed text, so the first words of a
reply stutter; too short and a normal inter-delta gap reads as quiet and the clock returns
to full cadence for nothing. Deltas in a working stream land a few hundred milliseconds
apart, so 250 ms sits below the gap that means "stopped" and above the gap that means
"still going", but it is a judgement and not a measurement.

Still to confirm on the real binary with a real engine and client: that the spinner still
moves during a stalled stream, and that when tokens resume the paint returns to full
cadence with no visible lag.

## Reproducibility note, and four explanations that turned out to be wrong

**The measurements in `bench/idle-surface/attribution.md` and `results.md` are real** and
the stub traffic proves it: every one of the cell's runs shows the expected POST count for
its K (6 at K=1, 26 at K=5, 251 at K=50, 1001 at K=200, and 998 to 1004 across the wait
points) with `first_deltas` matching and a pane state word of `working` or `idle` that
exists only on a real chat surface.

**They are not currently reproducible.** Running the committed harness with the recorded
binary commit and the pinned profile, on a quiet box, produces no rows at all: the surface
stops on the first-run settings screen, the harness polls 300 times at 0.5 s for the
composer, never finds it, falls out of the loop with no failure branch, logs `surface up`
as though it had, and types a turn into a settings page. Adding `daily_budget_usd` to the
profile fixes it, which is why the rows above carry that key.

Why the cell did not need that key is **unknown**. Four explanations were tested and
withdrawn, recorded here so nobody spends the time again:

- **Machine load.** Wrong: it fails identically on a quiet box at load 3.07.
- **The rebase onto newer trunk.** Wrong: the four files that decide first-run are
  byte-identical between the cell's commit and trunk.
- **The build.** Wrong: the cell's own commit, built with `make build`, fails the same way.
- **The profile.** Wrong: the profiles the cell's successful runs left on disk carry nine
  keys with neither `setup_seen_at` nor `daily_budget_usd`, and the harness re-extracts the
  tarball on every point, so the cell's runs reached the composer without either marker.

The first-run screen itself is by design: it opens when stdin is a terminal, the launch is
not a `--host` connection, the conversation was not resumed and has no entries,
`setup_seen_at` is absent, and at least one setup step remains unanswered. A profile
carrying an API key and every model key still meets it until `daily_budget_usd` exists.
That is a product fact for the owner, not a regression, and it is not what this changes.

**The harness's silent non-row is a real defect** and belongs to the follow-up cell rather
than to this PR: a poll that exhausts its attempts should fail, not log success.

## What exactly was measured

The numbers above came from two binaries, and this section pins them to content rather than
to commits, because a rebase changes every sha in this branch and the measurement does not
move with it. The identity that survives is the blob hash of the one file that differed.

| side | binary sha256 (first 16) | `internal/tui3/app.go` blob |
| --- | --- | --- |
| before | `ace79e7e0ae61a84` | `34ee6e566d1dc42858c8a2127eb984e03aa6858d` |
| after | `b7b2ce659b50786a` | `fb40344e917650121daf4a60fa438a85ec938fe7` |

Both were built with `make build` in the same working tree with only that file swapped, so
nothing else differed: no build-tag difference, no regenerated asset, no second worktree.
They are **byte-for-byte the same size, 52,429,065**, which is why the per-row guard is the
sha256 of the file `/proc/<pid>/exe` resolves to and not the size. Every `before` row in the
run carries `ace79e7e0ae6` and every `after` row `b7b2ce659b50`, so no row compared a binary
with itself.

An earlier attempt built the baseline in a fresh worktree and produced a binary 3.5 MB
smaller, which would have compared two different programs rather than one change. That is
why both sides are built in one tree.


santoshkumarradha:
Landed in santos/dev as squash 692d4a15c. Gated as a five-PR stack (#1220, #1237, #1232, #1235, #1240) on the stack top 608c459f2, tree e0413ce99: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Run 1 of the same commit failed once in cmd/codeaf, a package this stack does not touch (TestALaunchReusesOneEmptySessionAndReapsTheRest, TempDir cleanup: directory not empty); the log is kept and that class has its own diagnosis cell. Logs: spark:~/src/trees/check-sdev-608c459f2.log and .run1-flake.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1236, opened 2026-09-19
fix(connect): a reconnect reuses its registered identity instead of introducing itself again

santoshkumarradha:
## What

A reconnect no longer introduces the identity a second time when it can reuse the one it already registered. On reconnect the manager reads the saved registration and tries its loopback redirect before the ordinary listener candidates; when that address is free the reconnect serves on the redirect the service already knows, so the identity check passes and no second introduction is issued. The fixed listener candidates and their order are unchanged.

## Why

After a first connection persisted its registration on a fallback ephemeral redirect (both fixed ports were busy), a disconnect kept the registration but a reconnect could bind a different ephemeral redirect. mcpRegistration.fits then rejected the saved identity, because the service was never told the new redirect, and the reconnect issued a second real registration request: codeaf introduced itself twice where once was wanted (mcp_auth_test.go:283). This is a real reconnect path, not a test counting artifact; the fake counts only requests to its own registration endpoint.

## Verification

A forced-ordering regression injects preselected listeners at the listener.New seam so the initial connection and the sequential reconnect receive distinct ephemeral redirects; on the base the reconnect introduces a second time, with the fix it reuses the saved redirect and introduces once. go test ./internal/connect passes; the fixed-port constants and selection order are untouched. build, gofmt, codeaf-changes, guard and namelaw all pass.

## Releasing the callback listener before a reconnect (follow-up)

Reusing the saved redirect first is not enough on its own. The callback listener was torn down by a fire-and-forget goroutine on the flow's context, and Flow.finish signaled done before that shutdown had closed it, so Flow.Wait could return while the saved loopback port was still bound. A reconnect that followed at once then found the port busy and, through the product's listener.New falling through to the next candidate, bound a fresh redirect and introduced the identity again. beginToolServer now captures when server.Serve returns and the flow's completion cancels its context and waits for that return before reporting done, bounded by the graceful shutdown's own timeout, so a finished flow has released its callback listener and a reconnect finds the saved port free. TestAFinishedFlowReleasesItsCallbackListenerBeforeItReportsDone gates the listener close open and proves Flow.Wait does not report done until the listener is released; without the wait it fails by construction. internal/connect passed through one-suite at -count=20.


santoshkumarradha:
Landed on `santos/dev` as 45137b0af, inside a five-PR stack whose top is 443c661df (tree 53c8bb4da). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-443c661df.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1237, opened 2026-09-19
test(connect): the second-door tests use dynamic ports, not the fixed callback ports

santoshkumarradha:
## What

The internal/connect tests that prove begin-auth falls through to the second door no longer bind the machine-global fixed ports. During those tests the package address list is swapped for test-owned dynamic loopback listeners and the busy-first-door condition is forced in-process, so the anchor test and its two neighbors assert the second-door behavior without competing for 8765 or 18765.

## Why

8765 is real product code, the preferred fixed callback with an 18765 and then an ephemeral fallback, so any unrelated process holding it could fail or stall these tests, and the heavy-suite lock cannot cover a holder that is not one of our suites. The product does not require that exact port for the behavior under test, so the tests use dynamic ports and force the busy-first-door condition themselves. The production address order and binding path are unchanged.

## Verification

go test ./internal/connect passes and no test binds 8765 or 18765. build, gofmt, codeaf-changes, guard and namelaw all pass. Test-only, no change entry.


santoshkumarradha:
Landed in santos/dev as squash b208cccd2. Gated as a five-PR stack (#1220, #1237, #1232, #1235, #1240) on the stack top 608c459f2, tree e0413ce99: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, level with dev 9258f7052. Run 1 of the same commit failed once in cmd/codeaf, a package this stack does not touch (TestALaunchReusesOneEmptySessionAndReapsTheRest, TempDir cleanup: directory not empty); the log is kept and that class has its own diagnosis cell. Logs: spark:~/src/trees/check-sdev-608c459f2.log and .run1-flake.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1238, opened 2026-09-19
test(exec): identify a leaf-end survivor by process identity, not a bare PID

santoshkumarradha:
## What

The leaf-end survivor test now identifies a survivor by process identity, not a bare PID. After leaf teardown it reports a survivor only when the PID is still live and its start time matches the one captured while the survivor was proven running; a missing PID, or a recycled PID that now denotes a different process, is a settled success.

## Why

The test asserted a survivor lived by probing kill(pid, 0) on a bare integer PID. After cmd.Wait reaps the original process, Linux can reuse that PID for an unrelated process, so the probe succeeds and falsely reports the original survived ("process N survived leaf end"). This is a test-side observation hole, not a product leak: product close cannot return while a job's waiter is unsettled, the waiter cannot settle before cmd.Wait reaps the recorded shell leader, and product signaling uses the captured group identity with an ownership check, never a bare PID. The false positive is reachable only in the assertion.

## Verification

A forced-ordering reproduction (a probe seam that returns success only after the real kill(pid, 0) reported the original gone) made the false positive deterministic; the identity-aware assertion then passes 20 of 20 focused runs while still failing on a real survivor whose PID and start time both match. The fix reuses the existing production ProcessStartTime helper, with no second identity scheme. build, go test ./internal/exec, gofmt, codeaf-changes, guard and namelaw all pass. Test-only, no change entry.


santoshkumarradha:
Landed in santos/dev as squash 1cbcb4b7d. Gated as a two-PR stack (#1239, #1238) on the stack top 1cbcb4b7d, tree 69d7c70c3: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-1cbcb4b7d.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1239, opened 2026-09-19
chat: name unread config.json keys once before defaults apply

santoshkumarradha:
## What

A chat conversation now shows a one-time notice when `config.json` carries top-level keys that no reader consumes, naming them, in both hosted and `--no-host` mode. Before this, the unread-key finding added in #1223 went only to process logs that a chat user never sees.

## Why

`config.json` is a flat dotted-key map. A nested block like `{"models":{"tiers":{...}}}` or a mistyped dotted key reads as all-unset, silently, and every affected setting falls to a shipped default with nothing on screen saying why. The log line #1223 added lands in a per-profile file (chat) or `host.log` (hosted), neither of which a person reads.

## How

- **Single source of truth.** `config.warnUnreadProfileKeys` now returns the sorted unread-key list (the once-per-process log is unchanged); `Config.UnreadProfileKeys` carries it. The notice never classifies keys a second way.
- **Both transports converge on `newApp` -> `showUnreadProfileKeys`:**
  - `--no-host`: `settings` -> `v3Process` -> `runSurface(tui3.Options)`.
  - hosted: `proc` -> `bootEngine` -> `remote.Engine` -> `Welcome.UnreadProfileKeys` wire field -> client `hostOptions` -> `tui3.Options` (engine and surface are separate processes, so the list must cross the wire).
- **Gate.** The existing `noticeBoard`/`noticeLedger`, key `unread-profile-config-keys-<sha256 of the sorted keys>`. Shown once per profile; re-shows only when the unread set changes.

## Real-binary hosted acceptance (real engine + client, default hosted mode)

Captured from a build of this branch head, `codeaf chat` (hosted; a daemon spawned), throwaway profiles:

| Scenario | On screen |
|---|---|
| nested `{"models":{"tiers":{"reflex":...}}}`, first launch | `config.json keys are not read: models; anything set under them is ignored and defaults apply.` |
| same profile, second launch | (nothing) |
| same profile after adding a second unread key | `config.json keys are not read: models, workers; anything set under them is ignored and defaults apply.` |
| a correct flat config (`models.tiers.reflex`) | (nothing) |

The wording is general ("anything set under them is ignored and defaults apply") rather than the earlier "models fell back to defaults", which overclaimed for non-model keys: a real profile carries `display.theme`, which no reader consumes. Model config, the motivating case, still reads correctly under this wording.

## Tests

- `internal/config` `TestLoadWarnsOnceForEveryUnreadTopLevelProfileKey`
- `internal/tui3` `TestUnreadProfileKeysNoticeTracksTheSet` (nil silent / named / repeat silent / changed shows / prior silent, across ledger persistence on disk)
- `cmd/codeaf` `TestHostedWelcomeCarriesUnreadProfileKeysToSurface` (a real `remote.Loopback` engine and client carry the list into `hostOptions`; not a fake agent)

## Verification (Spark; full suites never run on the laptop)

- `go build ./...`, `go vet` (scoped), `gofmt -l ./cmd ./internal` empty
- the three focused tests plus `internal/guard` and `internal/namelaw`: all ok
- `internal/session` full package under the one-suite lock: PASS
- `go run ./cmd/codeaf-changes check`: clean
- the real-binary hosted captures above

Base: 4e176a59a.

## No product-written key is ever flagged (review follow-up)

`display.theme` (seen on one Spark profile) was never written by the product: its only trace in history is a comment in `internal/tui3/styles.go` showing `settings.Get("display.theme")` as an example. It was never a settings row and is never persisted, so that key was hand-added, and the notice correctly names a genuinely unread key. Santosh's live laptop profile (its exact top-level keys) yields no notice. `TestNoShippedWriterKeyIsReportedUnread` locks the property: a profile of every settings-registry row plus every non-setting field the loader writes has an empty unread list, so the notice can never flag a key a person never typed.


santoshkumarradha:
Landed in santos/dev as squash 2f6559aba. Gated as a two-PR stack (#1239, #1238) on the stack top 1cbcb4b7d, tree 69d7c70c3: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-1cbcb4b7d.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1240, opened 2026-09-19
tui3: a task's page says only what is true of it

santoshkumarradha:
Follow-up to #1231. Four things a run task's page said that were not true, each with a failing test first.

## What changed

- **An ended task is offered no refused verb.** A `done` or `incomplete` task's row and page name no `x stop it` and no `p pause`. The store refuses both.
- **A note's author is a word a person knows, or nothing.** A note you left reads `you`. Any other note names no author. The page test now asserts that no id of the store's appears anywhere on the page.
- **An ended task's page promises no next step.** The sentence `the worker reads a note at its next step` is absent once the task has ended. One helper answers "has this task ended" for the key line and for that sentence.
- **A letter typed into the note is a letter.** The stop card's key is read above every page. The page opened from the side list sits over a conversation whose own box is empty, so a note holding that letter raised the card mid-word and the card swallowed the rest. A run task's page is now on the card's precedence list wherever it was opened from. The page still reads the key itself over an empty box. This fault came in with #1231 and was found by the hosted drive below, not by a test.

## Hosted drive on the real binary

`bin/codeaf chat` from this branch, hosted as it runs by default, throwaway home and repository, run switch on, 2026-09-19.

1. Handed off one task with `/task`. Clicked its `#1` row on the side list. The page drew over the conversation.
2. Typed `keep the examples short` and pressed enter while the task ran. Before the fix the box held `keep the e` and the store held no note. After it the page shows:
   ```
    notes
    you · now
    keep the examples short
   ```
   and the foot reads `the worker reads a note at its next step` over `↑↓ scroll · enter send · x stop it · p pause · esc back`.
3. Waited for the task to end. The same page's foot reads `↑↓ scroll · enter send · esc back`, with no sentence about a next step.

## Proof

Laptop, head of this branch: `go build ./...`, `go vet ./internal/tui3`, `gofmt -l ./cmd ./internal` clean. `make test-focus PKGS=./internal/tui3 RUN='Plan|Page|Rail|Note|Verb|Manual|Frame|Ended|Finished|Stop|WorkTab|Work'` ok. `make test-laws` ok. `go test ./internal/e2e/ ./internal/iconlaw/ ./internal/manual/ ./internal/namelaw/` ok. `make changelog-check` ok. No prompt, schema or session change.

Manual: `worker-harness.md`, `reading-a-task-page.md` and `task-controls.md` say the new behaviour in the same change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1241, opened 2026-09-19
scripts: a repeatable hosted drive of a run on the real binary

santoshkumarradha:
Adds `scripts/hosted-drive.sh`, a repeatable drive of a run on the real binary in hosted mode. It is the acceptance a person would do by hand, written down so anyone can repeat it.

**Lands after #1234.** Phase 3 asserts that an earlier run stays readable, which #1234 brings. On this trunk without it, phase 3 fails and phases 1 and 2 pass.

## What it does

```
make build
scripts/hosted-drive.sh bin/codeaf
```

It opens `codeaf chat` in a tmux server of its own, with a throwaway home and a throwaway repository. It types three hand-offs and reads the screen and the folder back. It prints PASS or FAIL per check and exits with the number that failed. A failed check keeps the screen it judged.

The fifteen checks, in a person's words, are in the script's header. In short:

1. **The run works in a copy of its own.** Your folder holds only your own unsaved file while it works.
2. **Its work comes home.** One new commit, your file untouched, the copy given back, no task branch, the card says merged, nothing still running.
3. **An earlier run stays readable.** Its record is kept, a press on its row opens its page with no id of the record's own, a note on it is answered in words, and esc returns to your draft.

## What it is careful about

- It makes real model calls, so it is wired into no make target.
- It never writes to your own home's state. It copies the profile's config file if there is one and otherwise relies on the key in the environment.
- It ends only its own tmux server, by its own socket name. It selects no process by command-line text.
- It removes only the folders it made, and only when everything passed.
- `shellcheck` is clean. One info class is disabled with its reason: the checks are functions handed by name.

## Proof

ALL PASS, 15 of 15, on the real binary built from the merge of 4e176a59a, #1232, #1234 and #1240, on 2026-09-19. `shellcheck scripts/hosted-drive.sh` exits 0. `go test ./internal/namelaw/` and `make changelog-check` ok. No Go change.


santoshkumarradha:
Landed in santos/dev as squash 12840eeb5. Gated as a four-PR stack (#1242, #1243, #1234, #1241) on the stack top 12840eeb5, tree 0d2a25941: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-12840eeb5.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1242, opened 2026-09-19
config: concurrent profile writes in one process compose

santoshkumarradha:
## What

`writeProfileValues` (internal/config/budget.go) reads config.json, copies in the update, marshals, and renames a temp file over it. A process-wide `profileWriteMu` now holds the whole read-copy-rename as one transaction, and update values are marshaled before the lock so a user-defined marshaler never runs inside the profile critical section.

## Why

Two goroutines in one process that both read config.json before either rename lose one update: the later rename discards the earlier writer's key. Serializing the transaction makes concurrent writers in one process compose.

## Scope, said plainly

This composes concurrent callers of `writeProfileValue` / `writeProfileValues` inside ONE process. Today's callers are settings writers that mostly run one at a time, so this hardens the writer against concurrent use rather than fixing a known concurrent caller. It makes no claim across processes: two codeaf processes writing config.json at once (chats side by side) still rely on the atomic rename to avoid partial JSON, and the cross-process lost update remains open. That cross-process case is being addressed separately.

## Verification

`go test ./internal/config -count=20` (ok, 69.9s), `go build ./...`, `gofmt -l ./cmd ./internal` (empty), `go run ./cmd/codeaf-changes check`, `go test ./internal/guard ./internal/namelaw`, all green on Spark. `TestConcurrentProfileWritesCompose` forces both writers past the profile read before either rename with blocking JSON marshalers and asserts both keys survive; it fails without the mutex.


santoshkumarradha:
Landed in santos/dev as squash 8f990d5e6. Gated as a four-PR stack (#1242, #1243, #1234, #1241) on the stack top 12840eeb5, tree 0d2a25941: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-12840eeb5.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1243, opened 2026-09-19
run: regression test for seating the check after a self-finished worker returns

santoshkumarradha:
## What

Adds `TestSupervisorSeatsTheCheckAfterAStoreDoneWorkerReturns` (internal/run/review_test.go). Test only, no product change.

## Why

`TestDoOnTheRunEngineSeatsACheckOnTheCheckModel` flaked once under box load and passes alone. Forcing the post-Done, pre-return interval with a deterministic barrier while the supervisor holds one slot shows the check is still seated correctly: the occupied-slot invariant keeps that interval from losing or mis-seating the check. No engine defect, no product change; the new test pins the ordering so a future regression is caught deterministically.

## Verification

`go test ./internal/run` (ok) and the focused test `-count=20` (ok), `go build ./...`, `gofmt -l ./cmd ./internal` (empty), `go run ./cmd/codeaf-changes check`, `go test ./internal/guard ./internal/namelaw`, all green on Spark.


santoshkumarradha:
Landed in santos/dev as squash 5d9c47b7c. Gated as a four-PR stack (#1242, #1243, #1234, #1241) on the stack top 12840eeb5, tree 0d2a25941: full make check on Spark from a fresh clone, EXIT 0, 114 packages ok, first run, level with dev 9258f7052. Log: spark:~/src/trees/check-sdev-12840eeb5.log. Closed rather than merged because the squash went in through the manager's stacked side branch.

PR #1244, opened 2026-09-19
tui3: what is typed after a press on a run's row goes to its page, never the conversation

santoshkumarradha:
What a person types right after pressing a run's row in the side list now goes to that task's page. Before this, it went to the conversation and was sent to the model.

## What was wrong

A task's page is read off the update loop. On a hosted conversation the answer took 2.4 s on a real screen, and up to 7.7 s under load. Until it arrived the conversation was still drawn and its box still took the keyboard. On the real binary, `keep the examples short` and `enter`, typed at once after the press and meant as a note for the task, was answered by the model and saved as a standing preference nobody asked for.

## What changed

- **The gap's keys belong to the page.** From the press on, keys are kept in order and handed to the page's own keyboard when it opens, `enter` included. `esc` withdraws the press. A second press replaces the first. A row that turns out to have no page opens its room as before, and those keys are dropped.
- **The stop card stands down while a page is on its way.** Its key is read above every page. With a task running, a note holding that letter still raised the card in the gap and lost the rest. This was found at review with a failing test first and is the same class as the fault #1240 fixed for a page that is already up.
- **Manual.** The section on opening a run's task from the side list says the page can take a moment and what happens to keys meanwhile.

The first two commits are the cell's. The third closes the stop card hole and documents the pending state.

## Hosted drive on the real binary

`bin/codeaf chat` from this branch, hosted as it runs by default, throwaway home and repository, 2026-09-19.

1. Handed off one task. Pressed its row on the side list, then typed `keep the examples short` and `enter` at once, with no wait.
2. The page drew with:
   ```
    notes
    you · now
    keep the examples short
   ```
3. Back in the conversation there was no message and no reply. The words exist in the run's record and nowhere else under the conversation's folder.

## Where the seconds go, measured

A temporary timer on the ordered door line, removed before this commit, on the same drive:

| Door | Asked | Took |
| --- | --- | --- |
| The page's read | once | 2 ms, after waiting 7586 ms in line |
| The run's summary | once | 10002 ms |
| The open page's follow | 509 times in about 90 s | 2 ms each |

The wire, the store read and the fold are two milliseconds together. The wait is one door: the run's summary asks a model with a ten second budget on the same ordered line as a person's gestures, so a press made meanwhile waits behind it. That is its own change and follows this one. The follow reading about six times a second is reported separately.

## Proof

Laptop, this head: `go build ./...`, `go vet ./internal/tui3`, `gofmt -l ./cmd ./internal` clean. `make test-focus PKGS=./internal/tui3 RUN='Rail|Plan|Page|Note|Stop|Frame|Hosted|Manual|Work'` ok. `make test-laws` ok. `go test ./internal/e2e/ ./internal/iconlaw/ ./internal/manual/ ./internal/namelaw/` ok. `make changelog-check` ok. tui3 only: no session, prompt or schema change.


santoshkumarradha:
Landed on `santos/dev` as 5a7748c19, inside a five-PR stack whose top is 443c661df (tree 53c8bb4da). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-443c661df.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1245, opened 2026-09-19
config: unread notice skips keys the product itself retired

santoshkumarradha:
## What

The unread-config-key notice (#1223's load-time log, #1239's chat notice) named top-level `config.json` keys that a shipped version once wrote as a real settings row but that nothing reads at head, telling a person about an ignored key they never typed. Seven such retired keys are now skipped silently.

## The seven retired keys

`linear_mode`, `nerd_font`, `rail_state` (readers removed by 74e16993c), `work.workers` (05b995376), `memory.consolidation` (10dcdfdd4), `practice_demand_pct`, `propose_new_skills` (84ba8503e). Each was a real settings-registry row in a shipped version and is read by nothing at head. A person on an older profile (nerd_font and linear_mode were shipped v3/tui2-era settings) was being told these keys are ignored.

## How

- `retiredProfileKeys` in `internal/config/config.go`, each entry dated by its reader-removal commit; `warnUnreadProfileKeys` skips them silently.
- `consumedProfileKeys` is extracted as the single definition the unread check and the ledger law both read, so the two cannot drift.

## Regression-proof by property, not memory

- `internal/config/testdata/profile-keys.ledger` lists every top-level key the product has ever written (current writers plus the seven retired).
- `TestProfileKeyLedgerLaw`, two halves: (a) every key a current writer can produce is in the ledger; (b) every ledger key is either consumed at head or in `retiredProfileKeys`. Adding a settings row forces a ledger line; removing one leaves its line behind and turns the law red until the key is retired on purpose.
- `TestRetiredProfileKeysAreNotReportedUnread`: each retired key alone is silent, and a genuinely unknown key alongside them is still named.

## Real-binary capture (this branch head)

- profile `{linear_mode, nerd_font, work.workers}`: nothing on screen.
- add a genuinely unknown key: only `genuinely_unknown` is named.

## Method and its blind spot

The seven were found by a static history sweep (c312): grep the settings-registry key literals across the dev history, subtract head's consumed set, confirm each remaining key was a shipped registry row with no reader at head. A key retired BEFORE the settings registry existed would not appear as a registry-row literal, so the sweep would not catch it; such a key is not covered here. The ledger law will still catch any FUTURE writer added or removed.

## Verification (Spark; full package never on the laptop)

- `go build ./...`, `go vet`, `gofmt -l ./cmd ./internal` empty
- `internal/guard` + `internal/namelaw` ok
- `internal/config` whole package under the one-suite lock: ok
- `go run ./cmd/codeaf-changes check` clean (after the pr-number reconcile)

Base: 1cbcb4b7d. Leads the next stack with #1234 and #1241.


santoshkumarradha:
Landed on `santos/dev` as 7eab247a0, inside a five-PR stack whose top is 443c661df (tree 53c8bb4da). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-443c661df.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1246, opened 2026-09-19
session: make the claim-unspent revision test deterministic by forcing the ordering

santoshkumarradha:
## What

Rewrites the interleaving in `TestANewDirectionArrivesWithTheClaimUnspent` (internal/session/completion_stale_test.go) so it is deterministic. Test only, no product change.

## Why

The test flaked under box load and passed alone. It queued the new direction the moment the first drop was taken, an interleaving that shifted with load. It now queues the direction in the first ordinary response after the first drop, the narrow window the case is about, and asserts the forced ordering fired. The product behavior is correct and unchanged: no revision inherits the claim spent on the sentence before it, so no task is admitted.

## Verification

`go test ./internal/session` through scripts/one-suite.sh (ok) and the focused test `-count=20` (ok), `go build ./...`, `gofmt -l ./cmd ./internal` (empty), `go run ./cmd/codeaf-changes check`, `go test ./internal/guard ./internal/namelaw`, all green on Spark.


santoshkumarradha:
Landed on `santos/dev` as 1e332c312, inside a five-PR stack whose top is 443c661df (tree 53c8bb4da). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-443c661df.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1247, opened 2026-09-19
test(connect): stop the second-door Slack test racing on a freed port

santoshkumarradha:
## What

Test only. `TestSlackBeginAuthUsesTheSecondDoorWhenTheFirstIsBusy` (internal/connect/slack_test.go) no longer flakes under load with 'both of the addresses ... are busy on this machine'.

## Why

The test held the first door busy, closed the second, and relied on the product rebinding the freed port. Slack's doored path has no listener seam (oauth2cli takes an address string, not a listener) and probes only the two vendor doors with no ephemeral fallback, so the second port cannot be held open by the test; under load another listener took it in that gap and both doors read busy. The test now retries the sign-in with a fresh free second address until it lands on the second door, bounded (50 attempts). The behavior under test is unchanged: a real regression would send the browser to the wrong port, which the existing URL check still catches; only the machine race is absorbed.

## Verification

`scripts/one-suite.sh go test ./internal/connect -count=200` on this branch: 0 failures, slack test 200/200 (loadavg 3.07, log kept). Combined with the #1236 held-listener fix, the full connect suite is 0 failures at -count=200. `go build ./...`, `gofmt -l ./cmd ./internal` (empty), `go run ./cmd/codeaf-changes check`, `go test ./internal/guard ./internal/namelaw` all green.

## Coordination

This and the #1236 held-listener change each fix one of two connect flakes; a full-suite -count=200 is zero-failure only with BOTH present, so they should land together (or this one first, then #1236 rebased onto it).


santoshkumarradha:
Landed on `santos/dev` as 443c661df, inside a five-PR stack whose top is 443c661df (tree 53c8bb4da). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-443c661df.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1248, opened 2026-09-19
config: make a profile write safe across processes with a cross-platform lock

santoshkumarradha:
## What

`writeProfileValues` (internal/config/budget.go) now takes a cross-process exclusive lock on a stable `config.json.lock` across the whole read-copy-rename, inside the existing in-process `profileWriteMu` (mutex outer, file lock inner). New `internal/config/profilelock.go` wraps `internal/filelock` (cross-platform Lock/IsBusy) with a bounded 2s wait built from IsBusy polling, so no second lock primitive is introduced.

## Why

Two codeaf processes writing config.json at once, a person running chats side by side, each read the file, add their own key, and rename; the later rename discarded the other's key, so a setting saved in one chat silently vanished. #1242 serialized concurrent writers within one process; this closes the cross-process case a person actually hits.

## Lock semantics

- Bounded wait: 2 seconds of IsBusy polling, then a plain `timed out` error, so a write never hangs a chat.
- No stale lock: the hold lives in the OS lock on the open file, released on close and, on a crash, when the OS closes the dead process's descriptor. The `.lock` file is never unlinked.
- Cross-platform: taken through internal/filelock, which the repo already ships for unix and Windows.

## Test

A forced two-process test re-execs the test binary twice, holds both at a real-flock barrier until both are contending, then shows both keys survive. A second test shows the 2s timeout; a third shows recovery after a holder crashes. These forced tests are `//go:build unix` (they build the barrier with flock); Windows recovery is by handle close, provided by internal/filelock. The test states which platforms it proves.

## Verification

`go build ./...` (linux), `GOOS=windows go build ./internal/config` AND `GOOS=windows go build ./...` (both pass; this is the check that catches the undefined symbol), `go test ./internal/config -count=5` (ok, includes the two-process/timeout/crash tests), `gofmt -l ./cmd ./internal` (empty), `go run ./cmd/codeaf-changes check`, `go test ./internal/guard ./internal/namelaw`, all green on Spark.

## Gate gap noted

`make check` is `vet fmt-check test test-packed-manual size`; it does NOT cross-compile Windows (cross builds are staging/nightly per the Makefile), so the Spark gate would not catch a Windows-only undefined symbol. This PR was verified with an explicit `GOOS=windows` build. A build-only cross-compile step for the platforms the repo ships (read from the release config, no hardcoded names) is worth a separate small PR.

## Related

Reworked from the cross-process profile-lock cell, which first shipped a unix-only flock that broke the Windows build and duplicated internal/filelock. This is the cross-platform version.


santoshkumarradha:
Landed on `santos/dev` as d38345f64, inside a two-PR stack whose top is 8ed67b790 (tree ec1daf2e4). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, on the second run. The first run failed one test in `cmd/codeaf`, a package this stack does not touch (`TestOpeningAndClosingConversationsLeavesTheProcessHoldingNothing`, temp directory not empty at cleanup, the known late home writer class, under heavy box load); its log is kept at `~/src/trees/check-sdev-8ed67b790.run1-flake.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1249, opened 2026-09-19
test(exec): read the job shell tmux floor by a marker line, not a full env dump

santoshkumarradha:
## What

Test only. `TestJobShellEnvCannotReachTheHostTmux` and `TestBackgroundJobShellCannotReachTheHostTmux` (internal/exec/tmuxenv_test.go) now assert on a single marker line instead of dumping and parsing the whole environment.

## Why

runShell and the background job log are captured through a capped output collector (preview budget). A full `env` dump overflows that preview on a machine with a large ambient environment, so the TMUX_TMPDIR line is cut off and the test reads it as absent although the job shell had it. That failed the foreground test on macOS (large ambient env) while passing on Spark (small env), and it made the TMUX and TMUX_PANE checks pass vacuously on any large env, since absence in a truncated dump proves nothing.

Confirmed on Darwin with an uncapped tee: the full env output was 25264 bytes but the captured preview was 4234, and the full output did contain TMUX_TMPDIR at the private dir. So the product was already correct; only the test could not see it.

## Fix

Both tests ask the job shell to print exactly the three variables on one short line, `JOBTMUX TMUX=[...] TMUX_PANE=[...] TMUX_TMPDIR=[...]`, with an explicit `unset` marker for any that is absent, through the same bash login path the product runs. The shared helper fails if the marker line is missing from the captured output, so a future truncation can never read as a pass. One short line always fits the preview.

## Scope note (capped-collector class)

A grep of internal/exec test files found no other test asserting on a full environment or other large dump whose asserted line position depends on ambient size; the rest assert on short, deterministic outputs (printf of a known string). The two fixed here are in the same file. The repo-wide capped-collector class is being swept by a separate cell.

## Verification

Darwin (macOS laptop), one focused test each: both PASS. Spark: whole internal/exec through scripts/one-suite.sh (green), plus `go build ./...`, `gofmt -l ./cmd ./internal` (empty), `go run ./cmd/codeaf-changes check`, `go test ./internal/guard ./internal/namelaw`.

## Related

Supersedes the c316 approach (a product change based on a login-shell-clears theory that does not hold on macOS); this is the real cause and a test-only fix.


santoshkumarradha:
Landed on `santos/dev` as 8ed67b790, inside a two-PR stack whose top is 8ed67b790 (tree ec1daf2e4). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, on the second run. The first run failed one test in `cmd/codeaf`, a package this stack does not touch (`TestOpeningAndClosingConversationsLeavesTheProcessHoldingNothing`, temp directory not empty at cleanup, the known late home writer class, under heavy box load); its log is kept at `~/src/trees/check-sdev-8ed67b790.run1-flake.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1250, opened 2026-09-19
fix: a person's stop reaches a run

santoshkumarradha:
## What was wrong

With the belt on, nothing a person pressed could stop a run. Measured on the real binary, hosted, at trunk 12840eeb5 on 2026-09-19:

| gesture | what happened | count |
| --- | --- | --- |
| `x` over an empty box, then `stop it` on the card | `there is no task 1 in this session`; the run carried on and landed by itself | 4 of 4 |
| `x stop it` on the run's own page | `the harness owns the root task` | every time |
| `p pause` on the run's own page | `the harness owns the root task` | every time |

Cause, read in the code. A run's rows wear task numbers and are not in the task graph, and the stop a surface sends for a row is `task:N`, which only the graph answered. The run was started on a context nothing could cut, and no cancel was kept anywhere. The store refuses every verb on the run's own task for every caller, and the page offered two of them.

## What changed

**The store** (`internal/plandb`). `StopRoot` is the runtime's verb for a person's word. It ends the run's own task and every task still open in one write, and leaves ended tasks as they ended. A run left open in its store is adopted by the next hand-off, so a stop has to be written there.

**The session.** The id a surface already sends is resolved to whoever owns the row: the live run first, the graph second. Nothing on the wire changes, so a window built before this change stops a run through an engine built after it, and the model's own stop ends a run too. A stop goes in this order: the store first, so no part reads as a failure and the next hand-off cannot adopt the run; the context next, so every worker and every call it has out ends now; the work last, committed on the run's own branch with the copy given back. Nothing is merged into the folder and no model turn is bought to narrate the ending. A stop on a row that joined a run ends that work only. The page's stop on the run's own task is the same stop, and on a run nobody is running any more it ends the run in the store.

**The surface.** `x` on the run's own page raises the stop card, and the page steps aside for it the way a background job's page does. `p pause` is not offered under the run's own task and is a letter there. `ctrl+c` is read before a page that is on its way and before the page. Every ending of the page's read ends the key hold, including an answer for a tab the person has left. That read is already bounded by the wire's own call deadline, which answers no page. A key that leaves the page ends the replay, so `x` typed while the run's own page opens reaches the card.

**What a person is told.** On the run's page and once in the conversation:

```
stopped · its work so far is kept on <branch> and did not go into <folder> · merge that branch to bring it in, or delete it to drop it
```

A run stopped before it changed anything says `stopped · it had changed nothing`, and then neither the sentence nor the row names a branch.

**No further model call.** A surface asks for a run's four-line summary again whenever its rows move, and a stop moves them. The engine declines a summary for a run a person stopped; the last reading stands.

## Two laws

- `internal/session`: every function that publishes a row in a state a stop means something in is listed with the cancel kind that reaches its owner and the test that proves it by stopping one. A new publisher fails the law until it has both. I removed the run's entry to check it: the law named `startKnownTaskRun` and its line.
- `internal/tui3`: for every kind of row a run's store can hold, every verb word the foot offers is pressed against a real store over the real wire, and none may be refused. I put `p pause` back under the run's own task to check it: the law quoted the store's refusal.

## Not in this change

`ctrl+c`, `/quit` and a closed window do not stop a run today: the window closes, the run carries on in the engine and its work lands by itself. Measured three of three for each at 12840eeb5, and ten of ten in a separate reproduction with the terminal killed under the window. That is unchanged here. It is a separate open question and not a regression of this change.

## Hosted acceptance, real binary, belt on

Nine drives at 8ca974cb4, the same code without the summary commit, then eight ledger probes on the later heads.

| path | card raised | row out of `running` after `stop it` | landed in the folder | copy given back | work on the kept branch |
| --- | --- | --- | --- | --- | --- |
| `x` in the conversation | 3 of 3 | 0.26, 0.27, 0.27 s | 0 of 3 | 3 of 3 | 26, 14 and 8 files |
| `x` on the run's own page | 3 of 3 | 0.27, 0.27, 0.54 s | 0 of 3 | 3 of 3 | 18, 14 and 22 files |
| `x` typed in the page-load gap | 3 of 3 | 0.27, 0.26, 0.26 s | 0 of 3 | 3 of 3 | 6, 6 and 12 files |

Child pids: the engine's pid and its child pids were recorded before each stop. Five drives had no child alive at that instant and four had one shell child. Every recorded pid was gone after the stop, and the engine had no children, nine of nine.

Page foot on the run's own task, every page drive: `↑↓ scroll · enter send · x stop it · esc back`.

Usage ledger rows stamped after the stop instant, read 60 s after it:

| head | path | rows before | rows after |
| --- | --- | --- | --- |
| f55c27ebe | conversation | 4 | 0 |
| f55c27ebe | page, three probes | 4, 3, 3 | 1, 1, 1 |
| f55c27ebe | gap | 2 | 0 |
| cdd8e777a | page, three probes | 5, 12, 9 | 0, 1, 0 |
| cdd8e777a | conversation | 6 | 0 |
| cdd8e777a | gap | 6 | 0 |

No billed row was ever stamped after a stop. The one kind of row that was is a marker, not a call: a model name, `unbilled: true`, no dollars, no tokens, no role, about 20.5 s after the stop. The engine writes it when it looks up the receipt of a call whose stream ended before its usage block and finds none. A stop cuts a call mid-stream, so the marker is the bookkeeping of a call that began before the stop and was cut by it. I first read it as a new summary call. The summary commit is still right, because a test shows the engine did make that call when asked, twice, and now makes none. It is not what the marker is, and the body says so rather than claim zero.

**The day figure on screen lags by a beat.** In one drive it moved after a stop. The ledger had no row for that; it was earlier spend drawn late. Nobody should read that figure as spend after a stop.

One drive took 7.09 s for the row to leave `running`, with a shell child alive at the stop. The row says `running` for that time. A small follow-up branch gives the row the `stopping` reading a stopped task already has; it is not in this PR so this head stays the one that was proven.

All drives used a throwaway home with private permissions. Each removed the key copy first, ended its own engine by socket holder and executable equality, then removed the home. No drive home remains.

How "the worker is gone and the spend has ended" is checked. The worker is not a process: it is a goroutine in the engine, and its steps are short-lived shell children. So each drive records the engine's pid and its child pids before the stop, and checks those pids are gone after it. Spend is read from the usage ledger's rows stamped after the stop instant. The day figure on screen is drawn on a beat and can show earlier spend late, so it is not the evidence.

## Proof

- Head aad44e486 sits on #1251's head bd0c2af56. It is cdd8e777a plus one line in the change entry.
- Spark, tree at cdd8e777a, under the one-suite lock: whole `internal/tui3` exit 0 in 324 s, whole `internal/session` exit 0 in 246 s.
- Spark, same tree: `internal/plandb`, `internal/run`, `internal/manual`, `internal/guard`, `internal/namelaw` ok; `make test-laws` exit 0; `gofmt -l ./cmd ./internal` empty.
- Failing tests first, each seen red with the sentence from the real screen: `there is no task 71 in this session`, `the harness owns the root task`, the foot offering `p pause`, `ctrl+c` held with the page's keys, the hold left standing, two summary calls after a stop.
- Manual: two new sections in `worker-harness.md` with three probes, and the sentence about the run's own task in `task-controls.md`. The store's old refusal is no longer quoted as the answer.


santoshkumarradha:
Landed on `santos/dev` as 7b8be4bd6, inside a stack whose top is 992779ed6 (tree 5c6d4b87b); the second commit removes a cell's `FINDINGS.md` that had ridden into the repository root with #1239. This PR was stacked on #1251's head, so its diff over that head was applied onto the trunk that already carried #1251; the directories it touches are byte-identical to the PR head aad44e486. Full `make check` on Spark on 992779ed6: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-992779ed6.log`; format clean and a Windows cross-build of `./...` ok on the same tree. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1251, opened 2026-09-19
fix: a gesture never waits behind a run's summary, and an open page reads on a beat

santoshkumarradha:
## What was wrong

Every ask the surface makes of its engine stood in one ordered line, so the engine sees a person's gestures in the order they were made. The run's summary stood in that line too, and the summary asks a model with a ten second budget. A press made while a summary was out waited behind it.

Separately, an open task page asked for its page again from the paint clock: about six reads a second over the wire for as long as the page was open, ended tasks included.

## What changed

Two commits, each with a failing test first.

1. **Nothing a person did not press stands in the ordered line.** A second road off the update loop, `besideLine`, asks on its own goroutine and folds the same way. Its property: a read nobody pressed for, where nothing a person does next depends on the engine having seen it first. The run's summary is its one caller. A ledger law (`doorsBesideTheLine`, `TestOnlyReadsNobodyPressedForAreAskedBesideTheLine`) lists every caller, and the structural law that keeps engine doors off the loop accepts both roads.
2. **An open page reads on a beat.** Once per three second beat, counted from the last page read, and only while the task is queued or running. A page on an ended task is read once, to open it. The follow stays in the ordered line, because its answer replaces the page and a note re-reads the page too.

## Who stands in the ordered line

Autonomy reads and writes, the replay on attach, the gestures that resolve a question, the plan's spend, the plan verbs, a note, and page reads. Of all of them only the run's summary was a model call or had a budget in seconds. One more finding is not in this PR: the tasks place reads the plan synchronously on the frame. That read is on the loop ledger and has its own cell.

## Measured, real binary, hosted, 2026-09-19

Press on a run's row to its page being drawn. The press is made the moment the row appears, which is when a summary has just been asked, so this is a gesture made during a summary.

| | seconds |
| --- | --- |
| before | 2.4, 5.4, 7.7 |
| before, same drive at trunk 12840eeb5 the same night | 9.9, 10.0, 10.1 |
| after | 0.09, 0.18, 0.18 |

Page reads over twenty seconds of an open page on a running task, in the hosted test: 120 before, 5 after. On an ended task: none after the opening read.

I did not time a note separately. A note takes the same ordered line as the press, so the press is its measure.

## Proof

- Spark, head 1ff247899 on trunk 443c661df, under the one-suite lock: whole `internal/tui3` exit 0, whole `internal/session` exit 0. The head pushed here adds the change entry only.
- Spark, same tree: `internal/guard`, `internal/namelaw`, `internal/manual` ok; `make test-laws` exit 0; `gofmt -l ./cmd ./internal` empty.
- Manual: the follow sentence in `worker-harness.md` now says what the page does.


santoshkumarradha:
Landed on `santos/dev` as c3ab6e03b (tree 77cca97f5). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, on the third run. Runs 1 and 2 each failed one test in a package this PR does not touch (`cmd/codeaf` `TestDoOnTheRunEngineSeatsACheckOnTheCheckModel`, then `internal/session` `TestANewDirectionArrivesWithTheClaimUnspent`); both logs are kept beside the green one as `~/src/trees/check-sdev-c3ab6e03b.run1-flake.log` and `.run2-flake.log`, and both failures are being worked as their own items. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1252, opened 2026-09-19
test(exec): read a position-independent marker in the skill-path shell test

santoshkumarradha:
## What
The internal/exec skill-path shell test asserted on a line parsed from output captured through the capped output collector (`newCappedOutput`). When a command prints more than the preview budget, the rest spills to a file, so the asserted line's POSITION depends on ambient output size. A large environment could push it past the cut and fail the test for a non-product reason (the same class as the tmuxenv fix).

## Change
Test-only. The command prints a short, bounded, position-independent marker line covering exactly what the test asserts, and the test FAILS if the marker line is missing from the captured output, so a future truncation cannot read as a pass. Meaning unchanged.

## Scope
Repo-wide audit of capped-collector assertions; the skill-path shell test was the one genuine at-risk case. internal/exec/tmuxenv_test.go was fenced off (fixed separately in #1249).

## Verification (Spark, off c3ab6e03b)
`go build ./...`; `go test ./internal/exec` (ok); `gofmt -l ./cmd ./internal` (empty); `go run ./cmd/codeaf-changes check`; `go test ./internal/guard ./internal/namelaw` (ok). No em dash on added lines.


santoshkumarradha:
Landed on `santos/dev` as 4bf367378, inside a two-PR stack whose top is eae7055ed (tree 8af63e8f7). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-eae7055ed.log`. This was the first full check to include `build-cross`; it built all six shipped targets in 63 s. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1253, opened 2026-09-19
fix: a run being stopped says so until it has ended

santoshkumarradha:
Draft until #1250 lands; this branch sits on it. Body follows.


santoshkumarradha:
Closing without landing: on the real binary in hosted mode this change alters nothing a person sees.

**The two-arm drive, on the Linux test box, both binaries built with `make build`.** One arm is #1250 alone (`cdd8e777a`), the other is this branch (`004b69f26`). Same drive: a live run under the belt, the stop taken from the conversation 40 seconds in, the screen kept every 50 ms for six seconds.

| | #1250 alone | this branch |
| --- | --- | --- |
| stop card raised | yes | yes |
| conversation line in the first frame | `stopping task 1 (…) — its branch is kept` | the same |
| side column foot during the window | `1 running` | `1 running` |
| status line during the window | `working` | `working` |
| length of the window | 0.30 s | 0.40 s |
| row after the window | `stopped` | `stopped` |
| landed in the folder | nothing | nothing |

A diff of one in-window frame between the arms, with spinners, clocks, prices and hashes normalised, differs only in the tab strip. The frames are kept on the test box under `~/src/doe/peer/stoppingrow-cmp-before` and `stoppingrow-cmp-after`.

**Why.** The branch publishes the run's row as running and stopped by a person the moment the stop is taken. The mark reaches the row, and no drawn word reads it. The surface spells `stopping` for a task in one place, the header of a task's own page for an ordinary task, and a run's row never opens that page: it opens the run's page, which reads the plan store. Everywhere else the engine's own law holds on purpose, that a stop still going through is still running, with the spinner turning. The 7.09 s figure in the entry measured `1 running` leaving the foot, which is the wind-down itself and is what every stopped task has always read. The person is already told at once by the conversation line, and a second stop already answers that the run is stopping.

**The lesson.** The test here asserted a notice, not a screen, so it was green over a change nobody can see.

**What the drive did find**, and what follows as its own change: in that window and after it, the run's own page reads the store, where a stopped root is cancelled, and spells that `incomplete`, while the side column says `running` and then `stopped`.

PR #1254, opened 2026-09-19
build: cross-compile the shipped platforms in make check

santoshkumarradha:
## What
`make check` (vet fmt-check test test-packed-manual size) never cross-compiled, so a platform-specific undefined symbol compiled clean on the host and surfaced only at nightly/release. New `build-cross` target cross-compiles every shipped platform, compile only, and is wired into `check`.

## How
- Reads the platform list straight from `.github/workflows/release.yml` (the release source of truth), so it tracks the shipped set with NO GOOS/GOARCH names hardcoded in the Makefile. Current set: darwin/amd64, darwin/arm64, linux/amd64, linux/arm64, windows/amd64, windows/arm64.
- `CGO_ENABLED=0 GOOS=.. GOARCH=.. go build ./...` per target, compile only (no tests, no packing).
- The six builds share the normal GOCACHE, so a warm box pays little.
- Prints its own duration, rounded up as an upper bound, and names the target it FAILED on (not just a non-zero exit).

## Cost (measured on Spark)
- Warm (shared GOCACHE): ~11s added to make check.
- Cold (empty cache): ~111s (upper bound; a fresh box only).

## Verification (Spark, off c3ab6e03b)
`go build ./...`; `make build-cross` (all six pass, prints duration); `gofmt -l ./cmd ./internal` (empty); `go run ./cmd/codeaf-changes check`; `go test ./internal/guard ./internal/namelaw`. No em dash.


santoshkumarradha:
Landed on `santos/dev` as eae7055ed, inside a two-PR stack whose top is eae7055ed (tree 8af63e8f7). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, log `~/src/trees/check-sdev-eae7055ed.log`. This was the first full check to include `build-cross`; it built all six shipped targets in 63 s. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1255, opened 2026-09-19
fix: no frame reads a run's rows, whatever made the read due

santoshkumarradha:
## What was wrong

The side list's rows of a run were read inside the freshness check that every drawing path comes through. So a frame asked the engine for them. Three things made that read due, and each made it from a frame:

- the first reading of a conversation
- a row of this window's own graph moving, which includes every verb that lands on a run's page
- the run's three second beat

In a hosted conversation that read is a call to another process with a ten second deadline. A frame that waits on it holds every key a person presses. An earlier cell moved the beat's read off the frame and left the other two. The law that keeps engine calls off the update loop never reported any of this, because it could not see a call reached through the surface's own plan reading.

## What changed

- The run's rows are held on the surface and asked for after a message, one read at a time, beside the ordered line. The frame re-files what is held and asks the engine nothing.
- What makes a read due is unchanged. A conversation at rest still reads nothing.
- A verb that lands while a read is out is answered by exactly one more read.
- Opening the run's tab used to read the rows and the run's page on the key press itself. It opens on the held rows and asks for the page through the one door every stored page already arrives through.
- The off-loop law now counts the plan reading as handing the agent back, so every plan call is under the law. The rows read is named in the ledger of reads asked beside the line, with its reason.
- The named budget of calls still made from the loop does not drop. The rows read was never in that budget, because the law could not see it. What changes is that the blind spot is closed.
- The hosted drive script opens the side list with its own key when the copied profile starts a conversation with it hidden. Without that, every side list check fails on the Linux test box on any binary, trunk included.

## Measured, on the Linux test box

The real remote client and a real engine, with 250 ms injected on the run's row read. Each trigger is raised and the very next frame is timed.

| what made the read due | trunk `eae7055ed`, frame | reads inside that frame | this branch, frame | reads inside that frame |
| first reading of the conversation | 258.1 ms | 1 | 0.7 ms | 0 |
| a row of this window's own graph moved | 254.2 ms | 1 | 0.1 ms | 0 |
| the run's beat came due | 256.6 ms | 1 | 0.1 ms | 0 |
| the chat was renamed | 0.2 ms | 0 | 0.2 ms | 0 |

The rows are held in both arms after the message that follows.

`scripts/hosted-drive.sh` against `bin/codeaf` built with `make build` at `27ca2daf5`: ALL PASS, exit 0. Fourteen checks, including the second hand-off showing on the side list, no row still running after the run ends, and a press on the first run's row opening its page.

The same drive against trunk `eae7055ed` passed the side list checks and failed two page checks in its one run, because the page it opened drew no `brief` line on its first screen. I did not chase that; it is one run and it is not this change.

Every drive removed its key copy first, ended the engine it started by socket holder and executable equality, and removed its home. The folders a failed drive leaves for reading were checked for a key copy, held none, and were removed.

## Tests

- New, on the real wire with reads counted: no frame path reads the rows under any of the four triggers, and the message after each costs exactly one read where one is owed.
- New: while a read is out the frame draws what it held, no second read is asked, and a stamp that moved in the meantime costs exactly one further read.
- Tests that called the freshness check and expected fresh rows now fold the read first, the way a window does after a message.
- Two tests that count every command or every call over the wire take their count after the window's first rows read.

## Suites, on the Linux test box at `eaf4ee6bc`

| what | exit |
| --- | --- |
| whole `internal/tui3`, under the box's one-suite lock, 324 s | 0 |
| `internal/manual`, `guard`, `namelaw`, `iconlaw`, `e2e` | 0 |
| `make test-laws` | 0 |
| `gofmt -l ./cmd ./internal` | 0, no lines |
| `go build ./...`, `go vet ./internal/tui3` | 0 |
| `make changelog-check` | 0, 36 entries well formed |

The first whole run failed two tests, both for one reason: a window whose agent can read a run's rows now asks once after its first message, and those two tests count every command or call. They take their count after that first read now. Nothing ran on the laptop.

No manual page changes: nothing a person reads or presses is different, and the three second beat the pages describe is unchanged.


santoshkumarradha:
Landed on `santos/dev` as 7cefdf1f4, inside a two-PR stack whose top is 7cefdf1f4 (tree 020ebd23a). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run on a quiet box, all six shipped targets cross-built in 66 s, log `~/src/trees/check-sdev-7cefdf1f4.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1256, opened 2026-09-19
fix(chatv3): stop and join the standing ticker on process close

santoshkumarradha:
## What
`startStandingTicks` launched a process-global endless ticker via a package `sync.Once`, and `closeAll` never stopped or joined it. After Close returned, the next pass created files under `<home>/v3/standing` (and appended `standing.log`), a writer outliving the process that left `<home>/v3` non-empty (the `<tmp>/v3 directory not empty` leftover class).

## Fix
The ticker is owned by `v3Process`: stop and done channels taken under `p.mu`, a select loop that exits on stop, and `closeAll` calls `stopStandingTicks` (closes stop, waits on done). The package `sync.Once` is removed, so a later process in the same test binary starts its own ticker. Behavior preserved: same interval, silence, at-most-once per process, ErrHeld silence.

## Interaction with #1250's settle-at-stop path (asked at review)
No adverse interaction. `closeAll` calls `stopStandingTicks` AFTER `stopPoolErrands` and BEFORE the agent-close loop where #1250's settle path runs. `stopStandingTicks` takes `p.mu` only briefly to lift the stop/done channels, then waits on `<-done` OUTSIDE the lock, so it neither runs concurrently with nor shares a lock with the main-agent settle path; the standing pass builds a separate standing session, not one of `p.agents`. Worst case, closeAll waits one bounded standing TickWindow for an in-flight pass, which is the intended no-writer-outlives-close guarantee.

## Verification (Spark, off eae7055ed, all via scripts/one-suite.sh for the heavy suite)
- Focused new test `TestCloseAllJoinsStandingTickerAndLaterProcessCanStart` `-count=8`: pass (zero writes under `<home>/v3/standing` after Close; second process starts its ticker).
- Whole `cmd/codeaf` + `internal/standing` `-count=5` under the one-suite lock: ok (459s, 0 FAIL).
- `go build ./...`; `gofmt -l ./cmd ./internal` (empty); `go run ./cmd/codeaf-changes check`; `go test ./internal/guard ./internal/namelaw` (ok). No em dash.

## Flake note
One earlier cmd/codeaf failure occurred on the prior base via a direct `go test` at load ~26 (not through one-suite); its goroutine dump pointed at chat.go:4622, the design-intentional detached worker-leaf flush goroutine, not this change's standing/close path. It did not recur across 5 one-suite runs on eae7055ed, and this change's own close-path tests pass 5x.


santoshkumarradha:
Landed on `santos/dev` as 7ab882003, inside a two-PR stack whose top is 7cefdf1f4 (tree 020ebd23a). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run on a quiet box, all six shipped targets cross-built in 66 s, log `~/src/trees/check-sdev-7cefdf1f4.log`.An earlier run on this PR alone (7ab882003) failed one test in `cmd/codeaf`, `TestAConversationThatFailsHalfOpenClosesItsAgent`, with a temp directory not empty at cleanup, under a box load of about 30; that test opens a conversation through the seam and never reaches the launch road, which is the only caller of the ticker start, and the new stop returns at once when no ticker was started, so the failure is outside this change. Its log is kept as `check-sdev-7ab882003.run1-flake.log`. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1257, opened 2026-09-19
fix: a run's step rows show the work, cut from the commands as they ran

santoshkumarradha:
## What a person saw

Every step a run's worker recorded was drawn on its task page as typed. On a real run most rows began with a change into the run's own copy, and about a third of the rows were the run updating its own record. The work was hard to find among them.

## What changes

- A step row leaves out a leading change into the run's own copy.
- It leaves out any part of the command addressed only to the run's own record, together with whatever that part is piped through.
- A step with nothing else in it has no row. Every other row keeps the number its step ran as, so the numbers can skip and the head's step count stays true.
- Everything kept is drawn exactly as it was typed. A command with nothing left out is drawn whole.
- The recorded command is untouched and is still the record.

## How

- **One reader of command boundaries.** `internal/approval` already owned the reader of where one command ends and the next begins. It now hands out each part with the span it was typed in. The session asks it, and the earlier private splitter is gone.
- **The surface cuts, it never retypes.** Parts cross the wire with their spans, and the surface cuts left-out spans from the recorded line.
- **A part is marked only where cutting is safe.** It must stand in sequence with its neighbours. A part inside a substitution or a group is never marked. A pipeline is one command: it is left out whole when its first part is addressed to the record, and work piped into a record command is never left out.
- **The run's copy is the belt run's own.** A folder directly under the conversation's folder of copies also counts, because an ended run's copy has been given back while its steps still name it. A folder further down is the work. The row's `Folder` keeps the meaning a landed test pins.

## Where this came from

Two cells built most of this and neither brought it home. The first ended on the rig's wall and the second on its checker cap, both with their commits in the run's cut copy. Both sets were recovered by hand. One recovered commit moved a law test in `cmd/codeaf` outside the cell's fence and is dropped.

The hosted review of the recovered work found three faults, fixed here by hand:

- The surface rebuilt a command from trimmed parts. A line with nothing left out drew `ls;go test ./...&&go vet ./...` for a command typed with spaces, and a line ending inside a substitution lost its closing bracket.
- Rows were renumbered to close the gaps, so a page headed `12 steps` drew rows 1 to 6.
- Two shipped manual sentences were reworded and the new section sat under a heading the manual's search does not index. The shipped sentences are restored and there is one `## ` section with two probes.

`scripts/hosted-drive.sh` against `bin/codeaf` built with `make build` at the final head `1ec367172`, which sits on #1255: ALL PASS, exit 0, fourteen checks. The first run's page, as drawn:

```
 steps
 1  pwd && ls -la && echo "---GO FILES---" && find . -name '*.go' -type f | head -50
 2  echo "=== calc.go ===" && cat calc.go && echo "=== go.mod ===" && cat go.mod && …
 3  cat > muldiv.go <<'EOF' …
 4  cat muldiv.go && echo "=== status ===" && git status --short
 7  echo "=== show t-1 ===" && echo "=== show t-2 ==="
```

Steps 5 and 6 were the run's own bookkeeping and have no row. Rows keep their recorded numbers. No row names the run's copy or the record. Every command keeps the spacing it was typed with. On trunk the same page drew `cd <the run's copy> && go test ./... -v` and four rows of record commands.

Every drive removed its key copy first, ended the engine it started by socket holder and executable equality, and removed its home.

**One limit, stated plainly.** The dim line under a row is the head of everything that command printed. Row 7 above had two record commands cut from its middle, and the line under it is what those printed. Cutting a command's output to match is a separate change.

Two other things on these pages are trunk's and not this change, and each has its own cell running: a call the belt refused is drawn as a step with the refusal's sentence under it, and step numbers restart at 1 where a worker is woken again.

## Suites, on the Linux test box

| what | head | exit |
| --- | --- | --- |
| whole `internal/session`, under the one-suite lock, 250 s | `5826d7411` | 0 |
| whole `internal/tui3`, under the one-suite lock | `5826d7411` | 0 |
| `internal/approval`, `remote`, `guard`, `namelaw`, `iconlaw`, `e2e` | `5826d7411` | 0 |
| `internal/manual` | `5826d7411` | FAILED, one probe |
| `internal/manual`, `internal/namelaw` | `1ec367172` | 0 |
| `make test-laws` | `1ec367172` | 0 |
| `make changelog-check`, `gofmt -l ./cmd ./internal` | `1ec367172` | 0, no lines |

The manual failure was mine: the new section's heading took the question "where did the workers go on the task page" away from the page that answers it. The heading and both probes are reworded in `1ec367172`, which changes only the manual page and its test. Nothing ran on the laptop.


santoshkumarradha:
Landed on `santos/dev` as 24e43856f (tree f63261ca6). Full `make check` on Spark on that exact commit: EXIT 0, 114 packages ok, first run, all six shipped targets cross-built in 63 s, log `~/src/trees/check-sdev-24e43856f.log`; `internal/approval` was also run alone on the same tree since this PR touches its command reader. Squashed by hand, so GitHub shows this as closed, not merged; the branch is level with `dev` 9258f7052.

PR #1258, opened 2026-09-19
fix: a task you stopped reads stopped wherever its state word is drawn

santoshkumarradha:
## What a person saw

A run a person stopped read `stopped` on the side list and `incomplete` on its own page and in the tasks place, at the same moment. The side list draws the row the session publishes. The run's page reads the plan store, where a stopped task is cancelled, and the surface spelled failed and cancelled alike as `incomplete`.

## What changes

- A task whose ending is a person's stop reads `stopped` wherever its state word is drawn: the side list, the tasks place, the task's page, and the list of parts on a parent's page.
- Every part that stop ended with it reads `stopped` too. The store ends everything under a cancelled task in the same write and at the same instant, so a part counts when it was cancelled in the very instant an ancestor of it was stopped.
- A part that had already failed, or that the run cancelled for its own reasons at another moment, keeps `incomplete`.
- A stopped task has ended: it is offered no verb and no next step, and it is not read as work that can still move.

## How

- The fact is established where it is known. The session reads it from the store and sends it over the wire as a field of the row. The surface never finds it by reading a sentence.
- The word is decided in one place. The three functions that answer a row's word, its reading and whether it has ended now take the row instead of the store's status string, and every caller goes through them.
- The glyph follows the word the way an ordinary stopped task's does.

## Where this came from

The two-arm hosted drive that closed #1253 found the disagreement. A cell built the field, the one-place word and the drawn-word tests, one of them through the real hosted wire. Review by hand found one fault and filled two gaps:

- The question of whether a task has ended still listed only `done` and `incomplete`, so a task a person had just stopped was offered its stop again. Fixed, with a test.
- Parts a stop took down read `incomplete` under a parent that read `stopped`. Answered as above, with a test against a real store.
- The manual's list of a run row's state words said cancelled reads `incomplete`. That sentence is corrected and `stopped` has its own line.

Same drive on both binaries, both built with `make build`: one `/task` that splits into parts, stopped from the conversation 40 seconds in through the stop card, then the run's page opened by a press on its row on the side list.

| | trunk `eae7055ed` | this branch `de4e3dd01` |
| --- | --- | --- |
| stop card raised | yes | yes |
| `stopped` on the run's page | 0 | 13 |
| `incomplete` on the run's page | 12 | 0 |
| parts under the run | `incomplete · Implement Mul` and so on | `stopped · Add Mul function with tests` and so on |
| landed in the folder | nothing | nothing |

On the final head `d28971eb6`, which sits on #1255, the same drive was run once more: the stopped run's page head reads `stopped · 2 steps · $0.02`, and `incomplete` is nowhere on the page.

Every drive removed its key copy first, ended the engine it started by socket holder and executable equality, and removed its home. No drive folder was left behind.

## Suites, on the Linux test box

| what | head | exit |
| --- | --- | --- |
| whole `internal/session`, under the one-suite lock, 246 s | `7786ad334` | 0 |
| whole `internal/tui3`, under the one-suite lock, 328 s | `7786ad334` | 0 |
| `internal/approval`, `remote`, `manual`, `guard`, `namelaw`, `iconlaw`, `e2e` | `7786ad334` | 0 |
| `make test-laws`, `make changelog-check`, `gofmt -l ./cmd ./internal` | `7786ad334` | 0, no lines |
| `internal/manual`, `internal/namelaw`, `make test-laws`, `make changelog-check` | `d28971eb6` | 0 |

`d28971eb6` adds only the manual sentence and the change entry on top of `7786ad334`. Nothing ran on the laptop.


santoshkumarradha:
Landed on `santos/dev` at 1c213967b as squash 37a9b3b74 (tree 6c57fe9e9). The squash was taken from d28971eb6; the branch was later rebased to 6738d0a25, and the diff between 6738d0a25 and 37a9b3b74 is empty, so what landed is what this PR now shows.

Evidence, all on Spark, fresh clone, exact commit:
- full `make check` on 37a9b3b74: run 1 failed one test in `internal/enginehost` (TestStopEndsAHostEvenWithASurfaceOnIt), a package this change does not touch, log kept as `check-sdev-37a9b3b74.run1-flake.log`; run 2 EXIT 0, 114 packages ok.
- full `make check` on the stack top 1c213967b (tree 8f3c1f092): EXIT 0, 114 packages ok, first run, log `check-sdev-1c213967b.log`.

PR #1259, opened 2026-09-19
session: join the usage writer at Close so no usage row is written after a process quits

santoshkumarradha:
## What changed

A v3 process starts a background usage writer per ledger path (internal/session/usage_ledger.go); it
drained its queue on its own and nothing joined it at Close, so a usage row could still be written after
closeAll returned. This adds CloseUsage, which stops each writer and waits for its run loop to exit, and
calls it from closeAll after the usage flush. RecordUsage, FlushUsage and CloseUsage all resolve, enqueue
to, and close a writer's queue under the one writers lock, so a close cannot race a concurrent record or
flush.

- internal/session/usage_ledger.go: usageWriter gains a stopped channel; usageWriterForLocked and
  CloseUsage; RecordUsage enqueues under usageWritersMu.
- cmd/codeaf/chatv3_process.go: closeAll calls session.CloseUsage() after the usage flush.
- doc.go, engine.go (runRemoteEngine + runEngineHost), exec.go, image.go, chatv3.go: call CloseUsage on
  the paths that record usage outside a v3 process.

## Hold fixes (manager review)

Two defects found in review, both fixed in this branch with failing-first tests
(internal/session/usage_close_ceiling_test.go):

1. CloseUsage kept FlushUsage's bargain. FlushUsage waits under one usageFlushLimit deadline because a
   writer parked inside openUsageLedger on a stalled mount never drains, and an unbounded wait at
   v3Process.closeAll is a hung ~/.codeaf holding the exit open ("a spending record is worth less than the
   exit"). CloseUsage now takes the same single ceiling for the whole call and leaves a still-stuck writer
   behind at that accepted cost. Test TestCloseUsageReturnsUnderTheCeilingWhenAWriterStalls fails on the
   unbounded wait (12s, "the wait has no ceiling") and passes with the ceiling;
   TestCloseUsageJoinsAndLeavesTheLastRowOnDisk keeps the normal case (writer joined, last row on disk).

2. FlushUsage no longer sends on a closed queue. It copied writers under the lock and sent on their queues
   outside it, while CloseUsage closes those queues under the lock, so a flush across a close panicked
   (send on closed channel). FlushUsage now enqueues its markers under usageWritersMu (non-blocking, the
   same door RecordUsage uses) and waits on the accepted markers outside the lock under the ceiling. Test
   TestFlushUsageAndCloseUsageDoNotRace fires flush and close at the same writers 200 times under -race:
   it panics/-races on the old code (panic at usage_ledger.go send site) and is clean with the fix.

The two locked sections that must not hold the lock across a wait (CloseUsage, FlushUsage) are extracted
into detachUsageWriters and enqueueUsageFlush, each with a deferred Unlock, so every usageWritersMu.Lock
is immediately followed by defer Unlock. Lock-defer guard law
(internal/guard TestEveryLockInTheGuardedTreeUnlocksFromADefer, TestLockAllowlistIsStillReal) passes with
no allowlist entry.

FlushUsage callers after this PR: only v3Process.closeAll (immediately before CloseUsage, on the same
goroutine, so the two cannot run against each other) and the exit doors in doc.go/engine.go/exec.go/
image.go/chatv3.go (each a process on its way out; none on a turn path). None can race a concurrent Close
of the same registry, and both now enqueue under the lock regardless.

## Why the usage writer is not part of the CODEAF_HOME re-resolution class

The usage writer HOLDS its path: usageWriterFor(path) captures the path when the writer is created, and
the only caller that resolves the path (RecordUsage via UsageLedgerPath) runs synchronously on a turn,
never on the background goroutine. It does not call home.Join/home.Dir at write time, so unlike the pool
judge sweep (poolrecord.go) and the catalog warm (catalog.go), a surviving usage writer cannot land a row
in a later test's CODEAF_HOME. The defect here is only the missing join; no cross-test leak. Recorded so
this writer is not re-investigated with that class.

## Tests

- internal/session/usage_close_ceiling_test.go: TestCloseUsageReturnsUnderTheCeilingWhenAWriterStalls,
  TestCloseUsageJoinsAndLeavesTheLastRowOnDisk, TestFlushUsageAndCloseUsageDoNotRace.
- cmd/codeaf/usage_close_test.go: TestProcessCloseJoinsTheUsageWriter.
- internal/guard/usagewriter_test.go and internal/session/retiredpin_wire_test.go.

## Acceptance (Spark, one-suite)

Failing-first proof (internal/session, focused -race): the ceiling test fails 12s ("the wait has no
ceiling") on the unbounded wait; the flush/close test panics "send on closed channel" at the send site
under -race. Both green after the fix (3.8s).

`scripts/one-suite.sh go test ./cmd/codeaf ./internal/session ./internal/guard ./internal/namelaw` on
trunk 7cefdf1f4 with the fix:
- cmd/codeaf ok 112.445s
- internal/session ok 246.493s
- internal/guard ok 0.664s (lock-defer law green)
- internal/namelaw ok 3.734s
- exit 0

Head 7083af1a3, tree 8c82cb52f144de1d4d73155c6836c9b837508c22. go build ./... clean, gofmt -l empty,
go run ./cmd/codeaf-changes check exit 0, no em dashes in the added lines.


santoshkumarradha:
Landed on `santos/dev` at 1c213967b (tree 8f3c1f092), squashed from head 7083af1a3, which carries both review fixes: one ceiling over the whole close, and a flush that can never send on a queue a close has closed.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok, then full `make check` on 1c213967b: EXIT 0, 114 packages ok, first run, log `check-sdev-1c213967b.log`.

PR #1260, opened 2026-09-19
tui3, run: a call that did not run is not a step; a refused action is one line

santoshkumarradha:
## What was true

A worker that answered one turn with several tool calls had none of them run, and each was answered by the belt with its own correction. Every one was recorded as a step, and a run's task page drew them as steps: a command a person reads as having run, with the belt's sentence to the worker in dim under it. On the real binary that was rows 2, 3 and 4 of a page, each reading `[not run] no action executed: return exactly one bash tool call per response`.

## What is true now

A step records whether its call ran, and if it did not, which of two things it was. Both facts are set where the event is known and cross the wire as fields. The page reads the fields and never the words of the answer.

- **A correction about the form of a reply has no row.** Several calls in one answer, a call that could not be read, a tool that is not on the belt, a held process rule. Nothing was attempted on the world. It stays in the record and in the head's step count.
- **An action the worker tried and was refused is drawn, as one dim line in the step's place.** `refused`, the word the permissions page already uses, then the command as typed, cut the way every step's command is. It carries no number, because a number on that page is a step that ran. The rows around it keep their recorded numbers. The answer written for the worker is not drawn.

## Where the separating fact is established

`HarnessMade` could not tell the two apart. The new fact is `Event.Refused`: a door refused an action the model attempted. It is set at the one place every pre-action veto passes through, from the veto's own author, so a door added later cannot forget it. The run engine records it on the step beside `NotRun`, and `PlanStep` carries it. A record written before the fields draws as it did.

One reading I made: "not counted as a step that ran" is kept by the line having no number. The head's count stays the record's count, which is also what the step cap counts, so the figures a person reads and the cap agree.

## Tests, failing first

- The drawn page, through the hosted fixture with the real remote client and server: a refused action draws exactly one line, `refused · <command>`, with no number and without the worker's answer. A correction in the same record draws nothing. Rows 1 and 4 keep their numbers and the head reads `4 steps`.
- The event, through the real turn loop on the belt: a two-call response is harness-made and not refused, a push the remote guard refuses is both, a call that ran is neither.
- The record: the run engine writes both facts, and no exit for either.

The manual said a batch had its first call run. It does not. The section is corrected, and a new section in the asker's words answers "a line under steps says refused", with its probe.

## Proof, on Spark

Head 7c9292344 on trunk 1c213967b, tree `~/src/trees/notrun-7c9292344`, every exit code read directly.

| Check | Result |
| --- | --- |
| `go build ./...`, `go vet` of run, session, tui3, manual | exit 0 |
| whole `internal/session` under the suite lock | exit 0, 245 s |
| whole `internal/tui3` under the suite lock | exit 0, 328 s |
| run, plandb, approval, remote, manual, guard, namelaw, iconlaw, e2e (untagged), cmd/codeaf | exit 0 |
| `make test-laws` | exit 0 |
| `gofmt -l ./cmd ./internal` | exit 0, no lines |
| `make build` | exit 0 |

**Hosted, the real binary, two arms, same drive** (`~/src/doe/peer/codeaf-60-notrun/{branch,trunk}`). One `/task` whose brief asks the worker to try reaching a remote once and then do its work. The run's page was opened from the side list and saved as text. Key copy, engine and home removed on exit, and no drive home remains on the box.

Trunk binary 1c213967b, the refused call is a numbered step with the worker's answer under it:

```
 steps
 1  git push origin HEAD 2>&1; echo "EXIT: $?"
    git push is not yours to run: This task's work comes home through its landing, and a pull request is the person's
 2  ls -la
```

Branch binary 7c9292344, one dim line with no number, and the rows after it keep their recorded numbers:

```
 steps
    refused · git push origin HEAD 2>&1
 2  cat calc.go
    package calc
 3  ls *.go && cat calc_test.go 2>/dev/null || echo "no calc_test.go"
```

The branch run's record reads 11 steps, 1 not run, 1 of them refused. The refused step is not drawn with a number and its answer is nowhere on the page. The form-correction case did not occur in either run, since no worker sent several calls in one answer. It is pinned by the drawn-page test and by the event test through the real turn loop.

Built from cell c337's two commits, which set `NotRun` and left the row out. The split, the event fact, the drawn line, the event and record tests, the manual and the entry are by hand.


santoshkumarradha:
Landed on `santos/dev` at 09b50f8eb (tree ba075d3bd), squashed from head 7c9292344 on top of #1263.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (39), build ok for the host and for Windows, then full `make check` on 09b50f8eb: EXIT 0, 114 packages ok, first run, log `check-sdev-09b50f8eb.log`. The author's two-arm hosted drive on the real binary is recorded in the PR body.

PR #1261, opened 2026-09-19
run: a task's step numbers continue across a wake

santoshkumarradha:
## What was true

A run worker numbered its steps from one each time it ran. A task that waited on its children and was launched again recorded a second step 1. Its page drew `1` to `5` and then `1` to `3`, and the live step was called by a number an earlier row already wore.

## What is true now

- A worker starts from the last number the task's record holds. A task's step numbers are counted over the task and never repeat.
- The step cap and the worker's own report still count this run of the worker only. A task that came back has a fresh count against the cap, and its ending line reports the steps of that run. The test pins both: numbers `1 2 3 4 5` in the record, a two-step cap reached after two steps, a report of two.
- The manual says it in the section about a waiting task coming back.

The record's shape is unchanged. A record written before this keeps the numbers it has.

## From the cell

Cell c338, three commits, recovered from the run's cut copy after the rig's cap ended it. Replayed on trunk with one conflict in a test file's imports, resolved by keeping both sides. The manual sentence and the change entry are by hand. The cell's surface test passes on trunk as well, since the page already draws recorded numbers: it pins the drawing, not the fix. The run engine's test is the failing-first one.

## Proof

Head 7948ac8ea on trunk 137bb8cd1, Spark, 2026-09-19, under `umask 022`. Exit codes read directly.

| Check | Result |
| --- | --- |
| whole `internal/tui3` under the suite lock | ok, 329.6 s |
| `make test-laws` | exit 0 |

The change to the engine is in `internal/run`; nothing in `internal/session` moved, so no whole session run was owed. An earlier head failed its build on Spark: a rebase seam had dropped one closing brace in a test file. 7948ac8ea restores it.

The same task text on both binaries, hosted, in a throwaway home: split a task into two parts, wait for both, then verify. The root's worker parks on its children and is woken when they land.

| Arm | Binary | The root task's recorded step numbers |
| --- | --- | --- |
| trunk | 1c213967b | 1, 2, 3, 4, then the wait, then 1, 2 again after the wake |
| branch | built from 7948ac8ea | 1, 2, 3, then the wait, then 4 after the wake |

On the branch the run's page draws the woken worker's row as `4  printf …`. Steps 1 to 3 are commands addressed to the run's record, which have no row, so the page shows the number the step ran as and skips over them, as the manual says. Every child task on the branch numbers from 1 with no repeat.


santoshkumarradha:
Landed on `santos/dev` at d3762f414 (tree 9eeebd250, squashed from head 7948ac8ea on top of #1265).

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on d3762f414: EXIT 0, 115 packages ok, first run, log `check-sdev-d3762f414.log`. The author's two-arm hosted drive on the real binary is in the PR body.

PR #1262, opened 2026-09-19
run: a time limit a person sets ends a run by the same road as the cost limit

santoshkumarradha:
## What was true

A session's elapsed-time limit never reached a run that `/task` started. The flag becomes `session.Budget.Wall`, the interactive check reads it before a new turn, and nothing handed it to the run engine. A run carried a dollar limit and no time limit, so it ran past the hours its person set.

## What is true now

- `RunSpec.Elapsed` carries what is left of the session's wall when the run starts, on the same clock the interactive check uses. `run.Limits.Elapsed` ends the run on it.
- The ending is the cost limit's own: nothing new is launched, the outcome is `a limit you set stopped it`, and what finished lands the ordinary way. A session with no time limit puts none on its runs.
- The one difference is the work in flight. A cost limit lets it finish. A time limit ends it and absorbs each ending, the way the caller's wall already does, so a cut task reads `incomplete` and what it spent is in the run's account.
- A run that ended on a limit with no result of its own says the limit's sentence on its row. It said nothing there before.

## The dollar limit, for comparison

The dollar limit does not lose a worker. It lets in-flight workers finish and absorbs every one, so each is counted. Its hole is a different one: spend is added only when a worker RETURNS, so the ceiling is checked at returns and never during a worker's run. A long worker can carry the run past its ceiling by up to that worker's whole spend, and with several slots by several workers' worth. This change does not touch that. The first version of the time limit here did drop workers: it drained them and set the in-flight count to zero by hand, which left their tasks claimed on a run that was over. That is what the failing assertion on the run's spend and its held workers now pins.

## From the cell, and what I changed by hand

Built from cell c330's commits. By hand: the in-flight endings are absorbed instead of dropped, with a test that fails on the dropped version. I removed the cell's commit that detached the run from its caller's context, because what a closed window or a quit does to a live run is the owner's open call and not this change's. The manual section moved out of the flags section it had split, says which limit is which, and a redundant file-reading test went in favour of the probe.

## Proof, on Spark

Head d57b3c955 on trunk 137bb8cd1, one tree on Spark, 19:40Z to 19:48Z on 2026-09-19, under `umask 022`. Exit codes read directly.

| Check | Result |
| --- | --- |
| whole `internal/session` under the suite lock | ok, 260.5 s |
| `cmd/codeaf` | ok, 96.5 s |
| `make test-laws` | exit 0 |

Seen once at an earlier head and attributed: `TestAWokenTurnPastTheCeilingMovesWorkToATask` timed out in its 60 s wait inside a whole session run while the box also carried three cells, eight busy shells and a second suite. Alone on the same tree it passed 20 of 20 in 1.5 s. It is a test of the old task road, which this change does not touch, and it passed in the run above. A separate cell is reading what that wait depends on.

One drive per binary, hosted, in a throwaway home: `codeaf chat --yolo --max-hours 0.03` (108 s), then one `/task` whose worker writes a file and sits in a single `sleep 400`. The screen was read every five seconds for 260 s.

| Arm | Binary | What the screen did |
| --- | --- | --- |
| branch | built from d57b3c955 | the run ended at 105 s. Drawn: `◆ Write the word started into a new file · incomplete · 1m42s · 1 file · merged` and under it `a fault: a limit you set stopped it · ctrl+o output`. Record: 1 step, ending `the run's wall stopped it`. |
| trunk | 761cdab1e | still `working` at 261 s, more than twice the limit, no ending drawn and none in the record. |

My first try drew only the product's own refusal, because `--max-hours` is refused without `--yolo`. The second try is the one above. The trunk arm's `sleep 400` outlived its engine by about two minutes and ended by itself; nothing was signalled.

## What a worker owes this road

The time limit cancels every worker in flight and then waits on the returns channel for each of them, the same shape as the caller's wall. A worker that ignored its context would therefore hold the run's ending. The one shipped worker observes it: its turn runs under that context and the command it is running is cut with it. The branch arm above is that road on the real binary.

## What reaches a run now, and what does not

- `--max-hours` reaches a run, as what is left of the conversation's wall when the run starts.
- `--max-cost` does NOT reach a run. A run's dollar limit is the conversation's spend ceiling alone, and `--max-cost` is still read only before a new interactive turn. A person who types `--max-cost 2` and starts a run has not bounded that run. That is a following change, together with handing a run what is LEFT of a dollar limit instead of the whole figure.
- #1268, stacked on this, makes the dollar limit hold while a worker is still working.

`--max-cost` still does not reach a run: a run's dollar limit is the conversation's spend ceiling. The manual now says so.


santoshkumarradha:
Landed on `santos/dev` at b71454e31 (tree a071ccd3f), squashed from head d57b3c955 on top of #1269.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (41), build ok for the host and for Windows, then full `make check` on b71454e31: EXIT 0, 115 packages ok, first run, log `check-sdev-b71454e31.log`. The author's two-arm hosted drive on the real binary is in the PR body: under a 108 second limit the run ended at 105 seconds; on the previous head it was still working at 261.

Known and stated in the body: a dollar limit typed on the command line does not reach a run yet; that is a following change.

PR #1263, opened 2026-09-19
test(session): drive the claim-unspent revision through Submit so it cannot lose its ending

santoshkumarradha:
## What changed

TestANewDirectionArrivesWithTheClaimUnspent (internal/session/completion_stale_test.go) queued the
forced revision through the fixture-only helper queueDirection, which appends straight to agent.steering
and, unlike a person's real direction, has no "start a fresh turn if the agent is not running" recovery.
That is why the test intermittently journaled only one ending under load (about 1 in 40 in a gate run at
load ~9): the lifted revision could be dropped at turn teardown instead of earning its own ending.

The test now submits the revision through the real path, agent.Submit(ctx, revision), and collects that
turn's own event stream. Submit either splices the direction into the running turn with a live stream or
starts a new turn when the agent is not running, so the queued revision always gets a turn and an ending,
exactly as a person's direction does in production. The existing assertions are unchanged (queued==true,
admitted==0, at least two journaled ceilings, all dropped:nothing-left).

## Why this is a test fix, not a product defect (settled from the code)

The earlier note that this was "a genuine lost second ending" and this "test-only fix" are reconciled by
reading the teardown, not by -count (which proves nothing here: the old test also passed -count=3000 on an
idle box, the loss was about 1 in 40 only under an artificial a.mu perturbation).

A queued direction earns its second ending at turn teardown, and both paths that could start it are gated
on completed==true:
- agent.go:1759 lifts a spliced direction to the follow-up queue (liftSteersLocked), and it is started
  only if nextFollowUpLocked(completed) returns one at agent.go:1784; nextFollowUpLocked drops every
  follow-up when !completed (agent.go:1909-1910).
- the drained-but-unanswered wake (agent.go:1791) is likewise gated on completed.

completed is runTurn's return (agent.go:1836), and runTurn returns false ONLY on a stopped or errored
turn: ctx interrupt or deadline (loop.go:846, loop.go:943), a permanent mid-stream failure (loop.go:982),
or an abandoned turn (loop.go:1270). A normal checkpoint "nothing left" completion returns true
(loop.go:856). So a person's direction is dropped only when the turn was interrupted, errored, or
abandoned, which is the documented drain law: "a drain must never resurrect a stopped turn"
(nextFollowUpLocked). On a normal completion the queued direction always gets its own turn and ending.

And the person is told when a direction is not taken up on an abnormal end: a spliced steer gets an
explicit EventSteerFellThrough on their stream (steer.go:875) and a journaled appendSteerFellThrough line
(steer.go:882), both from liftSteersLocked (steer.go:856), before the follow-up is dropped, and a plain
direction is drained into the transcript in the person's own words by drainSteeringLocked (agent.go:1760)
regardless of completed. Only the automatic
re-ask is withheld, which is the drain law. So the drop is not silent.

submitUser while the turn is running appends to a.steering (agent.go:971) exactly as queueDirection does,
so it is the SAME teardown that governs both; the difference is only that Submit also subscribes a live
stream (agent.go:976). The old test therefore used a non-production injection (queueDirection) and needed
a perturbation to force the abnormal (completed==false) end that alone loses the ending. Routing the test
through Submit and collecting the revision's own stream makes it ask for exactly what the product promises
(a direction submitted during a normal turn gets an ending), and the >=2-ceilings assertion still fails if
that promise is ever broken. No product code changes; the product question is not left open, because the
only loss path is an interrupted/errored/abandoned turn, which is intended.

## Acceptance (Spark)

- Focused determinism: go test ./internal/session -run '^TestANewDirectionArrivesWithTheClaimUnspent$'
  -count=200 passed (2.0s), no lost ending.
- one-suite full acceptance on 1c213967b: `scripts/one-suite.sh go test ./internal/session ./internal/guard
  ./internal/namelaw` exit 0 (internal/session 247.4s, guard 0.7s, namelaw 3.9s).

Cherry-picked from the diagnosis run's commit onto trunk 7cefdf1f4; one file changed, 7 insertions,
1 deletion. gofmt -l empty, no em dashes, go run ./cmd/codeaf-changes check exit 0.


santoshkumarradha:
Landed on `santos/dev` at 09b50f8eb as squash 8866dda0b (tree 937967bbe, identical to this PR's head tree at b65344025).

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top 09b50f8eb (tree ba075d3bd): EXIT 0, 114 packages ok, first run, log `check-sdev-09b50f8eb.log`.

PR #1264, opened 2026-09-19
scripts: hold the heavy-suite lock so it works across pid namespaces and dies with its holder

santoshkumarradha:
## What changed

scripts/one-suite.sh serialized heavy suites with a per-uid lock directory whose liveness was a pid check
(kill -0 plus a cmdline match). Cells run their test worker in a separate pid namespace over a shared
/tmp, so the pid a cell writes is a namespace-local pid that a host reads as a different process or none,
and the lock is judged stale and taken. Measured on Spark: three heavy suites at once, and, once a cell
rewrote the lock with its namespace-local pid between two host runs, host-versus-host double-took too.
Concurrent heavy suites drove box load into the 20s and manufactured the load-sensitive test reds that
cost gate runs.

The lock is now an advisory file lock held on an open descriptor, so its liveness does not depend on any
process seeing another's pid, and it is released when the holder dies (a suite killed mid-run leaves the
box unlocked, no stale pid bookkeeping):

- cmd/codeaf-suite-lock: a small helper that takes the lock through internal/filelock (the cross-platform
  lock #1248 already ships, unix.Flock on unix and LockFileEx on Windows), holds it on its fd for the
  child's lifetime, writes its pid and start time into the lock file only so a refused contender can name
  the holder, and forwards ordinary stops to the child. Exit status is propagated with the cross-platform
  ExitCode, so the helper compiles and ships on every release target including both Windows builds now in
  the full check's build-cross.
- scripts/one-suite.sh: builds and execs the helper instead of the pid-staleness dance; same refusal
  wording naming the holder, same package scope taking the lock.
- scripts/one-suite_test.sh: drives the real one-suite.sh and asserts refusal in both orders (a namespace
  starter against a host holder and the reverse), and the three-party case, a host holder, a cell starter
  refused, and a host arriving AFTER the cell's attempt STILL refused and still naming the original
  holder, which is the host-versus-host double-take the old lock allowed.

The lock dying with its holder's descriptor is the design point: there is no pid to go stale across a
namespace boundary at all.

## Acceptance (Spark, on the PR head; tree unchanged before and after)

- make build-cross: darwin, linux, windows for amd64 and arm64, all built, 73s, exit 0 (the raw unix.Flock
  helper could not build the Windows targets; internal/filelock does).
- go test ./cmd/codeaf-suite-lock (TestContentionNamesHolderAndDeathUnlocks): ok, the lock file names the
  holder pid, the refusal is exact, and a SIGKILLed holder leaves the next acquirer succeeding.
- bash scripts/one-suite_test.sh: one-suite namespace acceptance: ok (both orders and the host-after-cell
  case, driving the real scripts/one-suite.sh).
- scripts/one-suite.sh go test ./internal/guard via the new helper: ok, a normal run takes and releases.
- gofmt -l ./cmd ./internal empty; go run ./cmd/codeaf-changes check ok; go test ./internal/guard
  ./internal/namelaw ok. No em dashes in the added lines.


santoshkumarradha:
Landed on `santos/dev` at 137bb8cd1 (tree e129b8331), squashed from head a498a3940 on top of #1263 and #1260.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on 137bb8cd1: EXIT 0, 115 packages ok (the lock helper is the new package), first run, log `check-sdev-137bb8cd1.log`. That run took the lock through the new helper. The author's acceptance is in the PR body: all six release targets cross-build, a cell against a host in both orders, and a second host run after a cell's refused attempt.

Note for anyone with an older checkout: the lock file name changed, so a heavy suite started from a tree before this commit and one started after it do not exclude each other. Rebase before running one.

PR #1265, opened 2026-09-19
tui3: the dim line under a step row is only ever the row's own

santoshkumarradha:
## What was true

A step's row leaves out the change into the run's copy and any part addressed to the run's record. Under the row, one dim line drew the head of what came back. That head is cut from the ONE observation the record holds for the whole line. When a left-out part printed first, the dim line showed that part's words under a row that no longer showed the part.

## What is true now

**THE HEAD OF WHAT CAME BACK IS DRAWN UNDER A ROW ONLY WHEN EVERY PART LEFT OUT OF THE ROW CANNOT HAVE WRITTEN TO IT AND STOPS EVERYTHING AFTER IT IF IT FAILS.**

- A left-out part addressed to the run's record withholds the dim line.
- A left-out change into the run's copy keeps it only when it is joined so that its failure ends the line. Joined any other way, what follows runs after a failed change and the head may be that failure's.
- A row with nothing left out keeps it.

The engine establishes the fact with the other display facts and it crosses the wire as a field. The surface never reads the words that came back. The field is the negative on purpose (`ObservationHeadWithheld`), so a step from an engine built before it draws its head as it did.

Nothing can do better from the record as it is. The belt runs the whole line as one shell with its output interleaved, and no part's output has a boundary (diagnosis cell c340).

## How much of a real page this takes away

Counted with the engine's own facts over 25 real recorded task records on Spark, 309 steps, from today's hosted drives and two cells' runs:

| | Steps |
| --- | --- |
| no row at all, as before | 29 |
| rows drawn | 280 |
| rows that keep their dim line | 242 |
| rows that lose it | 38 |

Every one of the 38 lost it to a left-out record part. None lost it to the way a copy change was joined. On the hosted drive for #1260, all 7 rows keep theirs. So about six rows in seven keep their output, and the page has not lost it. Running record parts as their own step in the engine is not needed for this.

## The shipped head assertions: one decision, three places

Three shipped assertions ask for a head under a row. They are one decision with one reason: the head is the first line of the whole answer, so it belongs to the first part that PRINTED, not the first part SHOWN. Where a left-out part can print, the engine cannot say the head is the work's, and it withholds it.

| Test | Row | Head asserted | Now |
| --- | --- | --- | --- |
| `TestTaskPageOmitsOwnRecordIDsAndRunCopyPath` | `ls` beside a left-out record command | `files` | moved to the never list; the step carries the engine's fact |
| `TestHostedPlanPartsCrossTheWireAndFilterThePage` | `printf` beside a left-out record command, over the wire | `recorded-output` | moved to the never list; the test also asserts the fact crossed the wire |
| `TestTaskPageOmitsOwnRecordIDsAndRunCopyPath` | work after a lone leading change into the run copy joined by `&&` | `ok` | unchanged and still passes: that part cannot print and its failure stops the rest |

Each changed test says the reason in its own comment. The hosted one was found by the whole tui3 suite on Spark, not by reading.

## Tests

New drawn-page cases beside it: a record part before the work, after it, and as the head of a pipeline that stands in a sequence, each with an observation that is that part's own print, which appears nowhere on the page. Only the copy change left out keeps the head. Nothing left out keeps it. The engine's fact is pinned over the shipped table of command shapes, with one added: the copy change joined by `;` withholds.

The manual gets its own section, "Why is there no output under a step", with a probe.

## Proof, on Spark, under umask 022

Head eeef11101 on trunk 137bb8cd1, Spark, 2026-09-19, every proof under `umask 022`. Exit codes read directly.

| Check | Result |
| --- | --- |
| whole `internal/session` under the suite lock | ok, 251.3 s, at e0ab9d737. The one commit since touches a single tui3 test file. |
| whole `internal/tui3` under the suite lock | ok, 336.2 s, at eeef11101 |
| `internal/manual`, `internal/guard`, `internal/namelaw` | ok |
| `make test-laws` | exit 0 |

The first whole tui3 run at e0ab9d737 was red on one test, the hosted twin listed above. That is how the third assertion was found.

One drive per binary, hosted, in a throwaway home. The task told its worker to run, as its first command, one line holding a command addressed to the run's record and then `ls`. The run's page was opened from the side list after the run ended.

| Arm | Binary | The first row and what is under it |
| --- | --- | --- |
| trunk | 761cdab1e | `1  ls` and under it `Project overview: 0 tasks`. That line was printed by the left-out record command, drawn as though `ls` had said it. |
| branch | built from eeef11101 | `1  ls` and nothing under it. Rows 2, 3 and 4 keep their dim lines (`package calc`, `ok  calc 0.003s`, `package calc`). |

Cell c344, one commit, which stopped where its brief told it to. The negative field, the joining condition, the manual section and the ruling's test change are by hand.


santoshkumarradha:
Landed on `santos/dev` at d3762f414 (squashed from head eeef11101, under #1261).

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on d3762f414: EXIT 0, 115 packages ok, first run, log `check-sdev-d3762f414.log`. The author's two-arm hosted drive on the real binary is in the PR body.

PR #1266, opened 2026-09-19
run: a worker that has stopped making progress ends, by an observable property

santoshkumarradha:
**In one sentence:** a task whose worker keeps running the same command and getting the same answer is told once what the run saw and what it can do, and is ended three identical steps later instead of burning to the 200 step cap.

## What was true

A run worker's loop ends on a store finish, a store park, four replies in a row with no action, the step cap, the wall, or an errored turn. A reply that carried a tool call resets the no-action count, so a worker that alternates ONE action and ONE text reply never reaches that ending. Only the step cap, 200 finished calls, stopped it. A real pair of workers did about 820 model calls in 30 minutes in exactly that alternation.

## What is true now

**THE SAME COMMAND COMING BACK WITH THE SAME ANSWER, WITH NOTHING MOVING UNDER THE WORKER'S OWN TASK BETWEEN THE LOOKS, IS A WORKER THAT HAS STOPPED MAKING PROGRESS. BEFORE ITS TASK ENDS, THE RUN SAYS SO ONCE.**

- After 3 identical finished steps the run speaks once, in its own voice, inside the turn that is still running. It says what it saw and the three things a worker can do: try something else, wait for a thing outside the plan in one longer action, or park with `plandb wait` when the plan names what it waits on.
- The note states how long one action may run, 600 seconds, read from the belt's own constant and never retyped, so it cannot recommend a wait the belt would cut short.
- 3 more identical steps after the note end the task, `incomplete`, with `the same command came back with the same answer 6 times in a row: the work was not moving`.
- A different action after the note resets everything, the note included. The note records no step and draws no row.
- The property is read from what the run already records about a step: the tool, the command as spelled, the head of the answer, and the store's own record of movement. No word of any command or any error is read, so it holds whatever the tool and whatever the cause.
- The count runs over finished steps, not model calls, so the words a worker says between its actions change nothing.
- Three things reset it: a different command, a different answer, or a move in the worker's own part of the store.
- A worker the store says is blocked on another task is not ended. It is parked through the belt's own wait and wakes when that wait is over.
- The two bounds are two constants, interpolated into every sentence that states them.

## By hand, after the manager's review

- **Movement is read under the worker's own task, not over the whole store.** The cell read the whole store, which kept a broken worker alive for as long as its siblings went on working. It now counts the worker's own row and any row under it, at any depth. What a sibling does is not this worker's progress.
- **A call aborted by the law's own ending is not a step.** A model that answers at once had its next action begun and aborted after the law fired, and the record carried a fifth step under an ending that says four. The new busy-run test found it. The cell's own seat alternated a call and a reply, so its turn had always ended first.

## Cases it does not catch, all bounded by the step cap as before

An A, B, A, B alternation with static answers. A worker that moves its own part of the store on every look. A look that differs only beyond the recorded 2 KiB head.

## The note, from cell c357, with my corrections

A worker waiting on something outside the plan (a build, a remote job) looks with the same command and gets the same answer. That is the most common honest loop there is, and the first version ended it on a bound the worker had never been told about. The note goes in through the session's own steer door, mid turn. A turn that ended between the step and the steer is owed the note at the next round's opening, the road the no-action note already rides, so the sentence is said exactly once either way.

By hand: the manual said a person could "steer a note into" the task; it now uses the steering section's own word, `note`, and says that a note on the task counts as something changing, which the movement reading makes true. Its section lost the asker's word in my first trim and the probe `why did my task stop on its own` reached another page; the word is back and the probe passes.

## Tests, all through the real worker with a scripted seat, no sleeps

- A worker that changes its action after the note is not ended, and the note is in the messages its seat was sent exactly once.
- A worker that ignores the note ends at 6 steps under a cap of 60, with the plain reason, and the note was sent once.
- The note draws no step of its own in the record.
- The alternating shape, the parked road, the growing answer, the own-task movement, the busy sibling and the movement reading itself all still hold, with only the asserted number moved.

The manual's "Why did my task stop on its own?" says what the worker is told and when, and what a person can do, with four probes.

## Proof, on Spark, under umask 022

Head 7849efc78 on trunk da3368d46, tree `~/src/trees/noprogress-7849efc78`, finished 2026-09-19T21:17Z. Exit codes read directly.

| Check | Result |
| --- | --- |
| `make test` of `internal/run` | ok |
| `internal/plandb`, `internal/manual`, `internal/guard`, `internal/namelaw` | ok |
| `cmd/codeaf` | ok |
| `make test-laws` | exit 0 |

The change touches `internal/run` and the manual only, so no whole `internal/session` or `internal/tui3` run was owed.

**No hosted frame, and why.** The only thing a person sees is the ending reason on the task's page, which the page draws from the record's ending line as it draws every other reason. A real model cannot be made to repeat itself on demand, so the ending is asserted on the record through the real worker and a scripted seat.

Cells c343 and c357. The per-worker reading, the aborted-step guard, their tests and the manual corrections are by hand.


santoshkumarradha:
Landed on `santos/dev` at 70b3a3c3b, squashed from head 7849efc78.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (48), build ok for the host and for Windows, then full `make check` on the stack top 70b3a3c3b (tree 3cf4f76e9): EXIT 0, 115 packages ok, first run, log `check-sdev-70b3a3c3b.log`.

PR #1267, opened 2026-09-19
chatv3: stop and join the pool judge sweep at close, and inventory process goroutines with a class law

santoshkumarradha:
chatv3: stop and join the pool judge sweep at close, and inventory process goroutines with a class law

## What changed

Every openV3Launch started guard.Go("pool/judge-sweep", ...) as a bare goroutine, not registered on the
profile errand tracker, so closeAll (which joins only the tracked pool errands) did not wait for it. The
sweep writes under config.ProfilePath(profileDir, "pool"), which resolves at write time to home.Join("pool")
when the profile dir is empty, and home.Join reads CODEAF_HOME at call time. So a sweep that outlived Close
could resolve an empty profile path against a LATER CODEAF_HOME and write pool files into another launch's
home, the late-home-writer class behind the intermittent "TempDir RemoveAll: directory not empty" gate reds,
including in tests that open no process of their own.

- cmd/codeaf/chatv3.go, cmd/codeaf/poolrecord.go: the judge sweep is now owned by the profile pool errand
  set (registered through the errand tracker with a context), so stopPoolErrands in closeAll cancels and
  JOINS it before Close returns. The join is outside the enqueue lock. Per-launch semantics are kept and a
  second launch/process in the same binary starts its own sweep.
- cmd/codeaf/pooljudge_close_test.go: a real-close-path test proving no pool file is created after Close
  returns (fails on base with the bare goroutine), the cross-test variant (a survivor sweep from one
  CODEAF_HOME does not recreate a pool under a second launch's home), and that two processes each sweep.
- cmd/codeaf/chatv3_guard_class_test.go: a class law that inventories every guard.Go a v3 process starts
  and requires each long-lived one to name what stops it and where closeAll joins it. chatv3/models,
  engine/models and chatv3/sweep-home are recorded as KNOWN-OPEN with the cells that own them (catalog warm
  and place-sweep walk), so the law lands green now and shrinks as those land.

## Cancel does not burn a landing (review fix)

poolJudgeLandingContext wrote judgeLast and markJudged whenever no judge scored, INCLUDING the case where a
close cancelled the parent context mid-judge and every candidate failed on that cancellation, not on the run.
That permanently burned the landing so no later start rejudged it. It now returns without writeJudgeLast and
without markJudged when the parent context is done and no score came back, leaving the landing for the next
start. sweepClaimContext no longer counts such a landing (judged++ ran after the call) and leaves the claim
in place, so the next sweep recovers it instead of a completed sweep removing it. The shadowed per-candidate
context inside the loop is out of scope at the check; ctx there is the parent.

Failing-first: TestJudgeLandingLeftUnjudgedWhenCancelledMidJudge cancels the context before the ask, so every
candidate returns the cancel error and no score; on base the landing is markJudged (alreadyJudged true) and
the test fails at pooljudge_close_test.go:189, with the fix it is left unjudged and the test passes.

## What a close waits for, at most

closeAll cancels the sweep's context and joins the errand. The join waits at most for one in-flight judge ask
to observe the cancel and return; each ask is bounded by its per-candidate context, and the folder scan
between candidates is not waited on. So Close returns promptly rather than blocking on a full judge pass.

## Acceptance (Spark, on the PR head 74f26ae3c, rebased on santos/dev 137bb8cd1; tree unchanged before/after)

- scripts/one-suite.sh go test ./cmd/codeaf: ok 99.7s (whole package, via the new kernel-file-lock one-suite
  from #1264, on the landed head).
- go test ./cmd/codeaf -run TestJudgeLandingLeftUnjudgedWhenCancelledMidJudge -count=5: ok; failing-first
  proven by stashing only poolrecord.go (test red at :189), then restoring (green).
- gofmt -l ./cmd ./internal empty; go run ./cmd/codeaf-changes check ok (40 entries well formed); go test


santoshkumarradha:
Landed on `santos/dev` at 761cdab1e (tree 8f0e4f28f, identical to this PR's head tree at 74f26ae3c), which carries the review fix: a landing the close interrupts mid-judge is left unjudged and unmarked, with its claim in place, so the next start judges it.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (40), build ok for the host and for Windows, then full `make check` on 761cdab1e: EXIT 0, 115 packages ok, log `check-sdev-761cdab1e.log`. The run was refused twice by the new heavy-suite lock while another session's suite held it, waited, and then ran once.

PR #1268, opened 2026-09-19
A run's dollar limit holds while a worker is still working

santoshkumarradha:
Stacked on #1262. The diff against that branch is this change alone.

## What was wrong

A run's dollar limit was read only when a worker came home. A worker's spend reached the run's count in its report, so one long worker could pass the limit many times over before anything counted it. #1262's body named this hole; this closes it.

## What changes

- **A worker says what it has spent as each model call is paid for.** The worker reads the agent's running total as each event comes by and tells the run whenever it has risen. The figure is cumulative, so the run adds only what is new and a dollar is counted once.
- **The run adds it while the worker is still working.** When the sum reaches the limit the run launches nothing new, ends the work in flight, absorbs every ending, and answers `a limit you set stopped it`. That is the time limit's road from #1262. Nothing is dropped.
- **A worker never waits on the run to report.** It writes its figure under a lock and leaves a token the loop may miss. A missed token loses nothing, because the next read takes the whole figure.
- **A return still reconciles.** The report is the receipt. A report smaller than what was already counted takes nothing back, because the money was spent.
- A run with no dollar limit counts as it always did, at each return.

The call that reaches the limit is already paid for, so a run can end a little over its figure. The manual says so.

## Two faults the review found in the first version, and the tests that hold them

The first version came from a work cell. Both faults passed its own tests, which used stand-in workers only.

1. **A freeze.** Spend figures travelled over the returns channel with a blocking send. The drain waits for workers and reads nothing, so a worker that reported while the run drained blocked for good and the run never answered. `TestAWorkerReportingSpendWhileTheRunDrainsDoesNotHoldTheRunOpen` finishes the tree while one worker is still out, then has that worker report 64 times after its context ends. On the first version it fails with "the run never answered" after 3 of 64 reports, which is the channel's depth. It passes here.
2. **The real worker never reported.** The report was wired where the turn's usage arrives, which is the turn's end, and a worker's turn is its whole working life. `TestBashWorkerReportsItsSpendWhileItIsStillWorking` drives the real worker through a scripted seat. On the first version it saw one report for three paid calls. It passes here, one rising figure per paid call, the last equal to the report.

The cell's four supervisor tests are kept: one worker alone crossing the limit, two workers crossing it together, a run with no limit, and a worker that returns under the limit counted once.

## A run's spend never goes down

A worker whose task the store cancelled under it used to hand back no spend the run counted. Those were paid dollars missing from the run's figure and from a person's limit. The rule now: what a run counts as spent never goes down, and it includes every paid call whatever way its task ended. The cancelled worker's words and steps are still dropped. `TestSpendOfAWorkerWhoseTaskTheStoreCancelledStaysInTheRunsCount` holds it with a limit and without one. The shipped test that pinned the old figure now expects the cancelled leaf's dollars and none of its steps, and is renamed to say so.

The same rule reaches a worker the run outlives. When the tree finishes while a worker is still out, the drain ends it and used to drop its return whole. The drain now waits for it as before, then reads the returns left in the channel without waiting and settles only their dollars. The shipped cancelled-leaf test takes this road, because there the run ends before the leaf comes home.

## Proof

Head ca32515e4, stacked on #1262's d57b3c955 on trunk 137bb8cd1, Spark, 2026-09-19, under `umask 022`. Exit codes read directly.

| Check | Result |
| --- | --- |
| `go build ./...` and `go vet` | exit 0 |
| `make test` of `internal/run`, `internal/plandb`, `internal/manual`, `internal/guard`, `internal/namelaw`, `cmd/codeaf` | exit 0 |
| the new and changed run tests, 20 times over | ok |
| the same tests under `-race` | ok |
| `make test-laws` | exit 0 |

Failing-first, each on the head before its fix: the drain test on the cell's own supervisor fails with "the run never answered" after 3 of 64 reports; the real-worker test sees one report for three paid calls; the cancelled-spend test reads a run spend of 0. The shipped cancelled-leaf test is what showed that the drain, and not only the absorb, was where a cancelled worker's dollars went missing: there the run ends before the leaf comes home.

The change touches `internal/run` and one manual paragraph, and draws no new word, so no whole `internal/session` or `internal/tui3` run was owed. The ending a person sees is #1262's, whose hosted frame is in that PR.


santoshkumarradha:
Landed on `santos/dev` at da3368d46 (tree b22d6b9de, this PR's own diff over #1262's head ca32515e4 applied on top of #1261; bashworker.go is a clean three-way with #1261 and every other file equals this PR's head).

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on da3368d46: EXIT 0, 115 packages ok, first run, log `check-sdev-da3368d46.log`. The stack top was checked whole, so the three-way file is covered by the same run.

PR #1269, opened 2026-09-19
cmd/codeaf: the model-warm waiters are joined; name the open catalog warm one package down

santoshkumarradha:
cmd/codeaf: the model-warm waiters are joined; name the open catalog warm one package down

## What changed

Follow-up to #1267's guard.Go inventory law. Two rows were stale.

Since #1179 the chatv3/models and engine/models warm waiters (proc.warmModels -> poolErrandGoCtx) are seated
on the profile errand tracker and joined by stopPoolErrands in v3Process.closeAll. The law rows for both are
reclassified from KNOWN-OPEN to joined, naming their stop (the pool errand context) and their closeAll join.

The real remaining survivor is one package down and the law here cannot see it: the internal/catalog
catalog/warm goroutine that catalog.LoadLazy starts with guard.Go. It runs on context.Background() (every
production LoadLazy caller passes it), is cancelled by nothing at close, and calls writeCache(cachePath(...)),
which resolves home.Dir() at warm time when Options.Dir is empty. Each reclassified row names that open writer
and the cell that owns closing it (the catalog-warm-close cell, task/catalog-warm-close). The only known-open
row left in this file is chatv3/sweep-home (owned by the place-sweep walk cell).

Test-only change to cmd/codeaf/chatv3_guard_class_test.go.

## Acceptance (Spark, on the PR head; tree unchanged before/after)

- go test ./cmd/codeaf -run TestV3ProcessGuardGoClass -count=1: ok (the inventory still matches every scope;
  the two rows pass the joined-writer rule naming both stop and join).
- gofmt -l cmd/codeaf/chatv3_guard_class_test.go empty; go run ./cmd/codeaf-changes check ok. No em dashes.


santoshkumarradha:
Landed on `santos/dev` at b71454e31 as the first commit of a two commit stack (test only), with #1262 on top.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top b71454e31 (tree a071ccd3f): EXIT 0, 115 packages ok, first run, log `check-sdev-b71454e31.log`.

PR #1270, opened 2026-09-19
scripts: the heavy-suite lock follows the suite, not the wrapper

santoshkumarradha:
scripts: the heavy-suite lock follows the suite, not the wrapper

## What changed

#1264 made the heavy-suite lock a kernel file lock held on the wrapper's descriptor, released when the
WRAPPER exits. If the wrapper is killed while the go test child keeps running as an orphan, the lock RELEASES
while a heavy suite is still executing, so a second heavy suite starts concurrently, the load defect #1264
set out to end.

- cmd/codeaf-suite-lock/lock_unix.go (//go:build unix): appends the already-locked *os.File to
  exec.Cmd.ExtraFiles, so the child inherits a descriptor for the SAME open file description. The advisory
  lock is held until every fd on that description is closed, so it survives the wrapper's death and stays
  held until the suite exits.
- cmd/codeaf-suite-lock/lock_windows.go (//go:build windows): the build-tagged no-op counterpart, so the
  package still compiles on both Windows targets in build-cross.
- cmd/codeaf-suite-lock/main.go: inheritance is installed before Start, and cmd.Process.Pid is written after
  Start, so an ordinary refusal names the live suite, not the dead wrapper.

## The cost of an inherited descriptor, and an honest refusal

The inherited descriptor is not close-on-exec, so everything the suite starts inherits it: the test binaries,
and any daemon THEY start and leak. Our tests start engine hosts, which close no inherited descriptors on the
way up (internal/enginehost/detach_unix.go only sets SysProcAttr{Setsid: true}), so a leaked host keeps the
lock. This is not theoretical: the learning crew's queue daemon wedged its own restart the same way today,
its flock descriptor inherited by every cell it launched, the cells outliving it and keeping the lock. Their
fix was closing the descriptor in launched children, which this helper cannot do for go test's children.

So the refusal is made honest rather than clever. The pid in the lock file is meaningful only in the HOLDER's
pid namespace: read from another namespace it may be a live suite, a pid that maps to nothing, or an
unrelated process. The refusal therefore NEVER says the holder is gone. When the recorded pid is not visible
from here (pidVisibleHere: kill(pid,0) on unix, a no-op returning true on Windows where the descriptor is not
inherited) it says exactly that and gives BOTH readings, another process namespace or a holder that exited and
left the lock to a process it started, then names lsof on the lock file and that it must be run where every
process is visible. A visible holder still gets the ordinary wait message. The first line is byte for byte
unchanged, so a gate keying on it still matches.

## Acceptance (Spark, rebased on santos/dev da3368d46; tree unchanged before and after)

- Failing-first: the test commit fails on base at main_test.go:63 (metadata named the wrapper pid); with the
  fix, -count=20 passes.
- TestRefusalNamesADeadHolderWhenAGrandchildKeepsTheLock backgrounds a grandchild that inherits the locked
  descriptor and outlives the suite, and asserts the not-visible line and the lsof hint; it is red on the
  prior code, which said only "Wait for it".
- scripts/one-suite_test.sh: one-suite namespace acceptance ok, now also asserting that a contender in a
  fresh pid namespace, which cannot see a host holder's pid, is told the pid is not visible from here rather
  than that a live suite is gone.
- make build-cross: all six targets, exit 0. go test ./cmd/codeaf-suite-lock -count=10 ok. gofmt -l ./cmd
  ./internal empty; go run ./cmd/codeaf-changes check ok; go test ./internal/guard ./internal/namelaw ok.
  No em dashes in the added lines.


santoshkumarradha:
Landed on `santos/dev` at ec9953efb as squash ec9953efb, squashed from head 6ec4c0778.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (46), build ok for the host and for Windows, then full `make check` on the stack top ec9953efb (tree 6604fa1bd): EXIT 0, 115 packages ok, log `check-sdev-ec9953efb.log`. The run was refused once by the heavy-suite lock while another suite held it, waited, and then ran once.

PR #1271, opened 2026-09-19
A parked task is finished only by a worker woken for it

santoshkumarradha:
**In one sentence:** a task that is waiting on other work can no longer be marked finished by a late step of its own worker, which under load used to close a run before its work was checked.

## What was wrong

A worker that parks its task with `plandb wait` could still finish it. A round the worker had already started still ends its tool, and when that tool was the task's own `plandb done` the plan store admitted it: the claim a park releases was never what the store checked for the run's own task.

On the run's own task that closed a run over work nobody had reviewed. The root parked on its child, the child wrote its own done, the root's late finish landed, and the run answered `done` while the child's worker was still out. Five loaded runs in four hundred did this and no quiet run did. Every one left a task that was done and still flagged waiting.

## What changes

One function in `internal/plandb/store.go`, `refuseParked`, called by the three roads that write an ending: a worker's done, a worker's fail, and the run's own completion of its root. A parked task hears `task "<id>" is waiting and can only be finished after it is woken`. `Store.Wake` stays the only place the flag clears. Nothing else changes: `StopRoot` and `Cancel` are untouched, so a person's stop still ends a parked task.

Done together with waiting is now a pair no road in the store can write.

## What the run does with the refusal

Read from `internal/run/run.go`, which this change does not touch. `completeTree` calls `CompleteRoot` and drops its error, as it always has. When the root is parked the call is now refused, the root stays open and parked, and the run does not hang on it: the same pass goes on to `launchWaits`, which wakes a parked task as soon as everything it waited on has ended, clears the flag through `Store.Wake`, and launches its worker again. That worker's finish is an ordinary one. A park cannot repeat forever, because `plandb wait` refuses a wait with nothing open, and the run's own limits bound the rest. The forced test below goes through this road end to end.

## What this does not claim

The store rule alone does not give "no run answers done while a finished piece of work has had no review round". With the late finish refused, the run still wakes the root off the child's store row while the child's worker is out. That is the run's own law about what counts as a landing. It is fixed on `fix/no-run-answers-done-before-a-review`, stacked on this branch, and its forced tests are red against this head. A store that refuses the fingerprint is good by itself.

## Proof

On Spark, under `umask 022`, tree `~/src/trees/parked-c0925b8eb`, head c0925b8eb on santos/dev da3368d46, finished 2026-09-19T21:07Z.

| Step | Result |
| --- | --- |
| `internal/session`, whole, under the box lock | ok, 245 s |
| `make test-laws` | exit 0 |

Failing first: `TestAParkedRootsLateFinishIsRefusedAndTheRunCarriesOn` against trunk's store answers `the parked root's late finish answered ""`, the late finish admitted. The test forces the loaded order with channels and no sleeps: the child's done lands in the store, then the late finish arrives, and only then does the child's worker come home.

`internal/plandb/parked_finish_test.go` holds the store's side: done, fail and the root's completion each refused while parked and each admitted after a wake.

A person sees no new word.


santoshkumarradha:
Landed on `santos/dev` at ec9953efb as squash 0d3f3ed8d, squashed from head c0925b8eb.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (46), build ok for the host and for Windows, then full `make check` on the stack top ec9953efb (tree 6604fa1bd): EXIT 0, 115 packages ok, log `check-sdev-ec9953efb.log`. The run was refused once by the heavy-suite lock while another suite held it, waited, and then ran once. This is the store's half; that no run answers done before a review is #1275.

PR #1272, opened 2026-09-19
enginehost: a readiness gate proves a real accept, and two raw-Dial callers retry the stale-socket window

santoshkumarradha:
enginehost: a readiness gate proves a real accept, and two raw-Dial callers retry the stale-socket window

## What changed

The host readiness gate treated "the socket file exists and the lock is held" as ready. A host's own startup
order (internal/enginehost/host.go) takes the lock, then removes the old socket, then net.Listen. So there is
a window where the lock is held and a stale socket file is present while nothing is listening yet, and a
caller that dialed in that window was refused.

- internal/enginehost/stale_test.go: waitForHostQuietly now confirms a REAL accept by calling
  Ask(workspace, remote.WhoIs{}), so it returns only after Dial plus the remote whois handshake, not on a
  socket pathname plus a held lock. The forced-window regression preserves a closed Unix socket, holds the
  lock, proves readiness stays blocked, then starts a real host and proves it completes. Host startup order
  is unchanged. Failing-first: the pre-fix file+lock gate fails at stale_test.go with "stale socket plus
  held lock was accepted as ready".
- Product callers: both raw enginehost.Dial callers were traced and found exposed to the same
  remove-to-listen window with no preceding Attach to that host:
  - cmd/codeaf/chatv3_local.go (codeaf chat --once road probe): the former raw Dial had no preceding Attach
    and could race the window, silently choosing the in-process road. It now retries only on ECONNREFUSED,
    tolerates ENOENT during the replacement interval, and leaves unrelated errors and initially missing
    sockets unchanged. Regression: chatv3_local_test.go.
  - cmd/codeaf/chatv3_taskowner.go (cross-workspace task-owner open): the launch Attach is to the launch
    workspace only; a task-owner click can target another restarting host with no preceding Attach. It now
    uses the same narrowly bounded retry. Regression: chatv3_taskowner_test.go. Same-workspace launch stays
    protected by its earlier Attach.
  The retry is bounded: at most 26 attempts over ~250 ms, each Dial keeping enginehost's own 2s ceiling.

## Acceptance (Spark, on the PR head, rebased on 761cdab1e; authors are the bot; tree unchanged before/after)

- scripts/one-suite.sh go test ./internal/enginehost: ok 14.3s; scripts/one-suite.sh go test ./cmd/codeaf:
  ok 113.3s.
- Race nets: go test ./internal/enginehost -run TestWaitForHostQuietlyRequiresAListeningHost -count=50 ok;
  go test ./cmd/codeaf -run (the two window regressions) -count=20 ok. Failing-first for each recorded above.
- go build ./... ok; gofmt -l ./cmd ./internal empty; go run ./cmd/codeaf-changes check ok; go test

One known window, named so nobody rediscovers it: the readiness probe takes the host lock for an instant, so a host calling takeLock in that same instant would read it busy. The window is microseconds and it only opens after a dial has already failed, so a launch loses at worst one retry pause.


santoshkumarradha:
Landed on `santos/dev` at ec9953efb as squash 8eb26dc0b, squashed from head 3c3850b95.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (46), build ok for the host and for Windows, then full `make check` on the stack top ec9953efb (tree 6604fa1bd): EXIT 0, 115 packages ok, log `check-sdev-ec9953efb.log`. The run was refused once by the heavy-suite lock while another suite held it, waited, and then ran once.

PR #1273, opened 2026-09-19
approval: a backslash pair inside double quotes is text, read by one scanner under both readers

santoshkumarradha:
**In one sentence:** a check command that carries a backslash inside double quotes, such as a search pattern written `"a\|b"`, is now accepted as the one command it is, and both readers of "is this one command" share one quote scanner.

## What was wrong

`approval.FirstCompositionOutsideQuotes` refused every backslash inside double quotes, whatever it stood before. A declared check of one command, measured on 2026-09-19, was turned away naming the backslash. Its sibling `FirstBarOutsideQuotes`, which finds where a line's first stage ends, had no escape rule at all and read `"a\" | b"` as a quotation that closed at the escaped quote. Two quote loops were a safety argument with two versions.

## The rule, by the shell's quoting property

One scanner, `scanShellQuotes`, under both readers. A reader differs from the other only in which characters it calls live.

- Inside single quotes every character is text, a backslash included.
- Inside double quotes a dollar and a backtick stay live, because the shell still expands them. A backslash there makes the character after it text, whichever character that is, so the pair is stepped over whole. That is what keeps an escaped quote from closing the quotation and a doubled backslash from hiding the quote that does close it.
- **Conservative by construction.** A shape the scanner cannot prove is one whole command is uncertain, and both readers treat uncertain as refused: a quotation left open, a backslash with nothing after it, a backslash before a newline, and every backslash outside quotes, which was always refused and still is.
- The bar reader answers no bar when the shape ahead of any bar is uncertain. The line is then left whole, the uncertainty still in it, and the composition reader refuses it at the door. A cut there would hand the door a shorter command than the one written.

## The table, `internal/approval/composition_test.go`

The first row is a synthetic line of the shape and length that was turned away: one ordinary program and one double-quoted pattern holding escaped bars and an escaped quote, 203 bytes. Admitted: that row, `"a\" ; b"` as one word, `"a\\b"`, a single-quoted `'a\|$()'`, `"a\$b"`, an escaped backtick. Still refused, each naming its character: `"a\\" ; b` (the semicolon), `$(...)` and a backtick inside double quotes, a trailing backslash, a backslash before a newline inside and outside double quotes, `'a\' ; b`, both unclosed quotes, and a backslash outside quotes. The bar reader has the mirror rows.

## Every caller, through its real function

That 203 byte line goes through `auditableDeclaredCheck` in the plan store, `preparedAuditCommand`, `firstStage` and `wholeClause`. `commandLike`, the declared-check list and the audit door built from it stand behind a length law, so they are fed the same shape at 45 bytes.

**A 203 byte check is still refused at those three, for length, and a following change owns that.** This change carries the safety reader and nothing else.

**The shell is the witness.** `TestACheckWithAQuotedBarIsAdmittedRunAndItsExitCodeRecorded` runs checks through the checker's real gate in a real shell and asserts a marker file no second command may create. Its row `grep "a\"; touch RAN; \"" walls.md` sat in the must-refuse list under the old rule. It is one command, so it is now admitted, RUN by that shell, and the marker is asserted absent. `grep "a\\" ; touch RAN`, which is two commands, took its place on the refused side. The same row moved the same way in the plan store's test and in the session's shape test.

## What I did not take from cell c353

- It rewrote the shipped refusal sentence and the checks schema description. Both shipped sentences still name a form that passes, and its schema sentence ran past the 200 byte description law. Both are byte for byte as shipped.
- It raised the declared-check length limit from 200 to 256. That is a different law with a different reason and is out of this change.
- It dropped the reader's law comment. It is back, over the scanner.

The manual changes one sentence, the one that became false: what a backslash inside double quotes does.

## Proof

On Spark, under `umask 022`, head 73b3afabe on santos/dev 70b3a3c3b, finished 2026-09-19T22:06Z. Exit codes read directly. The same steps were green earlier at 28f93b8e7, before the history was collapsed to three commits and rebased.

| Step | Result |
| --- | --- |
| `internal/approval`, `internal/plandb`, `internal/manual`, `internal/guard`, `internal/namelaw` | ok |
| `make test-laws` | exit 0 |

The first whole `internal/session` run, at 851794c8b, was red on exactly one test: the real-shell contract test above, the third shipped row pinning the old rule, which the cell's focused runs never met.

A person sees no new word.


santoshkumarradha:
Landed on `santos/dev` as 057d01f47, squashed from head 73b3afabe.

PR #1274, opened 2026-09-19
catalog: the lazy warm is owned, cancelled and joined at close, so it cannot write under a later home

santoshkumarradha:
catalog: the lazy warm is owned, cancelled and joined at close, so it cannot write under a later home

## What changed

catalog.LoadLazy started its warm as a bare guard.Go on the caller's context, and every production caller
passed context.Background(). Nothing cancelled it and nothing joined it, so after a process closed the warm
could still finish its fetch and call writeCache(cachePath(...)), which resolves home.Dir() at WRITE time
when Options.Dir is empty. That is the late-home writer behind the intermittent "TempDir RemoveAll:
directory not empty" reds: the write lands in whichever CODEAF_HOME the NEXT test has set.

#1179 had already joined the WAITER (warmV3Models gating on models.Warmed(ctx)), which is why the class
looked closed; it never touched the catalog package's own write inside LoadLazy's goroutine.

- internal/catalog/catalog.go: a lazy Catalog now owns a derived cancelable context and a warmDone channel.
  LoadLazy derives the warm context from the caller's and closes warmDone when the warm exits. Catalog.Close
  cancels and then synchronously joins. load() checks ctx.Err() AFTER a successful fetch and BEFORE
  writeCache, so a transport that succeeds LATE, after the close, does not write; the join is what makes it
  airtight, since the check alone still races and the join alone still writes a late success.
- cmd/codeaf/chatv3_process.go, cmd/codeaf/engine.go: the process owns a context it passes to LoadLazy, and
  closeAll cancels it and closes the catalog. The engine's real-remote and hosted exits both route through
  closeEngineProcess. doc.go and image.go deliberately stay on context.Background: each is a one-shot command
  with no long-lived process close to belong to.
- internal/catalog/lazy_test.go: a real close-path test blocks a fetch that eventually SUCCEEDS, proves Close
  waits for it, swaps CODEAF_HOME while it is blocked, releases it, then proves neither the original nor the
  replacement home was written, and that a second launch still warms and caches normally.
- The guard class law is left alone deliberately: an internal/catalog lifecycle assertion proves ownership,
  cancellation, joining and the no-post-close-write guarantee, which a cmd/codeaf AST inventory reaching one
  package down could not.

## Acceptance (Spark, rebased on santos/dev da3368d46; tree unchanged before and after)

- scripts/one-suite.sh go test ./internal/catalog ok; scripts/one-suite.sh go test ./cmd/codeaf ok; the new
  regression at -count=10 ok.
- go build ./... ok; gofmt -l ./cmd ./internal empty; go run ./cmd/codeaf-changes check ok; go test


santoshkumarradha:
Landed on `santos/dev` at 70b3a3c3b, squashed from head ec9961df5.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass (48), build ok for the host and for Windows, then full `make check` on the stack top 70b3a3c3b (tree 3cf4f76e9): EXIT 0, 115 packages ok, first run, log `check-sdev-70b3a3c3b.log`.

PR #1275, opened 2026-09-19
run: no run answers done while a finished piece of work has had no review

santoshkumarradha:
**In one sentence:** a run can no longer report `done` while a finished piece of its work has not been checked, whatever order its workers came home in.

## What was wrong

A worker writes its own done to the plan store and comes home a moment later. Its return is what adds its check. The run read the store row as the landing, so anything that happened in that moment went ahead on work nobody had reviewed:

- **A root that finishes while its child's worker is still out.** The store admits the root's done, because every child it has reads done. The child's check then had to go beneath a done root, the store refused it, and the run dropped the refusal without a word and answered `done`.
- **A root woken off the child's row.** The same moment, the other way round.
- **A return already waiting in the run's channel** when a pass looked at the root's ending.
- **A root that returned an error after storing its own done** was not reviewed at all.

None of this happens on a quiet machine. The parked-root order, the one #1271 closed in the store, was measured in five loaded runs of four hundred; the others are the same moment reached by other roads.

## The law

**NO RUN ANSWERS DONE WHILE A FINISHED PIECE OF WORK HAS HAD NO REVIEW ROUND, WHATEVER THE ORDER OF RETURNS. A LANDING IS A RETURN THE RUN HAS ABSORBED, NOT A STORE ROW.**

- The root's completion, a parent's wake and the run's own answer all wait for the worker of a done task to come home. A failed or cancelled row has no review to seat and counts as landed at once.
- **The wait is bounded by the worker.** It reads its own row when the command it is in ends, so the run waits at most the belt's ceiling on one command, 600 seconds. Before, the run ended over that worker and its review was silently lost. The pass does NOT cancel such a worker: `endCancelledWorkers` carries the shipped law FAILED AND CANCELLED, NEVER DONE for exactly this reason.
- `Store.AddReviewCheck` seats a check beneath its parent whatever the parent wrote about itself meanwhile, and moves every done task above it back to waiting on it, each keeping the result it earned. It was `AddRootCheck`, for a childless done root only. Nothing but a check comes in that way; `AddMany` refuses every other child of a terminal task as before.
- A root reopened like this finished over a landing it was never given, so once the check lands it is owed the wake any parent is owed.
- A review the store will not seat ends the run `incomplete`. It is never skipped.
- A pass absorbs the returns already waiting before it looks at the root, and reads the root's row after that, never before.

## From the cells, and what review found

Cells c348 (queued returns), c349 (the root that returns an error over its own done) and c358 (the landing rule and the store seam). c358 was green on its own white-box tests and NOT against the order it was written for:

- Two tests force the loaded orders through the run's own loop with channels and no sleeps, in `internal/run/review_order_test.go`. With c358 alone the running root still answered `done` with its check sitting `ready`, 13 runs in 50: the pass read the root's row, THEN absorbed the return that reopened it, then judged the stale row. Fixed by reading after.
- c358 edited the landed entry of #1233. Restored; this change has its own.
- The store seam kept a name that was no longer true and had no store test. `internal/plandb/review_check_test.go` holds it: both done ancestors reopen with their words kept, a child that is not a check and a check with no parent are refused, the task between closes by itself when the check lands, the root is the run's to complete.
- My own drain test from #1268 forced the drain with a done worker that lingers until cancelled. Under the landing rule that state no longer exists, so it now drains by the caller's ending. The shipped `TestStartWaitsForAWorkerTheCompletedTreeLeftBehind` has its leaf come home by itself for the same reason, and drain keeps a direct test.

## Proof

On Spark, under `umask 022`, head f125824bd on santos/dev 70b3a3c3b. Exit codes read directly.

| Step | Result |
| --- | --- |
| `internal/run`, `internal/plandb`, `internal/guard`, `internal/namelaw`, `cmd/codeaf` | ok |
| `internal/manual`, `make test-laws`, `make changelog-check` | ok at f125824bd |
| The two forced orders, 50 times each, and `internal/run` under the race detector | ok |

Everything but the manual row ran at 2740aabac, tree `~/src/trees/review-2740aabac`, finished 2026-09-19T21:55Z. That run was red on one manual probe: after the rebase over #1266 my new paragraph pulled `why does my task say not done under it` onto this page. The paragraph was reworded, the page and not the test, and f125824bd differs from 2740aabac by that one paragraph; the manual package, the laws and the entry check were re-run there.

Failing first: at #1271's head, before this change, the running-root order is red, `the run answered done and the child was never reviewed`.

A person sees no new word. The manual's "Who checks a task's work?" says what the run waits for and for how long.


santoshkumarradha:
Landed on `santos/dev` as c44b488af, squashed from head f125824bd.

PR #1276, opened 2026-09-19
chat: closing a process stops its place sweep before it can change a later home

santoshkumarradha:
The once-per-binary place sweep was started as an unowned background goroutine. Closing the chat process neither cancelled nor joined its folder walk, so the walk could outlive close and rename or remove stale session folders under a later value of CODEAF_HOME. In a single binary that opens a process, closes it, and then points CODEAF_HOME somewhere else, the surviving walk is a destructive writer under a home that no longer belongs to it.

What changes:

- internal/session: the sweeps take a context and check it between entries and immediately before each rename or remove, so no destructive operation under the home can begin after cancellation.
- cmd/codeaf: each chat process owns its place sweep. Close cancels the sweep context and joins the walk, so the goroutine is finished before closeAll returns.
- The regression fails on base (the walk lands under the later home) and passes with the fix.

Verification on Spark, whole cmd/codeaf and internal/session through scripts/one-suite.sh, plus internal/guard and internal/namelaw, gofmt and codeaf-changes check.


santoshkumarradha:
Landed on `santos/dev` as cd2bf32a0, squashed from head aceb4b83d.

Evidence, all on Spark, fresh clone, exact commit: five separate whole `cmd/codeaf` runs on aceb4b83d, five passes, a full log kept for each; then gofmt clean, change entries pass, build ok for the host and for Windows, and full `make check` on cd2bf32a0 (tree 6e8c7f290), gated alone on top of the previous landing: EXIT 0, 115 packages ok, first run, log `check-sdev-cd2bf32a0.log`. The record also keeps one unnamed red on the earlier head 244ca878c that did not come back.

PR #1278, opened 2026-09-19
session: a woken turn's fixture holds its script like every other turn's

santoshkumarradha:
**In one sentence:** a test of the turn the chat wakes by itself stopped failing about once in forty runs on a busy machine; nothing a person sees changes.

## What was wrong

`TestAWokenTurnPastTheCeilingMovesToAFreshContext` was red about once in forty runs on a loaded machine and never on a quiet one. Cell c351 ran it under deliberate load and gave the verdict: the test's fault, not the product's.

The checkpoint fixtures pin the order of a reading and an answer through the turn's context. A submitted turn inherits that context from its caller. A turn the agent wakes by itself has no caller and started under the plain background, so the fixture could reach every turn but that one, and a late reading raced the answer.

## The change

One unexported field on the agent, `wokenTurnBase`. A fixture sets it so a woken turn starts under the watched context. It is nil in the running product, which starts a woken turn under the background exactly as before, and its comment says so.

**Why a seam and not a lifetime context.** Closing a session already ends a woken turn the way it ends any turn: every turn gets its own cancel when it starts, and `Close` takes that cancel and waits for the turn. The agent has no lifetime context; its memory pass, its naming errand, its wall reader and its turn each have a stop of their own. Deriving woken turns from a new lifetime context would be a larger change than this fix deserves. The manager accepted the one-field seam on that reading.

## Tests

- `TestAWokenTurnHandsOverHoweverLateItsReadingLands`: a completer that lands the sketch reading 50 ms late. With the fixture's line removed it fails after 60 s; with it, it passes.
- Both woken-turn tests, 50 times over: ok.
- The law `TestTheRunningProductNeverWaitsOnAReadingsLanding` still holds: the product calls nothing that waits on a reading.

## Proof

On Spark, under `umask 022`, head a570959de on santos/dev 70b3a3c3b. Exit codes read directly.

| Step | Result |
| --- | --- |
| `internal/manual`, `internal/guard`, `internal/namelaw` | ok |
| `make test-laws` | exit 0 |

Finished 2026-09-19T22:12Z.


santoshkumarradha:
Landed on `santos/dev` as 8832349c9, squashed from head a570959de.

PR #1279, opened 2026-09-19
run, session: an ending a person caused is not a fault, and it names the limit that caused it

santoshkumarradha:
**In one sentence:** a task a limit ended now says which limit it was, `a time limit you set stopped it` or `a dollar limit you set stopped it`, and no longer calls it a fault.

## What was wrong

A run its elapsed-time limit or its dollar limit ended drew its reason as `a fault: a limit you set stopped it`. A limit a person set is not a fault, and a person who set both could not tell which one fired.

## What changed

- **The run records which limit ended it**, where it decides it, beside its one outcome word. The outcome word is unchanged: such a run still reads `incomplete`. The fact crosses the engine seam one for one.
- **The reason line names the limit** and carries no `a fault:` in front. Every other ending keeps its reading.
- **A joined row the run's ending took down mid-flight** drew a fault of its own. It now draws the run's ending.
- **The limit named is the first one reached.** A worker the time limit cut can come home with a receipt past the dollar limit; that return no longer renames the ending.

## Found in review and fixed by hand

The cell counted a part a person stopped alone as cut by a limit that ended the run later. A limit writes no store row, so a cancelled row under a standing root is a person's stop. The run now reads the root's store row to tell the two apart, which does not depend on the order things arrive in. `TestAPartStoppedAloneIsNotCutByTheLimitThatEndsTheRunLater` pins it.

## Tests

- Through the real `Supervisor`: the time limit, the dollar limit, the first limit reached, the rows an ending cut, the part stopped alone.
- Through the session's real door: each limit's reason line, spelled exactly, with no fault prefix; a joined row drawing the run's ending.

## On the real binary, hosted

Four hosted drives on Spark, real binary, real model, one `/task` whose worker sits in a long command. The branch arm is the binary built from this change's tree; the trunk arm is built from cd2bf32a0. The landing card's reason line, as drawn:

| Limit that ended the run | Trunk | This change |
| --- | --- | --- |
| elapsed time (`--max-hours`, 108 s) | `a fault: a limit you set stopped it` | `a time limit you set stopped it` |
| dollars (the conversation's spend limit, set very small) | `a fault: a limit you set stopped it` | `a dollar limit you set stopped it` |

The row above it reads `incomplete` in all four. Each drive removed its key copy, ended the engine it started and removed its home on exit.

## Proof

On Spark, under `umask 022`, head 3e8b41e9d on santos/dev 5240a37b4. Exit codes read directly.

| Step | Result |
| --- | --- |
| `internal/run`, `internal/plandb`, `internal/manual`, `internal/guard`, `internal/namelaw`, `cmd/codeaf` | ok |
| `make test-laws` | exit 0 |

Finished 2026-09-19T23:07Z. The first whole run, at a3df035e2, was red on exactly one test: the complexity law, because the cell gave each limit a case of its own in `TaskReasonOf` and took it to sixteen decisions against a ceiling of fifteen. The two limits are now one arm with a helper. The hosted frames were taken on a binary one commit earlier; that commit changes no word and no behaviour.

The manual's two sentences on a limit's ending are updated in the same change. Cell c360. The words were agreed with the manager before the cell wrote them.


santoshkumarradha:
Landed on `santos/dev` as 2a04d63a0, squashed from head 3e8b41e9d.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top b293a7206 (tree 4fa71dd69): EXIT 0, 115 packages ok, first run, log `check-sdev-b293a7206.log`. The check ran under this stack's own suite lock wrapper, and the lock was free the moment the check ended.

PR #1280, opened 2026-09-19
session, run: a run's dollars reach the conversation's own total

santoshkumarradha:
**In one sentence:** what a task spends now counts toward the conversation's own total while it is still spending it, so your spend limit, `--max-cost`, `/cost` and the status line no longer leave task runs out.

## What was wrong

A task run under the bash belt spent money the conversation never counted. Its workers are agents of their own. Each banked its paid calls on its own account and wrote them to the usage ledger on disk, and the road that drives a run to its landing added nothing home, on any ending. A conversation with a five dollar limit could start any number of five dollar runs, one after another. The two older roads, a task node and an adaptive run, always folded theirs.

Found while reviewing the dollar-remainder cell, which had taken the saved recipe's spend test for this road.

## The law

**A RUN IS THE PERSON'S MONEY AND IT REACHES THE CONVERSATION'S OWN TOTAL, EXACTLY ONCE, WHATEVER WAY THE RUN ENDED.** What the conversation counts as spent never goes down, and a limit on the conversation counts what a run still working has spent so far.

- The run hands its owner its reconciled cumulative total whenever it rises, through one callback on the run's spec. It listens to its workers' live figures whether or not the run itself was handed a dollar limit.
- The session keeps one rising figure per run and adds only the difference each time. The run's own final total closes any gap, before the road splits for a person's stop, so done, incomplete, a limit, a person's stop and an error all count.
- It goes through the fold door that writes no usage ledger line, because each worker already wrote its own calls there. The day's figure on disk is unchanged.

## Tests

- Through the real door, `StartTask` under the bash belt: a finished run, an incomplete run and a run a person stopped each leave their dollars in the conversation's total once. Each was written first and seen red at zero.
- The turn after a run that spent past the conversation's limit is refused with the limit's own shipped sentence, unchanged.
- By hand, through the real `Supervisor` with scripted workers: a run with no dollar limit of its own still tells its owner the live figure before the worker comes home, the figure never goes down, and its last word is the run's own total. The cell had proven the run side only through the session's engine double.

By hand in the manual: the cell rewrote a shipped sentence about adaptive runs that was still true. It is restored; only the task sentence changed.

## Proof

On Spark, under `umask 022`, head b92890341 on santos/dev 70b3a3c3b. Exit codes read directly.

| Step | Result |
| --- | --- |
| `internal/run`, `internal/plandb`, `internal/manual`, `internal/guard`, `internal/namelaw`, `cmd/codeaf` | ok |
| `make test-laws` | exit 0 |

Finished 2026-09-19T22:21Z.

A person sees no new word. They see bigger, truer figures.

Cell c362. The dollar-remainder change stacks on this one.


santoshkumarradha:
Landed on `santos/dev` as 5240a37b4, squashed from head b92890341.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, vet clean on the touched packages, then full `make check` on 5240a37b4 (tree 39f73f9b1), gated alone on top of the previous landing: EXIT 0, 115 packages ok, first run, log `check-sdev-5240a37b4.log`.

PR #1281, opened 2026-09-19
session: a run is handed what is left of every dollar limit

santoshkumarradha:
**In one sentence:** a task run is now allowed only what is left of the dollar limit you set, and `--max-cost` bounds it too; before, every run was handed the conversation's whole spend limit afresh.

## What was wrong

A run's dollar limit was the conversation's spend limit, whole, however much the conversation had already spent. A conversation that had spent four of its five dollars could start a run allowed five more. `--max-cost` did not bound a run at all, and the manual said so.

This only became fixable with #1280: until a run's dollars reached the conversation's own total, there was no true figure to subtract.

## What changed

- A run is handed **what is left of the smaller dollar limit the person set**: the conversation's spend limit and `--max-cost` when codeaf was started with one, less what the conversation has spent so far.
- **Which limit is the smaller is decided in one place**, `railCap`, the same function an adaptive run already reads. The cell had computed it a second time; review replaced that with the one call, so the two roads cannot drift.
- **A limit already spent hands the run the smallest positive figure, not zero**, because zero means no limit to the run engine. Such a run ends on its first paid call with `a dollar limit you set stopped it`.

## Tests

- Through the real door, `StartTask` under the bash belt: a conversation with a $9 spend limit, a $7 launch limit and $2.50 spent hands its run $4.50.
- Through the real `Supervisor` and a real worker: a run handed a remainder ends on it.
- The table for the remainder itself: no limit, nothing spent, part spent, all spent, overspent.

## The manual

Two passages on a run's dollar limit change, each keeping its shipped sentence and adding the one new fact. The old sentence saying the limit on a run is "not `--max-cost`" is gone. By hand in review: the cell had deleted the shipped pointer to where the spend limit is set; it is restored.

## Proof

On Spark, under `umask 022`, head ed9159a40 on santos/dev b293a7206. Exit codes read directly.

| Step | Head | Result |
| --- | --- | --- |
| `go build ./...`, `go vet` | f4b1a12d3 | exit 0 |
| `internal/session`, whole, under the box lock | f4b1a12d3 | ok |
| `internal/run`, `internal/plandb`, `internal/guard`, `internal/namelaw`, `cmd/codeaf` | f4b1a12d3 | ok |
| `gofmt -l ./cmd ./internal`, `make build-cross` | f4b1a12d3 | empty, exit 0 |
| `internal/manual` | f4b1a12d3 | **red**, two probes |
| `internal/manual`, `make test-packed-manual` | ed9159a40 | ok |
| `make test-laws`, `make changelog-check` | ed9159a40 | exit 0 |

The whole proof ran at f4b1a12d3 and was red on the manual alone: the cell's rewrite took "how much will a task cost me" away from the harness page and "how do I set a spending limit" away from the cost page. The pages were fixed, not the test: both passages went back to their shipped sentences with the one new fact added. ed9159a40 differs from f4b1a12d3 in those two Markdown files only, so the Go proof stands. Finished 2026-09-19T23:33Z.

No new word on screen. Cell c359.


santoshkumarradha:
Landed on `santos/dev` as 3a154ceaa, squashed from head ed9159a40.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top 3a154ceaa (tree ce39beaa5): EXIT 0, 115 packages ok, first run, log `check-sdev-3a154ceaa.log`.

PR #1282, opened 2026-09-19
exec/bare: the arm that ends a cancelled call's wait ends its command

santoshkumarradha:
**In one sentence:** a command cancelled in the same instant it started could keep running after its call answered `Command aborted`; the call now ends it itself.

## What was wrong

A foreground bash call waits on three things: the command's exit, the cancellation, and a handover to a background job. The cancellation arm returned at once and its comment said a second goroutine, the cancel watcher, had already sent the kill by then. Nothing ordered that. Returning closes the watcher's own stop channel, and a watcher that had not reached its wait yet then finds two reasons to wake, is handed one at random, and on the wrong one sends nothing.

The window is a cancel that coincides with a start: a run whose dollar limit is reached by the very call that issued the command, a time limit that fires as a command begins, a context already done. The command is started as the leader of a session of its own, so it lived on under no parent, in a folder that was then removed.

Seen once on the real binary: a hosted run ended `incomplete` on its dollar limit, and its command's shell was still running three minutes later under no parent. It ended by itself; nothing was signalled.

## The law

**THE ARM THAT ENDS THE WAIT ENDS THE COMMAND.** The cancellation arm sends the group kill itself. Two exceptions, both stated beside it:

- A call that was handed on as a background job is not killed and answers with its handover, whichever of two ready cases the wait was given.
- A process already reaped is not signalled, because its pid is free to be somebody else's.

The watcher stays for the stretch before the wait is reached. A second kill to a group that is gone changes nothing.

## Tests

Both find the command's own process through the promoter's door and assert the process is gone, not only that the call came back.

- `TestACommandStartedUnderAContextAlreadyDoneIsEnded`: 200 rounds. **Red at round 1 without the fix**, with the command still running; green 600 rounds with it.
- `TestACommandCancelledWhileItRunsIsEnded`: the ordinary stop, mid-command.

## What this does not claim

Shells that tests of a time limit left looping on a shared box have the same shape. They were not reproduced: 100 quiet runs of that test at trunk left nothing, and eight hosted dollar-limit drives, four on trunk and four on this change, left nothing either. The change entry says the same.

An engine ended by a signal in the middle of a command leaves that command running too. That is a separate fact and a product decision, not part of this change.

## Proof

On Spark, under `umask 022`, head 8ba2160f4 on santos/dev b293a7206. Exit codes read directly.

| Step | Result |
| --- | --- |
| `internal/exec`, `internal/exec/bare`, whole, through `make test` | ok |
| `internal/run`, `internal/processgroup`, `internal/manual`, `internal/guard`, `internal/namelaw` | ok |
| `make test-laws` | exit 0 |

Finished 2026-09-19T23:24Z.

Nothing a person reads changed. The manual already says a stopped command is stopped.


santoshkumarradha:
Landed on `santos/dev` as bbdd7cdc3, squashed from head 8ba2160f4.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top 3a154ceaa (tree ce39beaa5): EXIT 0, 115 packages ok, first run, log `check-sdev-3a154ceaa.log`.

santoshkumarradha:
Three shells that this fix would have prevented, found on a shared box and recorded here because they date it.

While cleaning stale processes on a benchmark machine on 2026-09-21 I found three survivors of `TestBashWorkerClearsTheLiveStepWhenTheWallStopsIt`, each holding a `/bin/bash -c export PATH=/tmp/TestBashWorkerClearsTheLiveStepWhenTheWallStopsIt<N>/002/bin:$PATH ...` command line whose temp directory had long since been removed:

| age when found | created | CPU burned |
| --- | --- | --- |
| 3.2 days | 2026-09-18 ~12:00 | 42m45s |
| 3.0 days | 2026-09-18 ~16:00 | 40m24s |
| 2.9 days | 2026-09-18 ~19:00 | 38m41s |

**All three predate this commit** (2026-09-19 19:34), so they are consistent with the fix having worked rather than with an open defect.

The command the test issues is `while [ ! -f <workspace>/release ]; do sleep 0.02; done`. The test never creates `release` and `t.TempDir()` removes the directory when it ends, so a survivor can never exit and spins at roughly 50 iterations a second — which is where ~40 minutes of CPU each comes from.

I could not reproduce it on the current tree: ten consecutive runs of that test on `santos/dev2` leaked nothing. That matches this PR's own scoping, which already says these shells "were NOT reproduced in a hundred quiet runs, so this change does not claim them". No new issue filed; this is only the dated evidence, which was not available when that sentence was written.

One thing worth separating from the above: `bbdd7cdc3` is not an ancestor of `origin/dev` or `origin/main` — it is on the `santos/dev2` line only, while `dev` does carry the harness files. Whatever the intended promotion path is, this fix has not travelled it.

PR #1283, opened 2026-09-19
session: a check refused for its length says so, and says how long it is

santoshkumarradha:
A declared check longer than the limit was refused as though it were not one rerunnable command, even when it was one real command: the refusal named no cause, because it only explains a composition character or a leading cd. The same text was then treated two ways, refused by the proposal door and silently dropped by the audit door, so a checker could be told its own declared check was not on the door.

What changes:

- The refusal names length and says how long the check actually is, keeping its shipped prefix byte for byte: `Invalid arguments: checks must each be ONE rerunnable command: this check is 1001 bytes and a check may be at most 1000`. A person shortens the check that was too long instead of guessing which one it was.
- One number stopped doing two jobs. `auditCommandClipLimit` (200) keeps its stated job, clipping a refused command so a refusal stays readable. `declaredCheckByteLimit` (1000) is admission only, and its reason is stated as the judgement it is: the longest check met in real use runs a little over 200 bytes, a thousand is five times that, and what it guards against is a program pasted in where a check belongs. Its cost is said plainly, at most sixteen declarations in a check's prompt, so at most sixteen thousand bytes.
- The two doors agree. `runnableChecks` applies command shape without the admission length, so a check already recorded in a contract stays on the audit door instead of being silently erased; `declaredCheckList` still refuses a new overlong one. A boundary test pins 1000 admitted and 1001 refused at both doors.
- The pinned refusal bytes and the manual quotation follow the new words. `checksSchemaJSON`, `internal/approval` and #1273's rows are untouched.

Verification on Spark: whole `internal/session` through `scripts/one-suite.sh`, both prompt prefix budget tests, the three length tests, `go build ./...`, `gofmt -l ./cmd ./internal` empty, `go run ./cmd/codeaf-changes check`, and `go test ./internal/guard ./internal/namelaw`.


santoshkumarradha:
Landed on `santos/dev` as c37f46738, squashed from head 7199cbf0b.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top cb3f0f356 (tree 56fb6e9c7): EXIT 0, 115 packages ok, first run, log `check-sdev-cb3f0f356.log`.

PR #1284, opened 2026-09-19
scripts: the heavy-suite lock is held beside the suite, not by what the suite leaves behind

santoshkumarradha:
The locked descriptor was handed to the heavy suite itself. An advisory lock lives on the open file description, and an inherited descriptor is not close-on-exec, so every process the suite left behind kept the whole box locked. That is not hypothetical: on 2026-09-19 an orphaned test shell, with its working directory already deleted, held this lock for eight minutes after a green run, and the next gate waited on a process that was running no suite at all.

What changes:

- The suite is started with no lock descriptor. The locked descriptor goes to a holder started beside it, in its own session, which execs nothing and so can pass the lock to nobody. The holder watches the suite's pid and exits when it is gone. Watching a pid is honest here in a way reading one from a file is not: the holder was started beside the suite, by the wrapper that started it, in the same pid namespace. The recorded start time from the kernel guards the one remaining misreading, a pid reused after the suite exits.
- The wrapper drops its own copy once the holder has it, so the lock is the suite's rather than the wrapper's: a wrapper killed mid-run still does not unlock a box that is running a suite, which is what #1270 set out to fix and this keeps.
- The lock file now records the suite's pid, when it started, and the holder's pid.
- After the suite ends, the wrapper checks the lock and, if it is still held after two seconds, says so and names `lsof`. It ends nothing: the one thing worse than a locked box is a process killed by someone who has not identified it.

If the holder itself is killed, the lock frees while the suite may still be running. That is the failure this chooses, and it is worth stating plainly. The holder waits and does nothing else, so it dies only to a deliberate kill, an out-of-memory sweep, or the box going down; it lives in its own session, so a signal aimed at the suite or at the wrapper misses it; the lock file names both pids, so a free lock under a running suite can still be read back to whoever dropped it; and a lock that nothing can release stops the box, where a lock freed early costs one concurrent suite and a red nobody caused. The kernel makes the same trade when a lock holder dies.

Two shipped tests changed, and both are consequences rather than tidying:

- `TestRefusalNamesADeadHolderWhenAGrandchildKeepsTheLock` described a situation that can no longer arise, since a grandchild never receives the lock. It is now `TestRefusalNamesAPidItCannotSeeWithoutCallingItGone`, which proves the same refusal wording on the case that wording is really for, a holder in another pid namespace, and asserts the output never calls a holder it cannot see gone.
- The release is asynchronous now, within the holder's 50ms poll, so the killed-wrapper regression waits a bounded two seconds for the lock instead of asking once.

Failing-first: the new `TestTheLockIsFreeWhenTheSuiteEndsThoughItLeftAProcessBehind` fails on base with the incident in miniature, the contender refused by a suite that has already exited, and passes with the holder.

Verification on Spark, tree 7a4691643 unchanged before and after: `go test ./cmd/codeaf-suite-lock` at `-count=10` ok; `make build-cross` green on all six targets; `bash scripts/one-suite_test.sh` ok, including both namespace orders and the host-B case, with the script taught to read the lock file's fields rather than split the line in two; `gofmt -l ./cmd ./internal` empty; `go run ./cmd/codeaf-changes check` ok; `go test ./internal/guard ./internal/namelaw` ok.


santoshkumarradha:
Landed on `santos/dev` as b293a7206, squashed from head 5d2953e9d.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top b293a7206 (tree 4fa71dd69): EXIT 0, 115 packages ok, first run, log `check-sdev-b293a7206.log`. The check ran under this stack's own suite lock wrapper, and the lock was free the moment the check ended.

PR #1285, opened 2026-09-19
session: a usage flush says whether it drained, and one writer can be stopped and joined

santoshkumarradha:
FlushUsage waits under a ceiling and returns the same way whether every row reached the disk or the deadline fired first, so the caller cannot tell a join from a timeout. That is how an eighth test joined the late-home-write class tonight: internal/session TestRecordingUsageNeverWaitsOnTheDisk failed with `TempDir RemoveAll cleanup: unlinkat .../001: directory not empty`, and the leftover named itself, a regular usage.jsonl of 7446 bytes where the test had made a fifo. The test read a flush that had run out of time as a flush that was finished, returned, and its temp home was removed while rows were still queued; the writer's next append recreated the directory (openUsageLedger does MkdirAll, then opens with O_CREATE|O_APPEND) and wrote a regular file into it.

What changes:

- FlushUsage returns true when everything in front of it reached the disk and false when the ceiling fired. The wait itself is unchanged and deliberate: a stalled ledger must not hold a terminal open.
- StopUsageWriter(path) ends one path's writer and waits for it, the single-path form of the close the process already performs (#1259's CloseUsage), sharing its rule that the queue is closed under the writers lock so a concurrent RecordUsage can never send on a closing queue.
- usageFlushLimit becomes a var so a test can shrink the deadline in one line rather than putting load on the box to provoke it.
- The shipped test defers the stop and fails if the flush reports a deadline rather than a drain.

WHAT THE JOIN IS WORTH, PLAINLY. StopUsageWriter is bounded by usageFlushLimit, the same ceiling FlushUsage and CloseUsage accept, and it answers whether the writer really finished. Where it returns true, no row can land afterwards, which is the shipped test's case because its pipe is fully drained before the stop. Where it returns false the writer is parked inside its open or its write, and detaching closes the queue while run() ranges that queue to completion, so the rows ALREADY queued still land when it unblocks. Stopping prevents new rows and tells the truth; it does not cancel a backlog. Closing that gap would mean a writer that drops a backlog it can no longer write, which is a separate decision about a person's ledger and not this change.

Honest standing of the evidence: the original red was a red only in company, on a loaded box, and I could not reproduce it deterministically. Shrinking the deadline in the shipped test does not reproduce it, because the test's own drain goroutine keeps the writer fed. So this rests on the reading above plus two seam tests, not on watching the original red turn green: a parked writer's flush reports false rather than claiming a drain, and a stopped path leaves the registry so no later row can queue for a home that is going away.

Not in this change: the writer still recreates a ledger directory that has been removed underneath it, which is what turns a lost row into a rebuilt test tree. That is on the money path and is being put to the owner separately.

Verification on Spark, light and lock-free: go build ./...; the two usage tests at -count=3; gofmt -l ./cmd ./internal empty; go run ./cmd/codeaf-changes check. Whole internal/session runs when the lock frees and the head goes to the gate then.


santoshkumarradha:
Landed on `santos/dev` as cb3f0f356, squashed from head 3926e7991.

Evidence, all on Spark, fresh clone, exact commit: gofmt clean, change entries pass, build ok for the host and for Windows, then full `make check` on the stack top cb3f0f356 (tree 56fb6e9c7): EXIT 0, 115 packages ok, first run, log `check-sdev-cb3f0f356.log`.

PR #1288, opened 2026-09-20
standing: a row that needs you names what it needed and opens the item

santoshkumarradha:
**In one sentence:** a watch that stopped because it needed permission now says what it wanted to run, opens the watch instead of an idle conversation, and can be put down by changing the watch.

## What a person saw

The home screen's `needs you` list carried a row for a watch, marked as needing somebody, whose second line read:

```
refused in a task: default — nobody to ask
```

Four pieces of machinery in the one line that is supposed to tell a person what to do. `enter` on it opened the conversation the watch had been asked for in. That conversation was idle and held nothing to answer. There was nowhere on the screen to act, and the row could not be put down, because only the watch's next firing rewrites that line and this watch had spent its allowance for the day.

## What is true underneath

A firing runs with nobody to ask. A call the person's own approval rules would have put to somebody is therefore refused where it stands, and the worker is told so in the engine's own words, which is right: the worker has to act on them. Those words were then handed straight to the surface.

**This change does not widen what a firing may run.** That is a separate decision and it is open.

## The three halves

- **The line is the person's, and the engine's is still the engine's.** The reader that recognises a refusal now answers with `stopped: it needed your ok to run ` and what the call actually wanted: the command for a shell call, the name of the hand for anything else, `something` when the arguments cannot be read. `needs your ok to run` is the sentence this surface already says when a conversation waits for permission, so the two rows read as one sentence in two tenses rather than two accounts of one fact. The command is read through the same table every other reader of a call uses, so this line and the row the same call would have drawn in a conversation name the same thing.

  **It never says how much was refused.** It is written from the first refused call and knows nothing about the calls that came back fine before it, so it cannot claim that nothing was allowed.

- **The door is the item.** A standing row whose line is about a permission now says `enter on standing`. The thing that stopped is the watch; the conversation holds no question. The watch is where it can be paused, reworded or let go. A row whose line is a QUESTION the firing asked keeps the door it had, because the question was asked in a conversation and that is where it reads.

- **The person can put it down.** Changing the watch clears the line, alongside the next firing which already did. A question is never cleared this way. Opening the item and closing it again does not clear anything either, because nothing moved and the reason is still true.

## One predicate, and what it reads

The line is recognised in one place, asked by the store when a person changes an item, by the surface when it picks a row's door, and by the engine's own reader, so a line cannot be a permission in one place and a question in another.

**It matches what a line says, not how it says it.** Three doors in this build refuse a call that needed a person and had none, and they do not share a sentence:

| Door | What it writes |
| --- | --- |
| the turn inside a task | `refused in a task: <rule> — nobody to ask` |
| the turn with no resolver | `needs approval but no resolver is attached: <rule>` |
| the door that runs this program underneath another one | `<tool>" needs somebody to say yes (<rule>) and nobody is here to ask, so it was refused` |

Holding those as three literals is how the third one came to be unrecognised in the first place, so a shape is a set of parts a line must all carry. The absence of a person is not enough on its own: a QUESTION is model prose and can say "there is nobody on call, who do you want me to ask", so the shape also requires that something was **refused**. Both doors that state the absence say so.

**It knows the old spelling as well as the new one.** Every item stuck on disk today carries the engine's own refusal verbatim, and a watch that has spent its allowance for the day cannot fire again to have it rewritten. A predicate that knew only the new lead would have left every row that provoked this change exactly as it found it.

## Tests

- The line names a synthetic command, carries none of the four machinery words, and claims nothing about calls it did not see.
- A hand that is not a shell call names the hand. Arguments that will not read leave the line saying `something`.
- An ordinary tool failure still raises no line for the person.
- The engine's own refusal is unchanged by the reader that matches on it.
- All three doors' sentences are read as permission stops, and two questions carrying the absence of a person but no refusal are not.
- Changing an item clears the line and moves nothing else about the item; a question the firing asked survives the change; an item needing nobody is untouched.
- A row in the old spelling takes the item's door, and a row carrying a question still opens its conversation.

## Proof

On Spark, under `umask 022`, plain ssh, a detached tree of the exact sha. Exit codes read directly. **Every row below ran on head `078216012`, on santos/dev `cb3f0f356`.**

| Step | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet` on session, standing, tui3 | exit 0 |
| `internal/standing`, whole | ok |
| `internal/session`, the standing and refusal tests | ok |
| `internal/tui3`, the needs, standing and home tests | ok |
| `make test-laws` | exit 0 |
| **`internal/session`, whole suite** | **ok, 249.5s, exit 0** |
| **`internal/tui3`, whole suite** | **ok, 326.2s, exit 0** |

Both whole suites went through `scripts/one-suite.sh`. Finished 2026-09-20T15:49Z.

### Two reds on the way here, both real

An earlier head was red on `internal/tui3` on one test, and it was this change's own new one: it read the drawn frame for a word that only draws under the cursor. The test was wrong and the code was not; it now reads the row's own cell, which is the claim it makes.

A later head was red on `internal/standing`, and that one was a defect. The predicate held two literals and this build has three doors, so the third door's sentence read as an ordinary tool failure. Matching on what a line says rather than how it says it is the fix above.


santoshkumarradha:
Landed on santos/dev as 1357788e2, squashed by hand. The squash tree is identical to this PR head 078216012.

Gate: one full make check on Spark at 1357788e2, fresh clone of the pushed branch, EXIT 0 on the first run. 115 packages ok, zero FAIL, zero TempDir RemoveAll. Pre-checks on the same clone: gofmt clean over internal and cmd, 60 change entries well formed, go build ./... and GOOS=windows go build ./... both green, cross build 53s.

Suites owed on this PR, all at 078216012: light chain exit 0, whole internal/session ok in 249.5s, whole internal/tui3 ok in 326.2s.

Two defects were found during review and fixed before the gate. The predicate knew two of the three doors that write a permission refusal, missing the one the subharness writes. And the property that replaced the literals matched an absence without a refusal, which would have read a genuine question as a permission stop and put it down when the person paused the item; the shape now requires the refusal word alongside the absence.

PR #1289, opened 2026-09-20
tui3: a run task's page draws its brief through the reader the transcript already uses

santoshkumarradha:
**In one sentence:** the page a person opens for one task of a run draws that task's brief through the reader the conversation's transcript already uses, so it opens on the work rather than on the scaffold addressed to a worker.

## What a person saw

A run task's page has a `brief` block, folded to three lines. When the task's stored brief was the generated work order a run hands its workers, those three lines were:

```
WHAT THE PERSON ASKED FOR, IN THEIR OWN WORDS
This is the message this work came out of. Treat it as the ask...
```

Two of the three folded lines were machinery, and the third was the rule under the heading. The person's actual ask began below the fold. The page opened on a document addressed to a model.

## What was already there

The surface already owns a reader for exactly this text. The conversation's own transcript draws the same generated work order through it: it recognises the generated opening, relabels the shouted section headings into plain ones, and puts this task's own work first so a child does not open on its parent's whole project.

The transcript had it. The plan page did not, and drew the stored text raw.

## The change

The reader's shared half is lifted out to work over a plain string, and the plan page's brief section calls it. Both surfaces now give one text one reading.

- **A brief that is not the generated document is untouched.** The recognition gate comes first, so a brief a person typed themselves, or wrote by hand, wraps exactly as it always did.
- **The stored text never changes.** Only what is drawn does. The model's input, the payload used for identity and refresh, and the transcript's own caller are all as they were.

### And the fold counts what it draws

`ctrl+o` opens and closes the brief, and it decides whether there is anything to open by counting lines. It was counting the raw wrap while the block drew the reshaped one, so the key and the block could disagree about how much there was. They now count the same function. **This was not in the brief.** It was found while the work was being done, by reading the one other caller of the line counter.

## Tests

- A stored brief that is the generated work order draws with the scaffold heading gone and the task's own work in the first folded line.
- A brief that is not that document draws byte for byte as before.
- The transcript's own drawing of the same text is unchanged.
- The fold key and the block agree on the line count for both shapes of brief.

## How this was written

This change was written by `codeaf do` driving itself, on Spark, from a short brief naming the defect and pointing at the two files. The run planned, edited, tested and committed on its own. What is measured below is the run that produced the commit, read from the usage ledger rather than from the run's own account of itself.

| | |
| --- | --- |
| ending | done, exit 0 |
| wall clock | 8 minutes 52 seconds |
| cost | $0.97 |
| model calls | 73 |
| tokens in | 3,361,876 |
| tokens out | 36,529 |

Two seats: 68 calls on the work seat and 5 on the planning seat. Input dominates output by about ninety to one, which is the shape of a worker that reads a large tree and writes a small patch.

The commit is the run's own, with its own authorship kept. Two commits follow it: the change entry's number, which the run guessed before a pull request existed, and nothing else.

## Proof

On Spark, under `umask 022`, plain ssh, a detached tree of the exact sha. Exit codes read directly. **Every row ran on head `05410eb66`.**

| Step | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet ./internal/tui3` | exit 0 |
| `internal/tui3`, the brief, plan and request tests | ok |
| `make test-laws` | exit 0 |
| **`internal/tui3`, whole suite** | **ok, 326.1s, exit 0** |

## What this does not fix

The task page has other regressions and this is not them. The `folder` section was deleted while the manual still promises it, the page carries a fixed block of dead vertical space already logged at high severity, and the harness page has no captions where the room page has them. They are separate and they are still open.


santoshkumarradha:
Landed on santos/dev as f79621233, squashed by hand onto 1357788e2 (which carries #1288). The base moved, so the trees differ by #1288; the change this PR introduces is byte identical on both bases, checked by diffing the two patches.

Gate: one full make check on Spark at f79621233, fresh clone of the pushed branch, EXIT 0 on the first run. 115 packages ok, zero FAIL. Pre-checks on the same clone: gofmt clean over internal and cmd, 61 change entries well formed, go build ./... and GOOS=windows go build ./... both green, cross build 58s.

Suites owed on this PR, at head 05410eb66: light chain exit 0, whole internal/tui3 ok in 326.1s. The delta from c752196ec to 05410eb66 is one commit renumbering the change entry from 1286 to 1289 and touches no Go file.

PR #1290, opened 2026-09-20
session: a standing firing runs behind the gate a task node runs behind

santoshkumarradha:
A standing watch fires with nobody there, and until this it was refused its own work. standingRunConfig copied the parent conversation's config, set AskConsent false and InTask true, and never set an approval policy, so a firing carried the conversation's prompt-a-human default while also being told it was inside a task where no person exists. consent.go answers that combination with an in-task refusal: the decision needed a person and there is none. What survived was the read-only lift, which covers git status and nothing else, so a watch could not run one shell command, not even one the person had explicitly allowed. The owner's own watch fired 39 times between 2026-09-16 and today, spent $3.31, and did nothing.

What changes, and it is one line: a firing carries the policy a task node carries, `&approval.Policy{Default: approval.ActionAllow}`, the same value task_run.go sets beside the same AskConsent false and InTask true.

It is not wider than a node. approval's critical table still turns an allow into a prompt for the handful of shapes that destroy a disk or drop the machine, and a prompt with nobody to ask is still a refusal the firing can read and report. The calls that act in the person's name outside this machine stay refused. A bash call whose arguments cannot be read stays refused. No path rule was invented.

THE DECISION IS THE OWNER'S. Santosh chose this shape in his own words on 2026-09-20, "lets go with allow same as task node", after both shapes were put to him with the trade-off. This note is here because the change widens what an unattended firing may do, and a change of that kind should carry who decided it and when.

The test asserts the CALL RUNS rather than reading the policy field: a field assertion would go green the day someone sets the field and breaks the behaviour another way. Failing-first on base it reports "the call never ran: a firing behind its own gate still could not run one command", and it passes with the change. It deliberately pins no refusal wording, so #1288's sentence work is free to change without touching it.

HOW THE GREEN WAS GOT, INCLUDING THE RED. The first whole internal/session run on this branch came back FAIL, and not in code: its tests printed PASS and the package then failed the suite's own repository-churn guard, which saw the checkout's head move from 481ceb8ae to 2dee07970 while it was running. It moved because I amended mid-run. A grep over the added lines found an em dash in a comment where I had quoted the engine's in-task refusal verbatim, and this repository's prose carries no em dashes, quotation included; the comment now says what that line says without reproducing it character for character. The amend changed exactly four comment lines in standing_consent_test.go and no code, tree d5e082dd9 to daabb2f04. The guard fired correctly and the fault was mine, so the run of record is the rerun below, taken on a quiescent tree with nothing touching the checkout.

Verification on Spark: go build ./...; whole internal/session through scripts/one-suite.sh; the standing, consent and needs-line tests focused; go test ./internal/guard ./internal/namelaw ./internal/approval; gofmt -l ./cmd ./internal empty; go run ./cmd/codeaf-changes check.


santoshkumarradha:
Landed on santos/dev as 8b682938d, squashed by hand onto 3d95e7072.

Gate: the first full make check on Spark at 8b682938d was RED, on one named test. The rerun on the same commit was EXIT 0, 115 packages ok, zero FAIL. Both logs are kept; the red one was not overwritten.

The red was TestAParkedRootsRunReviewsItsChildBeforeItAnswers in internal/run, "outcome = ran and did not finish, want done", in 0.49s under a 30 second context. It is filed as one unexplained sighting with its load window, deliberately not folded into the open flake family, because that family is in a different package and a fourth member added for tidiness is how you stop looking.

What is known about it, none of which is a cause:

- internal/run was green on the four previous gates: cb3f0f356, 1357788e2, f79621233, 3d95e7072. First failure.
- The change is confined to standingRunConfig, which is unexported and reached only on the standing lane. internal/run never constructs a firing and references the lane nowhere.
- The test ran five for five green on the same commit afterwards, 1.88s total.
- The test's own comment says it holds the child's return back until the late finish has been tried, "the way a loaded box held it". It was written to reproduce contention.
- The gate ran it while three packages took between four and five minutes each, with another session's heavy suite running outside the suite lock.

A rerun going green says the failure is not deterministic. It does not say what it was.

Pre-checks on the gated clone: gofmt clean over internal and cmd, 63 change entries well formed, go build ./... and GOOS=windows go build ./... both green, cross build 55s. The owner's decision on the policy widening is recorded in the body with its date.

PR #1291, opened 2026-09-20
session: a run is work, and its life is the conversation's

santoshkumarradha:
**In one sentence:** a live run counts as work, so the engine stops retiring the conversation out from under it, and closing that conversation now ends the run instead of leaving it spending against a room nobody can read.

First cell of the settled close-window design, #1224.

### A run is work

The work tree is the one door every surface and the engine read to ask what is happening. A run's rows are kept beside the task graph rather than in it, so the walk over the graph's own nodes never saw them. A conversation driving a run and nothing else answered that it was working on nothing.

That answer is not cosmetic. The engine asks this same door what idle means before it retires a conversation at thirty minutes. So a run longer than half an hour had its room taken down while its workers were still out.

The run now joins the tree, under the number every other surface already knows it by, so a reader can hand the id straight to the stop road without knowing a run from a task. **Nothing else had to change to keep the engine alive while a run lives**, because the retirement and the host's own quiet both hang off that one reading.

### A run's life is the conversation's

The run drops the proposing turn's cancellation on purpose: it has to outlive the sentence that asked for it. Nothing else could reach it. So its life belonged to neither the turn nor the room but to the process, and a conversation that ended while a run was going left workers spending against a conversation nobody could read or stop.

Closing a conversation now cuts it, in the same place and for the same stated reason as the adaptive runs it already cut. **It writes nothing on the run's own record**: a person's stop puts their reason on the store's root and settles the row in their words, and the room closing is not that. What the run did is in its store, which is where the next launch reads it from.

## Not in this cell

Reading that store back is cell b. This cell does not change what a person sees on reopen, and an unsettled run still reads as stopped there.

## Tests

Five, and a negative control was run on all of them.

- A conversation driving a run is working, with the run's state, title, birth time off the conversation's own clock, and the id every other surface knows it by.
- A conversation driving a run is not idle, which is the reading that keeps the room open.
- A run a person stopped is still work until it settles, and leaves the tree when it leaves the conversation.
- Closing the conversation ends the run.
- Closing the conversation writes no stop on the run's record.

**The negative control.** With the two lines of the fix removed in place, four of the five fail and each names the defect rather than a shape:

```
--- FAIL: TestAConversationDrivingARunIsWorking
    a conversation driving a run says it is working on 0 things, want the run
--- FAIL: TestAConversationDrivingARunIsNotIdle
    a conversation whose only live work is a run reads idle, which is what retires it mid-run
--- FAIL: TestARunAPersonStoppedIsStillWorkUntilItSettles
    a run that is stopping stopped counting as work before its workers were home
--- FAIL: TestClosingTheConversationEndsTheRun
    the conversation closed and the run was left running against a room nobody can read
```

The fifth passes either way by design. It guards the fix against going too far, so it has nothing to say about the defect.

## Proof

On Spark, under `umask 022`, plain ssh, a detached tree of the exact sha. Exit codes read directly. **Every row ran on head `6638d40f2`.** The branch head is `c437c6f36`, which is that commit plus one that corrects the change entry to this pull request's own number. No Go file differs between the two.

| Step | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet` on session, remote, enginehost | exit 0 |
| the five new tests, named and verbose | 5 of 5 pass |
| the same five with the fix removed | 4 fail, 1 passes by design |
| `internal/remote`, whole | ok |
| `internal/enginehost`, whole | ok |
| `make test-laws` | exit 0 |
| **`internal/session`, whole suite** | **ok, 249.1s, exit 0** |

`internal/remote` and `internal/enginehost` are whole rather than focused because they own the retirement this change moves.


santoshkumarradha:
Landed on santos/dev as 3d95e7072, squashed by hand onto f79621233. This PR was built on the head before #1289, so the trees differ; the change it introduces is byte identical on both bases, checked by diffing the two patches against the true merge base.

Gate: one full make check on Spark at 3d95e7072, fresh clone of the pushed branch, EXIT 0 on the first run. 115 packages ok, zero FAIL. Pre-checks on the same clone: gofmt clean over internal and cmd, 62 change entries well formed, both builds green including Windows, cross build 59s.

One claim in review needed correcting before this landed. It was stated that an acceptance drive had shown a run still stepping after a window closes. No such drive was run; that sentence was the cell plan, not a result, and I repeated it as evidence. The question it was meant to answer is now settled by reading instead.

A closing window does not reach Agent.Close in hosted mode. internal/tui3/detach.go is Agent.openSession split along the line that used to close the conversation being left, and that line was removed on purpose. internal/remote/client.go Detach sends MethodDetach against a session host, with the reason written out: MethodClose would end a running task on behalf of somebody who only shut a window. MethodClose reaches Agent.Close through closeLeaving, documented as a person deliberately ending the conversation.

Against a one-shot engine, Detach says leaving IS ending, so Close is reached there. That is the better outcome: the process is going away regardless, and an orderly cut beats an orphaned run spending into a room nobody can read.

PR #1293, opened 2026-09-20
session: a run stopped on its own question says a person is needed, and what for

santoshkumarradha:
A sub-harness run that asks a question blocks inside subharnessEnv.Ask until somebody answers it, and the session went on telling every other window it was working. The question was registered in a.subharnessAsks and readable through OpenQuestions the whole time, but waitingOnPerson never read that lane, so presence never went to waiting, no row appeared on home, no mark was drawn, and the work stopped in silence until somebody happened to open the conversation. The lane's own comment already claimed the row said so.

Two halves, and neither is a fix on its own:

- taskpresence.go: the predicate reads `len(a.subharnessAsks)` alongside the lanes beside it.
- subharness_env.go: the lane raises through `presenceAskingWhole` rather than `raiseQuestion`, so its row is banked at the presence desk and comes down with the question. This matters because the sentence a person reads is taken off the desk (`presenceAsk`), so counting the lane alone would turn the session amber with nothing under it.

ONE TEST, THREE STATES, which is how both halves are shown to be load-bearing rather than asserted. On base it fails with "a session stopped on a sub-harness question says it is working: {waiting:false reason:}". With the predicate alone and no desk row it fails with "the row says a person is needed but not what for". With both it passes at -count=5. It drives the real seam, agent.subharnessEnv and env.Ask on a live agent with consent on, so it would catch this defect in whatever form it came back rather than proving that a predicate reads a map.

THE ROAD WAS SPLIT BECAUSE THE LAW SAID SO. Counting one more lane put waitingOnPerson at sixteen endings against a ceiling of fifteen (complexity_test.go's ratchet), and that ledger only shrinks, so the function is now two jobs along the seam its own comments had already drawn: personAskLanes reads every lane under the agent's lock, waitingOnPerson decides which answer a person reads first. The lock discipline is unchanged and a.mu is still released before PendingDecisions. The road's doc said "It is one function because it is one question", which the split would have made false, so it now says there is one answer given in two steps.

SIBLING LANES ARE DELIBERATELY LEFT. Sign-in, the sub-harness offer and the harness design card raise the same way without banking a desk row, so each already draws a mark with an empty sentence under it. They belong to the desk piece, which needs the owner's word on the state words first, and a change that quietly repaired three lanes while claiming to repair one would be a change nobody could review. A landed task that says `your call` on a page that cannot take the call is the same disease on another lane, also untouched.

Verification on Spark: whole internal/session through scripts/one-suite.sh; the complexity law; the presence, question, waiting and subharness tests focused; go build ./...; gofmt -l ./cmd ./internal empty; go run ./cmd/codeaf-changes check; go test ./internal/guard ./internal/namelaw. The whole-package run whose only failure was TestARecallThatLandsBeforeTheFirstWordIsAskedAgainWithIt, a separately tracked item that fails only under a loaded box and passes alone at -count=10, is being replaced by a clean run on a quiet lock; that head will be posted here.


santoshkumarradha:
## Clean whole internal/session run, as promised in the body

Green on this PR's head.

```
== head 77dbd72dd5f2bbe8aa49bb324b0b825ba99dd6db tree 1fe000b2a6bd26bb70fe3c99bb1082a98b162aff ==
== pkgs ./internal/session ==
ok  	github.com/Agent-Field/codeaf/internal/session	246.110s
== tree after 1fe000b2a6bd26bb70fe3c99bb1082a98b162aff ==
```

Host spark, worktree `/home/santosh/src/trees/subask-f79621233`, log kept at
`/home/santosh/src/trees/suite-subask-f79621233-internal-session-20260920T165926Z.log`.
Whole package, `-count=1`, under the heavy-suite lock (suite pid 754526, lock taken
17:04:28Z). The tree sha is identical before and after the run, so the repository
churn guard had nothing to report and the number belongs to this head.

This replaces the earlier run whose only red was
`TestARecallThatLandsBeforeTheFirstWordIsAskedAgainWithIt`. That test passed here
along with the rest of the package.

### What the box was doing, because the number is worth less without it

The run held the lock but did not have the machine. A full `internal/tui3` suite
was running outside the lock for part of the window, invoked directly rather than
through `scripts/one-suite.sh`, so nothing serialised it against this run. Load
average was sampled every 15s beside the suite log for the whole window:

| time (UTC) | 1 min | 5 min | 15 min |
| --- | --- | --- | --- |
| 17:04:27 (lock taken) | 4.29 | 5.01 | 6.11 |
| 17:04:42 | 9.51 | 6.07 | 6.44 |
| 17:06:27 | 5.07 | 5.52 | 6.20 |
| 17:08:42 (run ended) | 4.29 | 4.98 | 5.90 |

So this is a green taken with real load on the box rather than on an idle one,
which is the stronger of the two readings for a flake that has only ever appeared
under contention.

santoshkumarradha:
Landed on santos/dev as c975972c9, squashed by hand onto 8b682938d.

Gate: one full make check on Spark at c975972c9, fresh clone of the pushed branch, EXIT 0 on the first run. 115 packages ok, zero FAIL. Pre-checks on the same clone: gofmt clean over internal and cmd, 64 change entries well formed, go build ./... and GOOS=windows go build ./... both green, cross build 55s.

Owed suites were delivered on a quiescent tree at 77dbd72dd, whole internal/session ok 246.110s, tree identical before and after so the churn guard had nothing to say.

Two things about this one are worth keeping.

The test has three states rather than two. On base it fails saying a session stopped on a sub-harness question reports itself as working. With only the predicate and no desk row it fails saying the row claims a person is needed but not what for. With both it passes. Half the fix cannot land by accident, and a reviewer who reads only the test learns the whole shape of the defect.

Three sibling lanes share this shape and are deliberately left: sign-in, the sub-harness offer and the harness design card all raise without banking a desk row, so each draws an attention mark with no sentence under it today. They belong to a change about that desk and are named in the body rather than quietly repaired here.

One red on the way, not in this code: the first whole-package run tripped the repository-churn guard because the branch was amended mid-run. That guard did exactly its job. The run of record is the quiescent one above.

PR #1296, opened 2026-09-20
session: work nothing is driving reads as interrupted, not failed and stopped

santoshkumarradha:
**In one sentence:** work that was live when the process holding it went away comes back saying `interrupted` — nothing is driving it, everything it did is kept — instead of being stamped failed and stopped.

Second cell of the settled close-window design, #1224. Stacked on #1291.

## What a person saw

A run was mid-flight when codeaf closed. The next morning the conversation reopened and the run's rows were on the column again, stamped **failed** and **stopped**, with the run's own row reading:

```
it ended when codeaf closed; its journal is kept
```

Two untrue things in one line. **Failed** says something went wrong with the work; nothing went wrong, nothing was even found out. **Stopped** says a person ended it; nobody did anything. And `it ended` says the work is over, when every step it took is sitting in its store.

## The word

`interrupted` is the owner's own, settled on 2026-09-20. It means **nothing is driving this, and everything it did is kept**.

It is deliberately neither of the two words it used to be read as. `stopped` stays a person ending the work. `incomplete` stays work that ran and came up short. A closed window, a sleeping machine and an engine that went away are none of those, and collapsing them into either is how work whose only misfortune was a closed window came back looking as though it had gone wrong.

It is also the only landing word that does not mean the work is over.

## Where the word lives

On the reading, which is the one place this program spells a person's words. One rung on the state ladder and one on the presence ladder, and the tier, the mark, the two answers and the sentence beside the word all follow from them. **No surface grew a table of its own.** The only line touched outside that reading is home's mark, so an interrupted row wears the asking mark the rows beside it wear.

- **It sits in the person's tier.** It will not move until somebody answers, which is what that tier means.
- **The two answers are `continue it` and `leave it`,** spelled here so the card in cell c draws words it did not invent.
- **The answer is never the model's,** whatever a settle policy says, because continuing spends money. That is the owner's second answer applied where it bites.
- **Interrupted is settled in the scheduler's sense and not the person's.** No worker holds it and its slot is back, which keeps a restored row off the live work tree; the work itself is still waiting to be picked up.

## A background job is untouched

A job is a process codeaf forked, and a forked process cannot outlive the program that forked it. There is nothing to continue, so `stopped` stays the honest word and the row keeps the path to its log.

## The seam between this cell and the last

**Between #1291 landing and this one landing there is a real window, and it is not an oversight.** #1291 makes a conversation that genuinely ends cut its run. Until this cell lands, reopening still stamps that run failed. So for that interval the product cuts work and then misreports why. The money bleed is what #1291 stops and it is the worse of the two; this cell closes the other half.

## The manual, both directions

- The new word reaches the page that lists the words a task can land with.
- **The page that told a person their run ended when codeaf closed now says what is true.** That page quoted the old sentence verbatim, and a page naming a word while denying the thing around it satisfies every gate and still lies.
- `interrupted` has its own heading, in the words somebody would actually type: what it means, whether they lost their work, whether they can carry on.

**That heading exists because the corpus caught me.** Folding the word into the section that lists the landing words cost another question its page: `why does my task say not done under it` stopped reaching its page and landed on a different one. The section was still well inside the length rule, so this was **dilution rather than size** — a paragraph about a new word spread that section's weight over more ground and it fell behind a competing page. The fix is the page and never the probe.

## Tests

Six new, and two existing ones rewritten rather than deleted. Both of those pinned the old sentence, which is exactly the behaviour this cell changes, so they now assert the new truth.

- The word, and that a closed window is not drawn as a fault.
- The tier, the attention, the ask kind, both answers, the reason, and that the decision is the person's.
- That it is none of stopped, incomplete or done.
- That it holds no slot and is still not over.
- That a row live when the process went away comes back interrupted, unstopped, and invents no sentence about how it ended.
- That a row which had already settled is unchanged by being read back.

### Negative controls

Run twice, once per independent claim.

**Remove the store arm** and three fail, each naming the defect:

```
--- FAIL: TestAResumedConversationRedrawsAMidFlightRunInterrupted
    row 1 came back failed, want it interrupted — nothing is driving it and nothing went wrong
--- FAIL: TestARowThatWasLiveWhenTheProcessWentAwayComesBackInterrupted
    a running row came back failed
--- FAIL: TestAJobStillMovingComesBackStoppedAndKeepsItsLog
    a run's row came back as failed (stopped true), want it interrupted and nobody's stop
```

**Remove the reading arm** and two fail:

```
--- FAIL: TestWorkNothingIsDrivingSaysSoInTheOneWord
    it reads as ""
--- FAIL: TestInterruptedWorkIsThePersonsCallAndSaysWhatTheAnswersAre
    it sits in the "" tier, where nobody would be asked
```

The tree was restored and the whole set reconfirmed green after each.

## Proof

**Re-taken after a rebase, and again after a formatting fix.** This branch was first proven at `68d67632d`, before #1291, #1290 and #1293 landed. Nothing from that run is carried over.

**One row of the earlier proof was not a proof at all.** `gofmt -l` lists unformatted files and exits 0 whichever way, so a step that read its exit code reported green on any tree. An unformatted const block in this change reached a gate because of it. The block is now formatted by the tool rather than by hand, and the check captures the list and fails on non-empty, never consulting the exit code.

On Spark, under `umask 022`, plain ssh, a detached tree of the exact sha. **Every row ran on head `ccbfbe692`,** rebased onto santos/dev `c975972c9`.

| Step | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet` on session, tui3 | exit 0 |
| `gofmt -l ./cmd ./internal`, list tested for emptiness | no files |
| the new and rewritten tests, named | 23 pass, 0 fail |
| both negative controls, taken at `68d67632d` | 3 fail, then 2 fail |
| `make test-laws` | exit 0 |
| `make changelog-check` | 65 entries, all well formed |
| **`internal/session`, whole suite** | **ok, 243.8s, exit 0** |

The one-minute load average ranged from about 2 to 5 across the run, so this is a green on a working box rather than an idle one.


santoshkumarradha:
Landed on santos/dev as 3d3b7542b, squashed by hand onto c975972c9.

Gate: one full make check on Spark at 3d3b7542b, fresh clone of the pushed branch, EXIT 0 on the first run. 115 packages ok, zero FAIL. Pre-checks on the same clone: gofmt clean, 65 change entries well formed, both builds green including Windows, cross build 56s.

This took two attempts at the gate and the first stop is worth recording. My pre-check reported one unformatted file where the four landings before it had reported nothing. The branch was formatting-clean by its own chain, and that chain read the exit code of a tool that lists offending files and exits zero either way, so it could not fail. Its report filter then dropped the filename, so the one real signal never reached the eye reading it. Two independent reasons to see green.

Nothing got through. I re-ran the check with its output captured on the gate clone of all five commits landed earlier today and all five are clean; the gate runs every check itself on a fresh clone of the exact squash and reads none of the numbers in a PR body. That property is the only reason a broken chain cost one gate cycle rather than a day. Both the chain and this gate now test the list for emptiness and never consult the exit code, and an audit of the rest of that chain found one more of the same shape: two commands on one line with a single exit code at the end, which reported only the second.

The stacked pull request for the run copy record follows onto this head.

PR #1298, opened 2026-09-20
session: a run's working copy is written down, so carrying it on adopts the work

santoshkumarradha:
**In one sentence:** a run's working copy is written down when the run starts, so carrying it on later adopts the work instead of deleting it.

Third cell of the settled close-window design, #1224, split out ahead of the card on the manager's ruling: the piece where getting it wrong deletes somebody's work gets its own proof and its own gate.

## The hazard this exists to prevent

A run works in a copy of its own. The road that makes that copy carves the directory it is handed and cuts a fresh branch with a random suffix. It is documented as safe to carve, and it is — **exactly once**, because the only thing that can be sitting at a run's path is that run's own wreckage.

Ask it for the same run's copy a second time, which is precisely what a naive "carry on where it left off" would do, and it deletes every step of the work it was meant to resume, cuts a new branch off a new random name, and hands back something that looks perfectly healthy. **Nobody would see it until it was their work.**

## The asymmetry that makes the record necessary

**The directory is derived and the branch is not.** A run's directory is a pure function of the conversation's folder and the run's own number, so any process can work it out again. Its branch carries a random suffix minted at the moment the copy is cut, and nothing wrote it anywhere.

That is the whole reason this record exists. A reader who believes both are reconstructible will delete this change in six months.

## What it does

A run's row now carries its copy: the directory, the branch, the repository it merges into, the ground and how the copy stands on it, the checkout it was cut from and that checkout's commit, and which rung of the ground ladder made the world. Those are the same facts a node of the session's own graph has always carried in its own record, under their own names.

The reader **adopts and never makes**. Nothing in it carves a directory, cuts a branch or derives a path. Every road out is either the copy that was written down or a refusal.

- **A run with no copy recorded is refused, in words naming what is missing.** Every run row on disk today predates this field, so that refusal is the first thing a person will meet. **It is the correct answer forever, not a stopgap.** Such a run's branch was minted at random and recorded nowhere, so no reading of the record, the conversation or the disk can recover it. Anything the reader could do instead is the destructive road wearing a helpful face.
- **A copy that is gone from disk is refused too, and the refusal names the branch,** because the branch is the one thing left that a person can act on.
- **The copy is carried forward in the one function every publisher of a run's row goes through.** **Every publish of a run's row replaces that row outright** — it is not a patch, and that is the fact which makes the carry necessary; it is not visible in the diff. Only the first publish knows where the work is, so any later one that had not thought about the copy would quietly drop the single fact nothing else can recover, and the row would still look complete. Putting the carry in each caller instead would make the record depend on four people remembering, and on the next publisher someone adds knowing to. A notice that names a copy of its own still wins, because it is the more recent reading.

## Not in this cell

Nothing calls the reader yet. This cell makes a continue possible and safe; the card and the key are the next one.

## Tests

Seven, and the first is the one that matters.

**The destruction, shown.** A copy is cut, work is written into it, and the copy is asked for a second time exactly as a resume would ask. The work is gone, and the second branch has a different name from the first while the directory is the same. That is not a bug this change introduces — it is the behaviour of the road a resume would otherwise reuse, pinned so nothing reaches for it again. It skips with an explanation rather than failing if that road ever stops clearing the directory, because then the reason for the record has changed and the test should be rewritten rather than deleted.

The rest: the record is taken from the tree the run is actually working in; it survives the round trip through the store, which is the only trip that matters; a run with no copy is refused; a copy that is gone is refused and names the branch; a copy that is there is adopted whole with the merge still open; and a later publish keeps what the first wrote down.

### Negative controls

One per independent claim.

**Stop the record travelling through the store:**

```
--- FAIL: TestTheCopySurvivesBeingWrittenDownAndReadBack
    the copy did not survive the record, so nothing could be carried on
```

**Stop a later publish carrying it forward:**

```
--- FAIL: TestALaterPublishKeepsTheCopyTheFirstOneWroteDown
    the second publish dropped the copy, so the run became one nobody can carry on
```

The tree was restored and the set reconfirmed green after each.

## Proof

**Re-taken after a rebase onto the current head.** Both negative controls were re-run on the rebased code rather than carried over, because a resolution that dropped half the carry-forward would pass an ordinary suite and fail exactly one of them.

On Spark, under `umask 022`, plain ssh, a detached tree of the exact sha. **Every row ran on head `33640353b`,** stacked on cell b's `ccbfbe692` over santos/dev `c975972c9`.

| Step | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet` on session, tui3 | exit 0 |
| `gofmt -l ./cmd ./internal`, list tested for emptiness | no files |
| the seven new tests, named | 7 pass, 0 fail |
| negative control, the record stopped travelling through the store | the round-trip test fails, saying the copy did not survive |
| negative control, the carry-forward removed | the publish test fails, saying the second publish dropped the copy |
| `make test-laws` | exit 0 |
| `make changelog-check` | 66 entries, all well formed |
| **`internal/session`, whole suite** | **ok, 241.6s, exit 0** |

The tree was restored and the whole set reconfirmed green after each control. The one-minute load average ranged from about 1.7 to 6.2 across the run, so this is a green under contention rather than on a quiet machine; the range is given rather than a single number because a single number would suggest an idle box.


santoshkumarradha:
Landed on santos/dev as 1a692da6e, squashed by hand onto 3d3b7542b.

Gate: one full make check on Spark at 1a692da6e, fresh clone of the pushed branch, EXIT 0 on the first run. 115 packages ok, zero FAIL. Pre-checks on the same clone: gofmt list empty, 66 change entries well formed, both builds green including Windows, cross build 53s.

This pull request was stacked on the interrupted change, so before committing the squash I checked what the staged diff contained besides this change, and compared it by content against this branch own commits rather than trusting the file list. Two files appear in both changes. The squash is this change alone.

The proof was re-taken after a rebase rather than carried over, and both negative controls were re-run on the rebased code rather than labelled: removing the store carry fails the round trip, removing the publish carry fails the publish test. That mattered here, because a resolution that silently dropped half the carry forward would pass an ordinary suite and fail exactly one control.

The reason the work was split this way is worth recording. The first plan put this record and the card that uses it in one change. Reading the road that makes a run a working copy showed it mints a new branch name with a random suffix on every call and clears the directory it is handed, so asking it a second time for a run being resumed would have deleted the work and returned an empty branch that looks correct. The piece whose failure mode is destroying somebody work got its own change, its own proof and its own gate, ahead of anything that draws.

PR #1304, opened 2026-09-20
manual: the interrupted page no longer offers two answers nobody can press

santoshkumarradha:
#1296 landed the `interrupted` reading an hour ago and its page says the row "offers two answers: `continue it`, or `leave it`". Both words are real in the reading. Neither is drawn for a run. This removes the promise and states the limit in its place, rather than waiting for the change that makes the answers real to carry it.

## Why the answers are not drawn

There are exactly three readers of `TaskAsk.Yes` and `.No` in the tree, and a run row reaches none of them.

| Reader | Fed by | Why a run row cannot arrive |
| --- | --- | --- |
| `internal/tui3/homepanel_needs.go:424` | `session.TaskIndexEntry` | run rows never reach the task record; `recordTaskIndexEntry` is called only from `orchestrate.go`, and `TaskIndexEntry.StatusFacts` carries no copy |
| `internal/session/question.go:3086` `landingOptions` | `landingQuestion(PendingDecision)` | `PendingDecisions()` walks `graph.nodes` and admits only `state == TaskUnverified`; a run's rows live in the separate `g.runs` map and are never unverified |
| `internal/tui3/taskrecord.go:819` `taskRecordVerbs` | a `questionShown` | needs a question already on screen, and `app.taskRecordLanding` matches only the landing and the conflict |

## What is kept, because it is drawn

The word, the asking mark, and the column that waits for a person all come from `app.taskStatus`, which builds its facts from the surface's own node and so does read a run row. The line `nothing is driving it; everything it did is kept` is `TaskStatus.Reason` and is written by `taskNoteHead`. None of that changed.

In place of the two answers the page now says that nothing picks an interrupted run up yet, and that nothing ever will without being asked. Stating a limit is what somebody asking the manual usually wants.

The answers go back on this page in the change that makes them real, and not before.

## Proof

Fresh clone of `fbbe4d400` on the test box:

| Gate | Result |
| --- | --- |
| `./internal/manual/...` (every probe) | ok, exit 0 |
| `-run Manual ./internal/tui3/` | ok, exit 0 |
| `-run Manual ./internal/session/` | ok, exit 0 |
| `./internal/namelaw/` | ok, exit 0 |
| `./internal/ci/` (the change entry) | ok, exit 0 |

Docs only. No Go source is touched.


santoshkumarradha:
Landed on santos/dev as `952e0036a`.

Gated alone on that exact commit with one full `make check` on the test box: **EXIT 0, 115 packages ok, 0 failures**. Pre-checks on the same head were clean: `gofmt -l` list empty, `go build ./...` 0, `GOOS=windows go build ./...` 0, change-entry check ok (67 entries, all well formed).

Log kept at `check-sdev-952e0036a.log`.

PR #1305, opened 2026-09-20
session: a run nothing is driving can be carried on, in the copy its work is already in

santoshkumarradha:
Fourth cell of the settled close-window design (#1224), on the copy record (#1298). One door, `Agent.ContinueRun(ctx, row)`, that picks up a run nothing is driving.

It adopts the working copy the run recorded when it started and the store already on disk, and hands both to the same engine under the same seats. The supervisor releases claims whose owner is gone and re-reads the ready set on its own next pass, so there is no resume machinery and no second scheduler: carrying on is the ordinary pass, run by a process that was not there for the last one.

## It has no caller, on purpose

Continuing seats workers and spends money. A door that spends money and is reachable before anything means to reach it is not a smaller version of the feature, it is a worse thing than no door. Nothing calls this from any surface, and a person cannot carry a run on today. The card that offers the choice is the next change, and `internal/manual/chat/how-tasks-run.md` says so plainly as of #1304.

## The refusals, and who actually meets them

| In the way | What it says |
| --- | --- |
| a run that never stopped | `<run> is already running`, or `<run> is running, and a conversation drives one run at a time` |
| a number with no run behind it | `there is no run 41 in this conversation` |
| a run that said its last word | `<run> is done, so there is nothing to carry on` |
| a copy never written down | `this run's working copy was not written down when it started, so there is nothing to carry on from` |
| a copy gone from disk | `its working copy is gone from <dir> ... its branch <branch> is still there` |

**The fourth is not a guard against a hypothetical.** The record of a run's copy landed today, so every run already on disk carries none of it. That is the whole existing population, and the first interrupted run most people meet will be one of them. What it says is the permanent answer for those rows, not a message about a version skew that passes.

That refusal is the **same sentence** the row itself shows where the offer would otherwise be, taken from one constant (`errNoRunCopy`, read by `runCannotContinue`). A test compares the door's error text to the row's reason by equality, so a second spelling cannot be written without the build saying so.

## One spec, two roads

`beltRunSpec` is extracted from `startKnownTaskRun`. A continued run is the same run, so it works under the same seats and bounds; a spec built here would drift the day somebody moved a cap on one road.

## Proof

Fresh clone of `df81736f3` on the test box.

| Step | Result |
| --- | --- |
| `gofmt -l` (the list, not the exit code) | empty |
| `go build ./...` | exit 0 |
| `go vet ./internal/session/` | exit 0 |
| the seven new tests | ok, exit 0 |
| #1296 and #1298's tests beside them | ok, exit 0 |
| `scripts/laws.sh` | exit 0, zero FAIL lines |
| `./internal/ci/` (the change entry) | ok, exit 0 |

**Three negative controls, each restored green afterwards:**

1. The door patched to reach for `prepareTaskTree`, the road that **makes** a copy, called exactly as the start road calls it. `TestCarryingOnARunKeepsEveryStepOfTheWorkInItsCopy` goes red naming the deleted file: `the work in the run's copy is gone: .../trees/7/the-work.txt: no such file or directory`. That is the control that matters. The test detects the destruction rather than merely passing.
2. The settled guard deleted. All three subtests go red: `a done run was carried on`, and the same for failed and unverified.
3. The run installed with a cut that ends nothing. `TestAStopReachesARunThatWasCarriedOn` goes red: `timed out waiting for the carried-on run to end`.

Every refusal test asserts twice: that it said no in the person's words, and that **nothing started**. A door that refuses in words and seats workers anyway has spent the money either way.

## Two laws caught this change and both are fixed in it

`scripts/laws.sh` went red on the first run and both reds were mine.

`TestEveryPublisherOfARunningRowIsOneAStopCanReach` named `ContinueRun` as publishing a running row the ledger gave no owner a stop could reach. That is the right demand: a run restarted by a door nothing can end is worse than a run nobody can start. `ContinueRun` is now in `stoppableRowPublishers` under `CancelTask`, and `TestAStopReachesARunThatWasCarriedOn` stops a carried-on run through the ordinary road and checks the workers' context was cut.

`TestNoRoadInTheTaskEngineHasMoreEndingsThanItsLedgerRow` put `taskStatusWords` at 16 decisions against a ceiling of 15, and there is no road to raising the ledger. The interrupted arm is now `taskAskContinuing`, a reader of its own beside `taskAskOf`, which is where it belonged: that one walks the facts of a landing to work out which judgement is owed, and there is no judgement here.


santoshkumarradha:
Landed on santos/dev in one batch as `9420891d1` (continue door `ad33fa1ed`, proof library `9420891d1`).

Gated together on `abfeb313c`: **113 packages ok, one failure**, `TestS3ItComesBack` in `internal/lane`.

That failure is not attributable to this batch and was not cleared by a rerun:
- `internal/lane` does not reach `internal/session` in the package or the test binary (`go list` count 0 both ways), and nothing here touches lane.
- The assertion that fired compares two wall-measured beliefs across time: 1523ms over the first twenty requests against 1733ms after thirty more. Both are inflated by loopback overhead, so uniform load cancels but *rising* load does not.
- The test's own comment four lines above records a prior load failure on this same scenario; its sibling assertion was already weakened from Order to Frontier for that reason.
- Six real passes (not skips) in isolation at 1.79-1.84s, against 2.03s inside the gate.

Red log kept as `check-sdev-abfeb313c.log.red1`. Pre-checks on the gated head: gofmt list empty, build 0, Windows build 0, 70 change entries well formed.

Code verified byte-identical after rebasing onto the current head; the intervening commits are docs only.

Issue #1306, opened 2026-09-20
Tests that cannot fail


santoshkumarradha:
Three shapes of test in this repo are structurally incapable of catching the
defect sitting next to them. Each one looks like coverage, reads like a
contract, and stays green through the bug it appears to guard. All three were
found in one night's work on the question-presentation surfaces, each beside a
real defect, and in every case the test was doing exactly what its name
promised.

This is written as a review question, not a proposal for a lint rule. See the
last section for why.

## 1. A test that pins a string

It asserts what a surface SAYS, word for word, and never asks what the surface
DOES. The assertion and the defect never meet.

Example. `internal/tui3/place_tasks_test.go:45`
`TestTheTasksFootIsScreenOneEWordForWord` pins the tasks page footer exactly.
Three more pin the same foot: `:91`, `:129`, `:204`. All four pass while the
page answers landings with digit keys that the footer never names, which breaks
the page's own law that nothing bound goes unnamed.

What the replacement asserts: press every key the page binds and assert the
footer names it, and press the keys the footer names and assert they do
something. A string test is fine as a companion to that and is not a substitute
for it.

## 2. A test that compares two booleans

It asserts that two readings AGREE and never that either is RIGHT. Both sides
can be wrong together, and the test stays green while they are.

Example. `internal/session/attach_test.go:433`
`TestNeedsPersonAgreesWithPresence` walks the lanes and pins that the agent's
`NeedsPerson()` agrees with the presence file's boolean. It never asserts that
the accompanying reason is non-empty, so it passes for a lane that says a person
is needed and says nothing about what for. That is a mark a person cannot act
on, and it survived because the only test in the area compares two booleans to
each other.

What the replacement asserts: the CONTENT each side carries, not their
agreement. For this case, that every lane which sets the boolean also produces a
sentence, and that the sentence names that lane's own question.

## 3. A test that pins one side of a pair

Two readers describe the same state and only one of them is pinned. The unpinned
side drifts, and nothing goes red.

Example. A landed task whose acceptance is still merging is described by two
readers. The record card at `internal/session/question.go:2987` says the person
already answered. The line home at `internal/session/taskpresence.go:1004` still
says `your call`, because it reads `PendingDecisions`, which filters on node
state alone. The card side is pinned five times in
`internal/session/questionlifecycle_test.go`, at `:254`, `:297`, `:328`, `:354`
and `:405`. The home side is pinned zero times. Five to nothing.

[...]

PR #1308, opened 2026-09-20
scripts: three proof steps that could not report failure, in one tested form

santoshkumarradha:
Three shapes of proof step were each written more than once in this repo's chains, and each is structurally incapable of reporting the failure it is checking for. They were found in one night, each by a different person, each in a different instrument. This is the one written form of all three, to be sourced.

| The shape | Why it cannot fail |
| --- | --- |
| `gofmt -l ./internal; echo "exit=$?"` | `gofmt -l` prints the files it would reformat and **exits 0 either way**. The list is the signal; the exit code never was. |
| `cmd 2>&1 \| tail -25; echo "exit=$?"` | `$?` is `tail`'s status, and `tail` succeeds at reading a failure. This printed `exit=0` underneath a screen of `--- FAIL`. It does not hide a red, it **manufactures a green directly below the evidence**. |
| `go test -run 'TestA\|TestB\|...'` | A `-run` pattern is a second copy of the list of tests in a file. When they drift, `go test` answers a filter matching nothing with `ok`, because from its side nothing failed. **A named test that did not run is indistinguishable from one that passed.** |

`scripts/proof.sh` gives each one a written form: `fmt_check` reads the list; `step` captures its command's status before anything else can run, keeps output in a file rather than a pipe, and counts failures from that file as a second witness sourced differently from the status; `run_tests` compares the names asked for against the tests that reported, and names every one that reported nothing.

That last point is deliberate. A guard whose satisfaction can only be shown as a tick eventually becomes a tick, so the failure is loud and specific in the same line: both counts, and the names.

## The acceptance is the failing arm

`internal/ci/prooflib_test.go` runs each function **twice**: once where the thing it checks is sound, once where it is broken. Each broken arm also pins what the plain command says about the same situation, because that is the entire argument for the function existing.

| Test | What the plain command says | What the library says |
| --- | --- | --- |
| `TestRunTestsCatchesANameThatReportedNothing` | `go test -run '^(TestAlpha\|TestGamma)$'` on a package holding only Alpha and Beta reports **ok, exit 0** | `asked=2  reported=1`, names `TestGamma`, exit 1 |
| `TestFmtCheckFailsOnAFileGofmtWouldRewrite` | `gofmt -l` lists the file and **exits 0** | prints the list, `gofmt=DIRTY`, exit 1 |
| `TestStepReportsTheCommandsOwnStatusAndNotAPipesStatus` | the pipeline reports **`exit=0`** for a command that exited 3 | `probe_exit=3`, `probe_FAIL_lines=1`, exit 3 |

A guard against invisible failure that has only ever been seen passing is the thing it was written to prevent, wearing its own name.

### Those three assertions are the library's expiry date, not belt and braces

Read them again as a set. Each one says *the hazard is still real*.

Most guards never find out when the thing they guard against gets fixed. The upstream tool changes, the hazard goes away, and the guard lives on as cargo that everybody routes around and nobody dares delete, because by every signal available no one can tell what it was for.

This one announces it. If `go test` ever stops reporting `ok` on a filter naming a test that does not exist, if `gofmt -l` ever starts exiting non-zero on a non-empty list, or if a pipeline ever starts carrying the failing command's status, the corresponding test goes red — and **that red means the library's reason for existing has gone**, not that the library is broken. The failure message in each says so in those terms: *this test is no longer about a silent pass*, *this test proves nothing*, *this test is about nothing*.

The same idea runs through the third mutation control. Putting `step`'s command back behind a pipe reproduces the original defect on demand, so the bug that motivated the library is now permanently a test of it.

## Proof

Fresh clone of `436933a6d` on the test box.

| Step | Result |
| --- | --- |
| `gofmt -l` (the list) | empty |
| `go build ./...` | exit 0 |
| `go vet ./internal/ci/` | exit 0 |
| the five new tests, verbosely | 5 PASS, 0 FAIL, exit 0 |
| whole `./internal/ci/` | ok, exit 0, 0 FAIL lines |
| `scripts/laws.sh` | exit 0, 0 FAIL lines |

**Three mutation controls on the library itself**, each restored green afterwards. Breaking the library must make its own tests red, or the tests are decoration:

1. `run_tests`' count comparison removed → `TestRunTestsCatchesANameThatReportedNothing` red: *the library agreed with the plain command and reported success*.
2. `fmt_check` made to read the exit code instead of the list → red: *an unformatted tree was reported clean*.
3. `step`'s command put back behind a pipe → red: *step answered 0, not the command's own 3*. That is the original defect, reproduced on demand.

## Scope

**No caller is converted here, and there is none to convert.** A survey for these three shapes across `Makefile`, `scripts/` and `.github/` found zero instances: `make fmt-check` was already written in the correct form, no shipped chain reports a status after a pipe, and no `-run` pattern lists named tests that could drift from a file. Every instance was in an ad-hoc chain script written outside the repository.

That count covers those three shapes in those three directories and claims nothing about any other way a failure can be discarded, and nothing about Go source, which was not looked at. The scope is stated in the entry and in the file header for the same reason everything else here is written twice-proof: a finding whose boundary travels separately from it becomes a claim nobody tested. The survey and what the zero means are in the issue below.

**So these functions may have no caller, and that is not a reason to delete the file.** Its standing value is the test: three assertions about how the tooling behaves that our proof chains depend on, which run on every pull request whether or not a line of it is ever sourced.

Nothing a person using this program can see changes, so no manual page is owed. The test-shaped members of the same family, the review question, and the survey with its scope are in the issue titled "Tests that cannot fail" (#1306).


santoshkumarradha:
## A real instance of the third shape, found after this was written

Tonight, on the branch now open as #1315. Worth recording because it was found
by accident afterwards rather than used to design the guard, so it is evidence
rather than illustration.

An automated run declared its verification as:

```
declared check `go test ./internal/session -run TestPresenceAskReason` exits 0
```

No test of that name exists in `internal/session`. Verified twice, independently,
in the run's own tree: nothing matches, and nothing shares the prefix. The tests
that run actually wrote are `TestWaitingSentenceConnectSignIn`,
`TestWaitingSentenceHarnessAsk` and `TestWaitingSentenceWrongOverlap`.

Nothing lied and nothing failed. `go test` answers a filter matching nothing
with `ok` and exits 0, so the run reported its result faithfully. The check
simply could not fail on behaviour; the only thing it could have caught is the
package failing to compile.

`run_tests` catches exactly this, and only this: it compares the names asked for
against the names that reported, and returns non-zero naming the one that
reported nothing. So this proof step would have failed the moment it ran, and
the false green would have been visible without anybody reading the diff.

For scale, what it took to catch it the other way: a person read the branch's
diff line by line, noticed the test name in the report did not match the test
names in the file, and grepped the package to confirm. That is the only reason
it is known. Nothing in the automated chain objected, and the harness that
judged the run recorded it as clean.

The shape catalogue this belongs to is #1306; this is its third shape with a
name and a date on it.

santoshkumarradha:
Landed on santos/dev in one batch as `9420891d1` (continue door `ad33fa1ed`, proof library `9420891d1`).

Gated together on `abfeb313c`: **113 packages ok, one failure**, `TestS3ItComesBack` in `internal/lane`.

That failure is not attributable to this batch and was not cleared by a rerun:
- `internal/lane` does not reach `internal/session` in the package or the test binary (`go list` count 0 both ways), and nothing here touches lane.
- The assertion that fired compares two wall-measured beliefs across time: 1523ms over the first twenty requests against 1733ms after thirty more. Both are inflated by loopback overhead, so uniform load cancels but *rising* load does not.
- The test's own comment four lines above records a prior load failure on this same scenario; its sibling assertion was already weakened from Order to Frontier for that reason.
- Six real passes (not skips) in isolation at 1.79-1.84s, against 2.03s inside the gate.

Red log kept as `check-sdev-abfeb313c.log.red1`. Pre-checks on the gated head: gofmt list empty, build 0, Windows build 0, 70 change entries well formed.

Code verified byte-identical after rebasing onto the current head; the intervening commits are docs only.

PR #1314, opened 2026-09-20
tui3: a standing row stopped for permission says the line, not asks: it

santoshkumarradha:
A watch that stopped because a call needed permission drew a row reading

```
asks: stopped: it needed your ok to run bash
```

Nobody asked anything. `standing.Item.NeedsPerson` carries two different things,
and its own doc says so: a QUESTION a firing put to a person in its own words,
or the line this build writes when a call was refused for want of somebody to
allow it. `switcherStandingNote` put the ask word over both.

The fix asks the one predicate that knows the difference,
`standing.IsPermissionLine`, and says the refusal bare, because it is already a
whole sentence about the item. The ask word stays for the other tenant.

This is the reading three siblings already take and this surface skipped:
`homepanel_needs.go` picks a different door word for a permission line,
`homestanding.go` picks a different destination, and
`standing.ClearNeedsPerson` will only clear the permission tenant. No second
list of spellings is added here; the spellings live in `standing`, in that one
predicate.

It also follows the precedent set a few lines above it in the same function:
`switcherConversationNote` already returns the consent gate's own sentence bare,
for the same reason and with a comment saying so.

## The test, and its failure on base

New: `TestAWatchStoppedForPermissionIsSaidNotAsked` in
`internal/tui3/switcher_test.go`, covering both tenants, so the word is proved
present for a question and absent for a refusal.

On base it fails with the defect in its own words:

```
switcher_test.go:124: the row puts the ask word over a line nobody asked:
got "asks: stopped: it needed your ok to run ..."
```

There is none for this note in the permission state. The nearest test covers the
switcher's ranking and passes regardless of which tenant the line holds, because
it never reads the note's text in that state. Recorded here rather than left
implicit: this change adds the first assertion about what that row says when
nobody asked.

## Scope

One surface word. The mark, the glyph and the count are deliberately untouched:
whether a stopped-but-nothing-to-answer state should carry its own quieter mark,
and where its door goes, is a separate decision that has not been made. This
change does not prejudge it.

## Checks

`go test ./internal/tui3`, the whole package, ok in 340.670s, run on the base
the work was written against (`3d3b7542b`). This branch is that change
cherry-picked onto the current head (`952e0036a`), so the gate's run on this
head is the one that counts, and I am not quoting the earlier run as if it were
taken here.


santoshkumarradha:

PR #1315, opened 2026-09-20
session: a sign-in and a harness ask say what they are waiting for

santoshkumarradha:
Two lanes said a person was needed and said nothing about what for.

`personAskLanes` decides THAT a person is needed by OR-ing the lane books.
`waitingOnPerson` then takes the SENTENCE from `presenceAsk().Text`, which walks
the presence desk. Mark and sentence come from two different books, so a lane
counted in the OR that banks no desk row leaves `waiting: true, reason: ""`, and
a surface draws a mark nobody can act on.

Sign-in and the harness lane were both in that state. This routes them through
`presenceAskingWhole`, which is `raiseQuestion` plus the row, and returns one
closure that undoes both.

Three call sites, two conditions: `connect.go` for `connectAsks`, and
`harness.go` plus `harness_build.go`, which write the same `harnessAsks` book,
so the offer and the design card are one condition and one fix.

## What is deliberately NOT converted

Seven call sites raise through `raiseQuestion` and none of them banks a desk
row, so "banks a desk row" is the wrong criterion and would have converted all
seven. THE CRITERION IS WHETHER THE ROW ENDS UP WITH WORDS. One of the seven is
the door inside `presenceAskingWhole` itself and is not a lane. Of the remaining
six, three already get their words from another arm of `waitingOnPerson`: the
sub-harness offer, the fuel gate, and landings. Those three are untouched.

`task_landing_question.go` in particular is left alone: it drops its let-go
deliberately, because settle and answer own that question's retirement, and
banking a row there would leave nobody holding its remover.

## The failure is not always an empty sentence

When no lane has banked a row the reason is empty. When ANOTHER lane has one,
`presenceAsk()` returns that lane's words, so the row says a person is needed
and then describes a different question. Both cases are covered:
`TestWaitingSentenceConnectSignIn`, `TestWaitingSentenceHarnessAsk` and
`TestWaitingSentenceWrongOverlap`, the last being the wrong-sentence case rather
than the empty one.

`TestNeedsPersonAgreesWithPresence` in `internal/session/attach_test.go` walks
the lanes and pins that `Agent.NeedsPerson()` agrees with the presence file's
boolean. It never asserts the reason is non-empty, so it passes for a lane that
says a person is needed and says nothing about what for. A test that compares
two booleans asserts that the two agree, never that either is right.

## Two comments, and who wrote them

The third commit is mine, not the change's author's. Both converted sites
carried a comment naming `question.go`'s `raiseQuestion` as the door, which
became false at those lines the moment they were converted. They now name
`presenceAskingWhole` and say what it adds. A doc comment that describes the
call above it incorrectly is worse than no comment, because the next reader
trusts it.

## Checks

The change was written against `3d3b7542b` and is cherry-picked here onto
`952e0036a`.

Stated plainly because it bears on how much this body is worth: THE CELL THAT
WROTE THIS DID NOT RUN THE DECLARED CHECK. Its report quotes
`go test ./internal/session -run TestPresenceAskReason` as passing, and no test
of that name exists in the package; a `-run` pattern matching nothing cannot
fail on behaviour. The whole-package check the work was briefed with was never
run. So the verification behind this branch is the diff, read line by line, plus
the gate on this head. Nothing here rests on the cell's own green.

## What would have objected

Naming it, because "the declared check was never run" says a hole existed and
not what closes it. The form that catches this is `run_tests` in `scripts/proof.sh`
on #1308, which runs the named tests and then checks that every name it asked
for actually RAN. Its own failing arm is in `internal/ci/prooflib_test.go`.

Against this branch it would have objected in the same second the step ran, and
it would have named the test that reported nothing. The point is not that
anything lied. `go test` reported its result faithfully: a filter matching
nothing is a run with nothing in it, and `ok` is the honest answer to that. What
the step could ever have caught was compilation and nothing else. A cell can
declare a check, run it, get a true green, and have proved that the package
builds.

Catching it the other way cost a diff read line by line, a name mismatch
noticed, and a grep to confirm.


santoshkumarradha:

PR #1316, opened 2026-09-20
tui3: the front tab and the tab beside it agree about one conversation

santoshkumarradha:
The front tab and the tab beside it answered the same question differently about
the same conversation.

A held tab's mark comes from `behindWatch.signal`, whose cached fact is fed by
`keeper.go`'s `needsPerson`, which calls the engine's `Agent.NeedsPerson`. So a
held tab counts every lane the engine counts.

The front tab's mark comes from `frontSignal`, which re-spells the lane list on
the surface. Two lanes the engine counts were missing from it: standing answers,
and landed tasks saying `your call`. So a conversation could carry the mark
while held and LOSE IT when a person brought it forward, which is the opposite
of what the mark is for.

## Answered from the surface, as the file requires

`frontSignal` answering from the surface rather than the engine is deliberate
and documented: `Agent.NeedsPerson` takes the agent's mutex and allocates a map,
the strip is laid out every frame, and over a remote handle the far side would
be a call per tab per frame. That is unchanged here.

The new reading is `asksStandingOrLanding`, one allocation-free walk over
`a.questions`, the same slice and the same shape as `asksHarness` and
`consentOpen` already use. No lock, no disk, no engine call.

## The exclusion is kept

An automatic proposal carrying a deadline is a countdown, not a question, and
`frontSignal` already excluded it. The new predicate names the two kinds it
counts rather than treating every question as a signal, so the countdown stays
out on both sides, and its doc comment says why.

## The test

`TestTheSameConversationAgreesInFrontAndBehind` puts the SAME conversation in
front and behind and compares the two answers. It carries its control in the
same test: a deadline proposal must also agree, and agree at `tabIdle`, so the
test would fail if the fix had widened the predicate into counting countdowns.

`tabsignal_test.go` pins `tabNeedsPerson` for held tabs and for front tabs
separately, and `hopsignal_test.go` asserts the hop card and the tab signal
agree. None of them puts one conversation on both sides and compares, which is
the only shape that fails for this defect. This adds that shape.

## Checks

`go test ./internal/tui3`, the whole package, run four times over the work at
330.5s, 332.7s, 334.3s and 335.3s, on the base the work was written against
(`3d3b7542b`), plus focused runs of the new test. This branch is that work
cherry-picked onto the current head, so the gate's run here is the one that
counts.


santoshkumarradha:
## Amended: the landing arm is dropped, the standing arm stays

Pushed as `2b226e4f4`. The predicate is now `asksStanding` and counts one kind.

**Why.** A landing's question object stays open through the WHOLE SETTLE after
the person has answered it. `publishLandingQuestion` retires it only when the
node leaves `TaskUnverified`, and an accept does not move the state until the
merge finishes; its own comment says a notice naming a resolution in flight
"raises nothing and withdraws nothing". So counting landings here held a mark up
over an answer already given, which is the one thing a mark must never do.

**Why not a clause instead.** The fact that separates a live landing from a
settling one is not on this surface in any usable shape. The only place it
exists is `landingQuestion`, which writes it into `q.Reason` as the words
`still working on it`, and even that branch does not run in the ordinary case:
`landingQuestion` ends `return a.said(...)`, and `said` returns BANKED words
untouched, which every landing raised through `publishLandingQuestion` has. A
predicate keyed on that sentence would pass every test written for it, survive
every refactor that kept the words, and break silently the first time somebody
improved the sentence. That is not a fix, so this lane says nothing rather than
something stale.

**What it costs, stated plainly.** A live unanswered landing on the FRONT
conversation now draws no tab mark, while a held tab would mark it. That is a
real disagreement in the other direction, and it is the trade taken knowingly:

- it is not lost from the screen. `questionDrawnHere` returns true for
  `QuestionLanding`, so it is drawn in the conversation itself, and its card in
  the transcript says how the work came home and what is being asked. The
  person on this tab is already looking at it.
- the front tab is the weakest mark in the product, because the content it
  points at is on screen. Marks earn their place on tabs you are not on.
- and the two errors are not symmetric. Under-marking loses a notification.
  Over-marking trains a person to distrust the mark, and once a mark is noise
  you cannot make it a signal again by fixing the predicate later.

The engine-side fix that would let both lanes read this properly is separate
work: carry the answered-and-settling fact on the question as a FIELD, set where
the stamp is set, and refresh the bank so the new shape reaches surfaces.

The test is updated to drive the standing lane and keeps its control unchanged:
a deadline proposal must still agree on both sides, and agree at idle.

santoshkumarradha:

PR #1322, opened 2026-09-20
session: a landing already accepted stops saying your call

santoshkumarradha:
A landed task whose accept is already in flight kept asking the person to accept
it.

`waitingOnPerson`'s landing arm took `PendingDecisions()[0]` unconditionally and
named it. `PendingDecisions` filters on node state alone, and an accept does not
move the state until the merge finishes, so the entry stays on the list through
the whole settle. The record card already knew better: `question.go` reads the
decision record and says `accepted · still working on it`. Home did not, so the
two readers of one state disagreed, and the person was prodded for a question
they had answered.

## It is a selection, not a guard

The obvious fix is a guard: skip the entry if its answer is in. That is wrong in
a way that passes its own test.

`graph.order` is admission order and the arm took the FIRST entry. With two open
landings, one answered and one not, a guard on `pending[0]` goes quiet in the
all-answered case and still names the ANSWERED landing while the unanswered one
is the reason the row is up. The change takes the first entry that qualifies
instead, and contributes nothing when none do:

```go
for _, pending := range a.PendingDecisions() {
	if strings.TrimSpace(pending.Notice.Settling) != "" {
		continue
	}
	return personAsk{waiting: true, reason: yourCallLine + strings.TrimSpace(pending.Notice.Title)}
}
return personAsk{}
```

`PendingDecisions` itself is untouched. Its membership is state in, state out,
and other callers depend on that, so the skip belongs at the arm that draws the
line.

## The condition is non-empty, not equality

`resolutionInFlightLocked` (`task_run.go`) returns two different strings:
`n.settling`, which a settle claims in plain words such as `your accept`, and
the literal `a merge round` when only the `resolving` flag is set. A test for
the accept spelling would fix half of this, and the missed half is
indistinguishable on screen. Both mean the same thing to the person: the work
proceeds without them.

## The test that matters is the second one

`TestAnAcceptedLandingInFlightDoesNotSayYourCall` is the all-answered case, and
it is the test anybody writes first. A guard on `pending[0]` passes it.

`TestTheYourCallRowNamesTheLandingStillWaiting` is the one that separates the
two shapes. It puts a SECOND landing beside a settling one and asserts which of
them the row names. A fixture with one candidate cannot tell a correct selection
from a lucky one, and a two-candidate fixture that still checks the outcome
rather than the choice is a one-candidate fixture with extra scenery.

Its negative clause is live, which was checked rather than assumed: the settling
node's title really is `Add the guard`, from `unverifiedNode` at
`internal/session/task_resolve_test.go:38`. An assertion that something is
ABSENT passes forever unless the absent thing exists in the fixture.

The card side of this pair is pinned five times in
`internal/session/questionlifecycle_test.go`, at `:254`, `:297`, `:328`, `:354`
and `:405`. The home side was pinned zero times. Five to nothing, and the two
readers drifted apart while every test stayed green.

## What is deliberately NOT here, and why

There is no three-state test, red on base and red with either half alone. The
brief this came from carried that as a requirement, on the premise that the fix
needed a new field on the Question plus a refresh of the question bank, because
the settled-and-settling fact was believed to be unavailable to any surface.

That premise is wrong, and reading settled it. `PendingDecisions` builds each
entry from `noticeLocked`, and `noticeLocked` sets `Settling` from
`resolutionInFlightLocked`, so the fact is in the arm's hand at the line that
was defective. Nothing needed plumbing. A fix that is not two parts cannot
produce a middle red, so the requirement stopped applying rather than being
skipped.

The bank findings were real and they belong to a DIFFERENT reader: the words
path, `landingQuestion` through `said` into the question bank, which is what
renders a question's own sentence. That is the front tab's problem and is not
touched here.

## Checks

Written against `3d3b7542b` and cherry-picked onto `32b4127bd`. Committed by the
author of the work with the author field corrected to the bot; the change note
is mine.

`gofmt -l ./cmd ./internal` is empty. `go test ./internal/session` is running
now through `scripts/one-suite.sh`, under the box's heavy-suite lock, on this
head. Its result is posted to this thread when it reports, and nothing here
rests on it until then.


santoshkumarradha:
Landed on santos/dev as `9b38c2dda`.

Gated alone on that exact commit: **EXIT 0, 116 packages ok, 0 failures**. Pre-checks clean: `gofmt -l` list empty, `go build ./...` 0, `GOOS=windows go build ./...` 0, 78 change entries well formed.

Change verified byte-identical after squashing onto the current head.

Log: `check-sdev-9b38c2dda.log`.

santoshkumarradha:
Declared check, on the record, including the run that was spoiled.

Clean run, on the pushed head `f08536cad`, through `scripts/one-suite.sh` under the box's heavy-suite lock:

```
ok  	github.com/Agent-Field/codeaf/internal/session	260.996s
```

THE FIRST RUN OF THIS CHECK REPORTED FAIL AND I CAUSED IT. The tests passed and the suite still failed, because I committed the change note into the same checkout while the suite was running in it. The guard that watches for that is correct and it caught me:

```
session tests changed the checkout they were running in (/home/santosh/src/c374-land):
it stood on 43e95a26f8ff and now stands on f08536cadf59; what was committed on top of it:
	f08536cad docs: change note for the landing your-call fix
```

Recording it rather than posting only the clean rerun, because a check that was run twice and quoted once is a check nobody can audit. Nothing about the change moved between the two runs: same commit, same tree, and the second ran in a tree nobody touched.

`gofmt -l ./cmd ./internal` empty, `codeaf-changes check` green at 78 entries, zero em dashes in the added lines.

Issue #1307, opened 2026-09-20
Two locks, neither exclusive


santoshkumarradha:
The heavy-suite lock is two locks. A tree at or past #1264 takes a flock on
`/tmp/codeaf-suite-<uid>.lockfile`. A tree behind #1264 takes the older
directory lock at `/tmp/codeaf-suite-<uid>.lock`. Neither mechanism can see the
other, so the flock's exclusivity is at present a CLAIM THE BOX DOES NOT HONOUR.

## Observed on the box, 2026-09-20T18:35:25Z

A heavy suite running while the flock reads free:

- pid 4029204, `bash ./scripts/one-suite.sh go test -timeout 15m -count=1
  -count=1 -p 1 ./cmd/codeaf ./internal/tui3`, cwd
  `/home/santosh/src/cw0918-adaptive-folders`, alive.
- pid 4029208, its `go test`, alive.
- That tree is at `991a83d384f5197e99471bb05b4ac8a4d3c1b07e`, behind #1264. Its
  `scripts/one-suite.sh:31` reads `lock="/tmp/codeaf-suite-$(id -u).lock"`, with
  the comment explaining that the lock is a directory because mkdir is atomic.
- `/tmp/codeaf-suite-1001.lock` exists, created 14:29, and its `pid` file
  contains `4029204`, the live process.
- At the same instant `flock -n /tmp/codeaf-suite-1001.lockfile -c true` returns
  0, free, `lsof` on that file returns no holder, and the file's contents are a
  stale line from an earlier run at 18:12:10Z.

That run is behaving CORRECTLY by its own rules. It holds a lock. It is simply
holding one that current readers do not read.

## The failure is symmetric, and only one direction has been hit

DIRECTION ONE, live tonight. A stale tree holds the directory lock. A current
reader checks only the flock, sees free, and starts a second heavy suite beside
it. This is the direction that has been costing us unattributable reds and
timing numbers taken under contention nobody recorded.

DIRECTION TWO, not yet hit and worse. A current tree holds the flock. A stale
tree's runner checks only the directory, sees no holder, and starts beside it.
Same collision, opposite parties, and here the party clobbered is the one that
did everything right.

Dual-checking on the READ side fixes direction one only. It stops a current
reader starting beside a stale holder. It does nothing to stop a stale runner
starting beside a current holder, because that runner will never be taught to
read the flock: being unaware of the flock is what makes it stale.

## Immediate remedy, for anything that reads the lock

[...]

PR #1324, opened 2026-09-20
scripts: a heavy suite takes both locks so a stale tree can see it

santoshkumarradha:
The heavy-suite lock is two mechanisms that cannot see each other, so its
exclusivity is a claim the box does not honour (#1307). A checkout at or past
#1264 takes an flock on `/tmp/codeaf-suite-<uid>.lockfile`. A checkout behind it
takes the older directory lock at `/tmp/codeaf-suite-<uid>.lock`. Two heavy
suites run side by side and each believes it has the box to itself.

Dual-checking on the READ side closes one direction only. It stops us starting
beside a stale holder; it cannot stop a stale runner starting beside us, because
never having heard of the flock is precisely what makes it stale, and it will
never be taught to read one. Taking BOTH locks closes the other direction
without changing a single old checkout: a current tree becomes visible to every
stale reader still out there, and the remaining blindness runs in the safe
direction only.

## Three decisions, each of which could have gone the other way

**The directory is CLAIMED first and the refusal is REPORTED last.** The claim
comes first so there is no window in which a current tree holds the flock and is
still invisible to a stale runner. The refusal is deferred until the flock has
had its say because the two collisions are not equally well understood: two
CURRENT trees collide on the flock, and that refusal knows whether the recorded
pid is visible from this pid namespace and what to run when it is not. Reporting
the directory first replaced that diagnosis with a cruder one for the common
case in order to describe a rarer one. The first draft did exactly that and the
existing acceptance arms caught it.

**The directory is freed by whoever holds the flock, which is the HOLDER beside
the suite and not this wrapper.** The flock's lifetime is deliberately the
holder's, and killing the wrapper is explicitly permitted: it must not unlock a
box that is still running a suite. A directory lock released on the wrapper's
schedule would open the box to a stale reader while the flock was still held,
which is the two readers of one state disagreeing, again. The holder inherits
the environment, so it knows the lock by the same name and drops it when the
suite ends.

**The two paths are named independently in the configuration.** Deriving one by
trimming `file` off the other would make two names one fact, and a fact spelled
twice can disagree with itself. It would also follow a test harness driving a
private lock straight to the real one, silently. An empty setting means no
directory lock at all, which is what a platform that never had one wants.

## The test, and the arm that matters

`scripts/one-suite_test.sh` gains two functions and every existing arm is pinned
to a private directory lock, without which the suite would have taken the real
one on whatever box it ran on.

- `run_both_locks`. A held suite is visible to a reader of the DIRECTORY
  mechanism, which is the only road a stale checkout has; it is still visible to
  a reader of the FLOCK, unchanged; and RELEASE DROPS BOTH.
- `run_refused_by_directory`. A stale holder turns a current run away, which is
  the direction that has been costing unattributable reds. It also asserts that
  a refused run leaves the lock exactly where it found it, since it never held
  it.

The release assertion is the one that matters and the one that was awkward to
construct. A half release leaves the box locked to one of the two populations
forever, which is worse than the defect being fixed, because the defect at least
fails open.

### Both new arms were shown to fail, by mutation

Not asserted, run. Each mutation was applied, the arms run against it, and the
code restored:

| mutation | what the arm said |
| --- | --- |
| the holder no longer drops the directory lock | `the directory lock /tmp/.../both.lock.dir outlived the suite that took it` |
| `takeDirLock` reports success without creating anything | `a held suite did not take the directory lock /tmp/.../both.lock.dir` |

Unmutated, the same run prints `one-suite namespace and dual-lock acceptance: ok`.

## One defect found and fixed on the way

The first draft wrote the directory's `pid` file between the suite starting and
the flock being named. The existing arms read the lock file as soon as the suite
reports it is up, so that extra write widened a pre-existing window and they
began reading an empty lock. The pid file is now written after the flock is
named, and the arm that reads it waits for it rather than reading once and
calling an empty file a defect.

## Checks

`bash scripts/one-suite_test.sh` passes, five arms.
`go test ./cmd/codeaf-suite-lock ./internal/guard ./internal/namelaw` all ok.
`go vet ./cmd/codeaf-suite-lock` clean, `gofmt -l ./cmd ./internal` empty.
`GOOS=windows go build ./cmd/codeaf-suite-lock` builds: that platform has no
holder, so the wrapper keeps the directory lock and frees it itself.

Written against `9b38c2dda`.

Refs #1307.


santoshkumarradha:
Landed on santos/dev as `bc116d35c`.

Gated on that exact head: **EXIT 0, 116 packages ok, 0 failures**. Pre-checks clean: gofmt list empty, build 0, Windows build 0, 80 change entries well formed, and `scripts/laws.sh` exit 0 with zero FAIL lines — run *before* the gate this time, which is the twenty-second check that would have saved the previous run.

The earlier gate on `4f0d60f68` was **red** and is kept as `check-sdev-4f0d60f68.log.red1`: two environment laws naming `dirlock.go` by line, and two of this package's own tests failing on EOF because the exported lock path was inherited by the suite, which then refused its own children. Both causes fixed here rather than worked around.

Also folded in: the change entry from #1325 was missing its `pr:` field and did not carry the required `<pr>-<slug>.md` filename, so trunk was failing its own change-entry check independently of this work. Corrected in its own commit, content untouched.

Issue #1328, opened 2026-09-21
a failed landing with its work kept asks forever, long after nothing is waiting on the answer


santoshkumarradha:
## What happens

A task that ends `failed` with its work still held keeps wearing `your call` and keeps counting toward `needs you`, for the rest of the session, with no way to reach zero except answering each one.

`internal/session/task_status.go:441`:

```go
case TaskFailed:
	if facts.Held {
		// THE ANSWER IS STILL THERE AND SOMEBODY MAY STILL WANT IT...
		status.Presence, status.On = TaskPresenceNeedsLook, TaskWaitPerson
		return status
	}
```

The offer is real — the branch is on disk and taking it anyway is a genuine answer. The problem is that the offer never expires.

## What it looks like

One session, 23 nodes. Four rows stand in `needs you`, all `state=failed`, `ending=refused`, `merge=aborted`, each with a kept branch that exists:

| node | title | branch |
|---|---|---|
| 4 | Skills: five-part record and store serving | `task/skills-five-part-record-and-stor-…` |
| 6 | Skills: chat shelf and use_skill tool | `task/skills-chat-shelf-and-use-skill-…` |
| 7 | Store skill fact fields | `task/store-skill-fact-fields-…` |
| 15 | File GitHub issue: standing firing sees empty task board | `task/file-github-issue-standing-firin-…` |

By the time they were read, the conversation had moved two tranches past them: a later node had already landed a continuation of the same store-record branch, none of the four branches had ever been pushed, and nothing in flight was waiting on any of them. The count said four people-shaped questions. Zero of them were blocking anything.

## Why it is wrong

A mark exists only when a person's answer is the one thing unblocking the work. A card may keep showing a state; a mark must not. A mark is a demand and a card is a status. Four standing demands that unblock nothing train a person to stop reading the count, which costs the marks that do matter.

## Shape of the fix

A held landing is a question while the conversation is still on it, and a status afterwards. When the work has been superseded, or the session has moved past it, the row keeps its branch and its offer on the card and stops wearing the mark and stops counting toward `needs you`.

Open question for whoever takes it: what "moved past it" should be measured against. Candidates seen here — a later node that landed the same branch, a node whose branch no longer exists, and plain distance in the graph. The first two are facts already on the record; the third is a guess. Pick from the record, not the clock.

## Not this

[...]

PR #1331, opened 2026-09-21
session: a held landing offers its answer and stops demanding one

santoshkumarradha:
A task that ended `failed` with its work still held wore `your call` and counted
toward `needs you` for the rest of the session. Nothing was waiting on any of
those answers.

Closes #1328.

## The engine already says this, in two places the reading disagreed with

`internal/session/task_landing_question.go` raises a landing question for
`TaskUnverified` and retires it for every other state. A failed node therefore
has no question object and never had one: no turn is parked on it, and no
conversation is holding it.

`Agent.handToModelOnAuto` writes a decider only for `TaskUnverified`, and its
comment says why in terms that settle this one:

> AND IT IS ONLY EVER ABOUT THE ONE STATE THAT HAS A DECISION IN IT. A node that
> landed done, incomplete or stopped is not waiting on anybody's word, and
> writing an owner onto it would invent a question nobody is asking.

`taskLifecycleStatus` gave that same node `TaskPresenceNeedsLook`, which becomes
`Attention`, which is a demand. Two readers of one state, and the fix is to make
the reading agree with the law already written rather than to add a third
opinion.

THE ROW STOPS ASSERTING A WAITER THAT DOES NOT EXIST.

## What changed, and what deliberately did not

`taskStatusDemand` now takes the facts, and one predicate turns `Attention` off
for a failed held landing a person did not stop.

The card is untouched, and that is the point. The tier, the word, the reason and
both answers all stay: the branch is on disk, taking it anyway is a real thing
to do, and a person who goes looking at the row is owed the offer. What goes is
only `TaskStatus.Attention`, which is what files a row under `needs you`.

It is the same separation, for the same reason, that `internal/tui3`'s
`taskstatus.go` already makes for a question somebody else is holding: "The
tier, the word and the question are untouched: the row still wears the `?` and
still reads its reason, because a person watching it is owed both."

No new field, no owner written, no clock, no count, no graph walk, no disk read.

## It does not delete the demand

It moves it back to whoever has one. If something in a conversation ever does
need this answer to carry on, it ASKS, and an ask raises the mark through its
own lane in `Agent.waitingOnPerson` with no help from this row. A held landing
will not wear the mark as the engine stands, because nothing ever waits on one
today; wire a real park onto that answer and it marks again, through the ask
lane, with no further change here.

## Three candidate signals that were rejected, with the reason

The issue asks what "moved past it" should be measured against. None of the
three it names can answer it.

A LATER NODE THAT LANDED THE SAME BRANCH cannot happen. A branch is
`task/<slug>-<shortID>` and `shortID` is three bytes of `crypto/rand`, so no two
nodes ever share a branch name. A predicate on it could never be true and a test
for it could never fail.

A CONTINUATION LEAVES NO SECOND NODE. `task_continue.go` re-arms the node in
place: same id, same spec, same tree. A new proposal is explicitly a different
piece of work, with no recorded link. There is no supersession edge to read.

A BRANCH THAT NO LONGER EXISTS needs a disk probe on the road that draws every
card, which is exactly a claim the notice does not carry.

A static `dependsOn` edge was considered and rejected too: it says what COULD
wait, not what IS waiting, so it would still raise a mark at the moment of
failure for any task with a dependent.

## Acceptance

New file `internal/session/task_held_mark_test.go`.

On `e87b147c9` as it stands:

```
--- FAIL: TestAHeldLandingOffersTheAnswerAndDemandsNothing
    a held landing stands in `needs you` with nothing waiting on the answer
```

With the change, `ok`. The red is the base's: `task_status.go` was checked out
to the index to produce it and then restored, rather than the change being
mutated.

THE TEST HAS CONTROLS, because a test that only asserts nothing demands anything
would pass on a reading that stopped demanding everything.
`TestTheRowsSomethingIsWaitingOnStillDemand` pins three rows that still set
`Attention`, each with something genuinely on the other side of the answer: a
conflict, which is two versions of the person's own files and only they can say
which survives; a landing nobody could check; and work nothing is driving. A
third test pins that a person's own stop is OVER and grows no question, since
both settle `failed` and the new predicate must not be what keeps a stopped row
quiet.

The existing held-landing cases in `task_states_test.go` pin the card for
exactly these facts, tier `your call`, `accept anyway` and `not right`, owner
person. They still pass, which is the evidence the card survived rather than an
assertion that it did.

## Checks

Written against `e87b147c9`, and the red above was re-verified on that head
rather than carried over from the earlier base.

```
ok  	github.com/Agent-Field/codeaf/internal/session	252.706s
ok  	github.com/Agent-Field/codeaf/internal/tui3	331.603s
```

Both under the box's one heavy-suite lock, in one run, with `-count=1`.
`internal/tui3` is in it because that is where `Attention` is read, and a change
to the mark checked only in `internal/session` is checked in the wrong place.

`bash scripts/laws.sh` exits 0. `gofmt -l ./cmd ./internal` prints nothing.


santoshkumarradha:
Landed on `santos/dev2` as `d39a4e308`, hand-squashed onto `e87b147c9` and verified byte-identical to this branch.

Full check on that head, fresh clone: EXIT 0, 116 packages ok, 0 failures, 0 cached. `santos/dev` was fast-forwarded to the same sha.

Closing rather than merging because the squash is already in.

PR #1355, opened 2026-09-22
session, run, cmd: the worker harness is the default belt again, with no worker bound

santoshkumarradha:
Restores the #1335 end state that #1340 reverted, and removes the worker bound the run engine was quietly applying.

## The default

Unset `CODEAF_TASK_BELT` is the worker harness again. `node`, `legacy` and `off` reach the older engine; anything else, including an empty value, is the harness. Every test file that #1340 touched goes back to its #1335 spelling, including the one-line `CODEAF_TASK_BELT=node` pin in `internal/session`'s and `cmd/codeaf`'s `TestMain` that names the older engine's suite once instead of a hundred and forty times, and `bashbelt_default_test.go` asserts the default in a truth table.

#1340 moved the default back on an eleven-run-per-arm comparison whose cells were launched in two unequal waves (four, then sixteen; load 2.5 then 38), so its wall times measured the launch schedule and its quality gap was one task. The comparison that replaces it is a 32-cell shuffled single batch on Spark (2026-09-21), eight DeepSWE tasks, two replicates, both arms interleaved so they saw the same load, `z-ai/glm-5.3-flash` on every seat, each cell graded by its fixture's own suite with the official test patch applied:

| arm | scored | passed | median cost | median wall |
| harness (bash) | 14 | 3 (21%) | $0.239 | 1010 s |
| node | 13 | 2 (15%) | $0.384 | 2924 s |

Pooled over every batch run this week (61 scored cells): harness 22% at $0.310 median, node 23% at $0.376. The harness is cheaper and roughly three times faster at the same quality; the wall gap is the node belt running its parts one after another. Crew arms with a separate plan seat and the flat single-worker arm were run beside these and were dominated on cost at equal or lower quality.

## The bound

`internal/config`'s `task.parallel` is 0 out of the box and its own hint says 0 is no limit. The chat door hands that 0 to the run engine, and `run.NewSupervisor` read it as 1, so every task a conversation put on the harness ran one part at a time. `codeaf do` did not read the setting at all; it carried a constant of four. This PR:

- makes a slot count below one no bound in the supervisor, with `full()` as the one place the bound is read;
- has `codeaf do` resolve its slot count from `task.parallel` like the chat door, with `--slots <n>` to name one for a run (`0` is no limit; blank is the setting), documented in `--help`, the `codeaf do` flag table and *How many tasks run at once*;
- has `drain` read the returns before waiting on the goroutines, because an unbounded run can have more workers out than the return channel is deep and a worker blocked on handing in its return never reaches `Done`.

Tests: `TestSupervisorLaunchesEveryReadyLeafAtOnceWithNoSlotBound` (six leaves on the seat together), `TestAnUnboundedRunEndedEarlyStillDrainsEveryWorker` (forty in flight, cost limit crossed by two, Run must come back under its own wall), `TestCodeafDoRunsAsManyWorkersAsTaskParallelSays`, `TestTheSlotsFlagTellsBlankFromZeroAndRefusesTheRest`.

## The tmux suite names its belt

`start()` launches every scenario with `CODEAF_TASK_BELT=node`, the road each was written for, and `startWithEnv` runs the child under `env -u CODEAF_TASK_BELT …` so a word exported in the runner's shell cannot choose the road; an assignment in a scenario's own env follows the `-u` and still wins. `TaskOnTheDefaultBelt` is new: it launches with the variable absent and reads the run road off the tasks place, the plan page and the thread, which is the only test of the default in the tree. The comment over `testTaskOnTheRunEngine` and the words table say what is true now. The tagged package compiles (`go vet -tags e2e ./internal/e2e/`); it has not been driven against a model on this branch.

## Proof

All on Spark, worktree `~/af-bash-restore` at `84e7d25f4` (this branch's tip), `BASE=origin/santos/dev2` (`8a83f4132`). Nothing ran on the laptop.

- `make pr-ready BASE=origin/santos/dev2` at `84e7d25f4`, pid 2148478, log `/tmp/bash-restore-prready3.log`: the whole quick gate green (build, vet, fmt, packed manual, change entries, manual gates, 31 law packages ok); `test-touched` refused the box lock because another session's `go test ./...` held it.
- `make test-touched BASE=origin/santos/dev2` at `84e7d25f4` once the lock freed, pid 2229868, log `/tmp/bash-restore-touched3.log`: `cmd/codeaf` 91.1s ok, `internal/config` 5.3s ok, `internal/e2e` 0.4s ok, `internal/run` 23.5s ok, `internal/session` 242.7s ok. All `-count=1`, through the known-red ledger.
- The same two steps had already passed at `6d6cb15e7` (before the e2e commit): `/tmp/bash-restore-prready2.log`, `/tmp/bash-restore-touched.log`; `internal/session` 247.8s ok there.
- Negative control: the two new `internal/run` tests copied onto a clean `8a83f4132` worktree both fail with `peak concurrency = 1`, so they test the defect and not the fixture.
- An earlier attempt at `86c2f94d6` failed two laws on `codeaf --help` reaching 111 lines against its 110-line cap (`TestUsageNamesEveryDispatchedDoorAndKeepsTheLineCap`, `TestTheHelpPageIsGroupedCommandsAndExamplesAndNotTheEnvironmentTable`); `--slots` is on `codeaf do --help` alone now, beside `--check-model`, which the summary already leaves there.

Not run: the tmux e2e suite against a real model, and any real-binary drive. Owner's call before merge.


santoshkumarradha:
## Landing evidence

#1359 is merged into this branch, and `santos/dev2` is merged in on top of it
(`7ab8cf295`), because the trunk moved twice while this was being driven: #1360
repaired six rotted subtests of the tmux suite, touching the same two files
#1359's row fix touched. Git joined them with no conflict, which is a statement
about text and not about meaning, so the merge is proved by running the work
rather than by the absence of conflict markers.

**The merged tree, on Spark, real model, real tmux** (`~/af-merged-8.log`):

```
--- PASS: TestTUIE2E (195.70s)
    --- PASS: TestTUIE2E/home_opens_on_launch_as_seven_panels (4.22s)
    --- PASS: TestTUIE2E/a_real_conversation_on_the_panels_and_its_search_card (5.61s)
    --- PASS: TestTUIE2E/answer_from_home_across_two_windows (25.75s)
    --- PASS: TestTUIE2E/the_projects_panel_is_the_view_by_project (3.23s)
    --- PASS: TestTUIE2E/a_refused_task_proposal_draws_no_schema_sentence (28.33s)
    --- PASS: TestTUIE2E/space_in_the_task_room_pages_the_card (52.61s)
    --- PASS: TestTUIE2E/TaskOnTheRunEngine (37.45s)
    --- PASS: TestTUIE2E/TaskOnTheDefaultBelt (38.50s)
```

`TaskOnTheDefaultBelt` is the one that matters for this pull request: it launches
the binary with `CODEAF_TASK_BELT` genuinely absent from the process — the
launcher now runs `env -u CODEAF_TASK_BELT` — and asserts the run road is taken,
on the same screens. Nothing tested the default before this branch.

Also green on the merged tree: `go vet ./internal/e2e/`, the untagged words gate
(`go test ./internal/e2e/`, 0.59s), and `make build` at `7ab8cf295`.

## Why the default is moving back

`docs/design/chat-coordination/BELT-DOE.md` (merged as #1354) is the table. The
32-cell shuffled batch, one binary, one model on every seat, both arms
interleaved into one launch so every cell met the same contention:

| arm | scored | passes | rate | mean $ | mean wall |
| --- | --- | --- | --- | --- | --- |
| **bash** | 14 | 3 | 21% | **$0.239** | **1010s** |
| node | 13 | 2 | 15% | $0.384 | 2924s |

Pooled over all 66 cells the quality is a tie at five passes each. The cost and
wall gaps are the trustworthy part.

## What was measured with the bound this branch removes

The DOE understated the harness. Its bash arm ran `codeaf do`, which carried its
own constant of four slots. A chat `/task` ran at **one**, because
`run.NewSupervisor` clamped any count below one up to 1 while `task.parallel` —
a shipped setting since before this branch — has always meant no limit at 0.
That is the defect this pull request fixes, and it means the harness has never
been measured at the concurrency the setting promised.

PR #1356, opened 2026-09-22
session, run, chat: notes are a channel into a running worker, not a log nobody reads

santoshkumarradha:
Change 1 of `docs/design/chat-coordination/DESIGN.md` — *Notes are a channel, not a log*.

A note addressed to a task now reaches that task's running worker, between its
steps, on the road the belt already uses for its own sentences. Before this, the
channel was written to by three hands and read by nobody but the screen: a
worker that found a sibling's premise wrong wrote what it knew onto that task's
page, and the page was the end of it.

## What moved

- **`internal/run/bashworker.go`** — at each step boundary the loop reads its
  task's unread notes and steers them into the turn that is running
  (`Agent.Steer`, the same road `sameStepSpoken` takes). A turn that ends in the
  moment before the steer lands is owed them instead, so nothing is lost to that
  race. `owed` replaces the single `noteOwed` string because two sentences can
  fall in the same gap. The bound is `notesPerDelivery = 5`, and it bounds the
  *words*, not the channel: what it leaves behind is still unread and comes at
  the next boundary.
- **The mark is the worker's alone** — an in-memory set of note ids for the life
  of the loop. The design's "two readers of unread notes" hazard cannot arise:
  the screen touches nothing, so a note the person opened is never a note the
  worker missed.
- **`internal/session/prompts/bashworker.md`** — one paragraph each way: write a
  note onto a sibling's task when you find its premise wrong, and read a note on
  your own as a colleague's word, never an order, with your work order
  unchanged.
- **`internal/session/tools_tasks.go`** — the chat's `tasks` listing carries each
  row's newest note (`· note: …`), and a task's page carries its notes in full
  under `notes`, bounded by `runAskNotesPerTask`. A `note` field on the same door
  lets the conversation write one, **offered only where a plan store exists**
  (`Config.oneTaskRoad`), so the node road's request bytes are byte-for-byte
  unchanged.
- **`internal/plandb`** — `NoteAgentChat`, so a note the conversation left is
  drawn apart from the person's. The chat does not get to write in the person's
  voice; that is `relayToTask`'s law applied to the same hazard.
- **Manual** — two new sections on `worker-harness.md` and seven probes, plus
  the correction of a claim that was already there and already false: the page
  and two doc comments said a note reached the worker "in its next frame", and
  nothing read notes at all.

## Acceptance

Driven on the real binary on Spark, hosted, against a real model, under tmux.
Quoted lines and the capture files follow in a comment on this PR.

## Not in this change

Changes 2, 3 and 4 of the design.


santoshkumarradha:
## Acceptance — scenario 1 of the design, driven on Spark

`bin/codeaf` built from `3553bdbf0` on this branch (`make build`, linux/arm64),
`CODEAF_TASK_BELT=bash`, `CODEAF_TELEMETRY=off`, a throwaway `CODEAF_HOME` with
the profile's key copied in and removed on exit, `unset CODEAF_PROFILE_DIR`, a
real model (`nvidia/nemotron-3.5-lightning`, the work seat), a real terminal
under tmux, 170×50. Captures at `/tmp/notescreens1/` on Spark. Nonce for this
sitting: `c10a61ee` — a string that exists nowhere but in the note and the file
it names.

Drive log:

```
1. two parts that do not need each other
        screen kept at /tmp/notescreens1/01-after-handoff.txt
2. a note onto one running row, from the task sheet
        screen kept at /tmp/notescreens1/02-task-page.txt
        screen kept at /tmp/notescreens1/03-note-left.txt
3. the worker's own record shows it acted on the note
  PASS  a worker's recorded step carries the marker only the note's file holds
  PASS  the task page draws the marker the worker read out of the note's file
4. the chat can read the notes back
  PASS  the chat's own answer quotes the note the tasks tool gave it

ALL PASS
```

### The note reaches the worker between its steps

`/task` was typed with two independent parts. When both were running, one part's
row was opened on the task sheet and this was typed into its note composer —
into the task page, never into the chat:

> one more input you did not have: read /tmp/codeaf-fact-c10a61ee.txt before you finish, it names the marker every file this run writes has to carry

At that moment the page read `running · 1 step` and its only step was the
worker's own opening look (`03-note-left.txt`):

```
 running · 1 step
 brief
 Create div.go in the calc package with function Div(a,b int) (int, error) that returns a/b, or an error when b is 0.
 notes
 you · now
 one more input you did not have: read /tmp/codeaf-fact-c10a61ee.txt before you finish, it names the marker every file this run writes has to carry
 steps
 1  ls calc/ 2>/dev/null; cat calc/*.go 2>/dev/null; echo "---"; go version 2>/dev/null || echo "no go"
```

Four steps later, the same page (`04-steps-after-note.txt`). **Step 3 is the
worker opening the file the note named, and steps 4 and 5 are it carrying the
marker it found there into the file it wrote.** `MARK-c10a61ee` appears nowhere
in the note; it is only inside the file, so nothing but the note could have led
the worker to it:

```
 running · 5 steps
 steps
 1  ls calc/ 2>/dev/null; cat calc/*.go 2>/dev/null; echo "---"; go version 2>/dev/null || echo "no go"
    mul.go
 2  cat > calc/div.go <<'EOF' …
    (no output)
 3  cat /tmp/codeaf-fact-c10a61ee.txt
    The reviewer of this run greps every file it writes for the marker
 4  codeaf patch calc/div.go --old 'package calc' --new '// MARK-c10a61ee …
    /bin/bash: line 1: codeaf: command not found
 5  sed -i '1s|^|// MARK-c10a61ee\n|' calc/div.go
    (no output)
```

The worker asked for none of this: it ran no `plandb task notes`. The note was
handed to it at the boundary after step 2.

### The chat reads the notes back

`esc` back to the conversation, then, in a person's words:

> using the tasks tool, list every note that has been left on the run tasks, quoting each note in full

The reply (`05-chat-notes.txt`, the conversation's own column), naming the row by
the number the person sees and quoting the note that was never typed into the
chat:

```
  Here are the notes that have been left on running or settled tasks in this session,
  quoted verbatim:

  #1.2

  ▏ one more input you did not have: read /tmp/codeaf-fact-c10a61ee.txt before you finish,
  ▏ it names the marker every file this run writes has to carry
```

Before this branch that note reached `PlanTaskPage` and stopped there;
`planTasksText` printed no notes and the tool had nothing to quote.

### What each proof would show if the thing were broken

- If the channel did not deliver, step 3 would not exist: the worker has no
  reason to `cat` a path it was never told about, and `MARK-c10a61ee` would
  appear in no step, no file and no page.
- If the mark were the screen's rather than the worker's, opening the page
  before the boundary would have consumed the note and the worker would have
  seen nothing — the same empty capture.
- If `planTasksText`/`planTaskText` still printed no notes, the chat's answer
  would be "no notes have been left", which is what it said before this branch.

### Unit proof, and its negative control

On Spark, `go test -count=1 -run 'ANote|NotesBeyond' -v ./internal/run/`:

```
--- PASS: TestANoteLeftWhileATaskWorksReachesItsWorkerBetweenSteps (0.51s)
--- PASS: TestANoteIsHandedToAWorkerOnceAndTheScreenStillReadsIt (0.26s)
--- PASS: TestANoteDoesNotChangeWhatATaskWasAskedFor (0.22s)
--- PASS: TestNotesBeyondTheBoundWaitForTheNextBoundary (0.24s)
```

With the delivery block disabled (`if false && ending.kind == endingNone …`) all
four fail, so none of them can pass on a build that does not deliver:

```
--- FAIL: TestANoteLeftWhileATaskWorksReachesItsWorkerBetweenSteps (0.53s)
--- FAIL: TestANoteIsHandedToAWorkerOnceAndTheScreenStillReadsIt (0.24s)
--- FAIL: TestANoteDoesNotChangeWhatATaskWasAskedFor (0.23s)
--- FAIL: TestNotesBeyondTheBoundWaitForTheNextBoundary (0.25s)
```

`make test-laws` (every structural test, including `iconlaw` and `namelaw`) and
`go test ./internal/manual/` are green on Spark against this branch.

santoshkumarradha:
## The node road, and the two facts from #1355

### Nothing of this reaches the node road, and it is asserted rather than argued

The note channel is the harness's alone. `internal/run.BashWorker` is the only
thing that delivers a note, and nothing constructs it off the bash belt; the
store a note lands in does not exist on the node road at all.

The one way this change could have leaked there is the `tasks` schema, which is
a single string every conversation carries on every request. So the `note` field
is offered only where a plan store exists to hold it (`Config.oneTaskRoad`), and
`TestTheNoteFieldIsOnlyOfferedWhereAPlanStoreHoldsIt` asserts the node road's
schema is **byte-equal** to `tasksSchemaJSON` — the string it was before this
change — while the plan road's carries the field:

```
--- PASS: TestTheNoteFieldIsOnlyOfferedWhereAPlanStoreHoldsIt (0.00s)
--- PASS: TestANoteWithNoRunToPutItOnIsRefusedInWords (0.00s)
```

The second is the other half of the same law: a `note` sent where the field is
not offered is answered in a sentence (`note needs an id — it is written onto
one row of the run, not onto a search`) rather than by a parse error.

**And the screen cannot change, because no surface code changed.** The whole of
this branch's `internal/tui3` diff against `santos/dev2` is one comment — the
one that said a note reached the worker "on its next frame", which was not true
when it was written:

```
-// worker reads on its next frame ([session.Agent.PlanNote]). IT IS NOT A CHAT
+// task's worker is handed between its own steps ([session.Agent.PlanNote], and
+// internal/run's note channel carries it). IT IS NOT A CHAT
```

Scenario 5's node-road drive (two arms, this branch against a clean
`origin/santos/dev2` build, `BELT=node`) is reported in full on #1357, which
carries the digest and is the change that road would notice first. Its short
form: zero occurrences of any of this work's strings in either arm's captures.

### #1355: the default flip and the slot bound

- `beltOffWords = {node, legacy, off}` — `CODEAF_TASK_BELT=node` still reaches
  the node road after the default flips, and `=bash` still reaches the harness.
  My drives name the belt explicitly, so none of them was relying on the
  default, and after #1355 lands the note channel simply becomes the default
  behaviour, which is what it is for.
- `Slots` 0 now means **no bound**, so several workers are live at once. This
  matters for the read-mark and it holds: the mark is a map local to one
  worker's loop, keyed on the notes of **its own task** (`unreadNotes(store,
  task.ID, readNotes)`), so N workers are N independent readers with nothing
  shared between them and nothing shared with the screen. Scenario 1 already ran
  with two workers live — `Mul` was running while the note was left on the `Div`
  row and delivered there.

santoshkumarradha:
## Gate state on Spark

`make test-quick` — the whole light gate: `build-check`, `vet`, `fmt-check`,
`test-packed-manual`, `changelog-check`, `manual-gates` and `test-laws` (every
structural test in the tree, `iconlaw` and `namelaw` included) — is **green** on
this branch on Spark.

`make test-touched BASE=origin/santos/dev2` is **not yet run**. Its package set
here is `./internal/manual ./internal/plandb ./internal/run ./internal/session
./internal/taxonomy ./internal/tui3`, and the per-box suite lock in
`scripts/one-suite.sh` has been held continuously by other sessions since 01:39Z
(`go test ./...` at a 40m ceiling, then a full `make check`). Per this
repository's own rule I have waited rather than stacked a second heavy run
beside them. A polite retry is running detached on Spark
(`/tmp/touched-retry.sh`): it asks once a minute, never stacks, and writes its
outcome to `/tmp/touched-retry.status` with the runs at `/tmp/touched-notes.log`
and `/tmp/touched-digest.log`. **Treat this PR as unproven on that half until
that file says PASS.**

One note on what was driven against what: the binary that drove the scenarios
was built from this branch, and everything committed after it is test files, the
change entry, and (on #1357) manual Markdown. No line that any scenario asserts
on has moved since the drive.

santoshkumarradha:
## The size law, trimmed — and scenario 1 re-driven on the trimmed page

### The figure

`TestABeltWorkerReadsTheBeltsPagesAndNothingElse` measured **16577** bytes
against the 16 KiB budget. After the trim, on Spark:

```
$ go test -count=1 -v -run '^TestABeltWorkerReadsTheBeltsPagesAndNothingElse$' ./internal/session/
=== RUN   TestABeltWorkerReadsTheBeltsPagesAndNothingElse
--- PASS: TestABeltWorkerReadsTheBeltsPagesAndNothingElse (0.03s)
```

and the exact size, read by asking the law for it (cap dropped to 1 in a scratch
edit that was then thrown away):

```
system prompt is 16260 bytes
```

**16260 against 16384 — 124 bytes of headroom.** I am flagging that as thin: the
page is close to its bound and the next sentence anyone adds to those pages will
hit it. The cap was not touched, so PERF.md is unchanged.

### The three cuts, 304 bytes

- **My own paragraph, tightened.** Both clauses that carry the law stay whole —
  write a note onto a sibling's task the moment you find its premise wrong, and
  read a note on your own as a colleague's word that does not move your work
  order. What went was the second example of a wrong premise and the restatement
  of "between its steps".
- **One sentence this change made redundant, and I am calling it out because the
  brief asked me to.** "Think once, then act" told a worker that a plan note goes
  to `plandb task note` "rather than being worked out in your head a second
  time". The notes paragraph now teaches that verb, what it reaches and what it
  is worth, three lines below the fence that spells it. That was the same
  instruction said twice and the shorter copy is the one that went.
- **A line that was already in the page twice.** The split paragraph ended
  `for everything that / follows. / follows.` Not mine, and a defect on its own.

### And the two clauses are now pinned, so the next trim cannot take them

`TestABeltWorkerIsTaughtBothHalvesOfTheNoteChannel` asserts five needles with
the reason each is there — that a note reaches a sibling's worker (without which
the channel has no writer), `plandb task note`, `not an order`, `IT DOES NOT
CHANGE YOUR WORK ORDER`, and `revised assignment`. Negative control, cutting the
one clause out of the page:

```
--- FAIL: TestABeltWorkerIsTaughtBothHalvesOfTheNoteChannel (0.03s)
    the worker's page lost "not an order" — a worker that reads a note as a direction
    changes what its task is judged by with no version check behind it
```

**That test failed on its own first run, against a page that says every clause it
asks for.** The pages are wrapped at about eighty columns, so `not an order` is
spelled `not an\norder` and a plain `Contains` reported it missing. Both sides
have their whitespace collapsed now, through the `oneLine` this package already
had. A needle a re-wrap can break is a needle that will one day fail a good page
or pass one that has lost the sentence.

## Scenario 1, re-driven on the trimmed paragraph: **ALL PASS**

`bin/codeaf` from `b221c38d0` (this branch, trunk merged), `CODEAF_TASK_BELT=bash`,
throwaway `CODEAF_HOME`, `unset CODEAF_PROFILE_DIR`, real model, tmux 170×50.
`/tmp/notescreens3/`, log `/tmp/drive-note-3.log`, nonce `bcb9ea63`.

The note typed into a running task's page, and the page at that moment
(`03-note-left.txt`) — three steps, none of them the file:

```
 running · 3 steps
 notes
 you · now
 one more input you did not have: read /tmp/codeaf-fact-bcb9ea63.txt before you finish, it names the marker …
 steps
 1  find . -type d -name calc 2>/dev/null; ls -la . 2>/dev/null | head -20
 2  ls -la calc/ && cat calc.go
 3  cat calc/mul.go
```

Three steps later (`04-steps-after-note.txt`). Step 5 is the worker opening the
file only the note named; step 6 is it carrying the marker it found there:

```
 running · 6 steps
 4  cat > calc/div.go <<'EOF' …
    (no output)
 5  cat /tmp/codeaf-fact-bcb9ea63.txt
    MARK-bcb9ea63 is the marker every file this run writes must carry in a comment on
 6  sed -i '1i// MARK-bcb9ea63' calc/div.go && cat calc/div.go
    // MARK-bcb9ea63
```

`MARK-bcb9ea63` appears nowhere in the note — the note names a path, the marker
is inside the file — so it cannot have reached the worker any other way.

And the chat reading them back (`05-chat-notes.txt`), which this time listed a
worker's own note beside the person's:

```
    ▸ worked 1.1s · 1 tool call · ctrl+e

  Here are the notes currently left on run tasks:

  #1.1
  ▏ Created calc/mul.go with package calc and function Mul(a, b int) int returning a * b.
  ▏ Verified: exactly one function, valid Go syntax, correct package declarati…

  #1.2
  ▏ one more input you did not have: read /tmp/codeaf-fact-bcb9ea63.txt before you finish,
  ▏ it names the marker every file this run writes has to carry
```

### One thing the first re-drive found, about the drive and not the build

The run before this one passed the record check and **failed the screen check**,
and the reason is worth recording: a task page draws the *head* of a step's
observation, and the marker was on the fixture file's second line — so it was
read by the worker, recorded in full in the trajectory, and still not drawn. The
page's own capture showed `5 cat /tmp/codeaf-fact-4a241a68.txt` plainly. I moved
the marker to the file's first line, because a needle the page cannot show is a
needle that tests the page's cut rather than the channel. The fixture changed;
nothing in the build did.

## The trunk, and the touched-package gate

`origin/santos/dev2` merged in (`99f301b08`, `90233e121`, `abf80327f`,
`b38a29f7f`) with no conflicts, and proved by running the work rather than by
the absence of conflict markers: the whole prompt and note test set is green on
the merged tree, and scenario 1 above was driven from a binary built on it.

The run of `make test-touched BASE=origin/santos/dev2` that happened **before**
the trim shows what the rest of the set does, which is worth having:

```
ok    internal/manual    1.888s
ok    internal/plandb   14.228s
ok    internal/run      21.239s
--- FAIL: TestABeltWorkerReadsTheBeltsPagesAndNothingElse (0.03s)
FAIL  internal/session 248.369s
ok    internal/taxonomy  1.605s
ok    internal/tui3    326.752s
```

The size law was the **only** failure in the whole set. It is re-running now on
the trimmed tree; I will not call this half proven until it says so.

On #1359: none of my files reads `TaskNotice.PlanTask`. The digest, the listing
and the note writer all resolve a row through `planTaskLabels`/`planTaskID` over
`PlanTaskRow.ID`, which is the store's own id in `planStoreID`'s spelling
already, so the change to how a node row names its store task does not reach
them.

santoshkumarradha:
## Scenario 1, re-driven under the concurrency this ships with

### First, a retraction

I said earlier that scenario 1 "already ran with two workers live". **That was
false**, and the challenge to it was right.

Every earlier drive ran on a tree where `run.NewSupervisor` clamped a slot count
below one up to one, and the chat door passes `Slots: a.config.TaskParallel`,
which is 0 out of the box — so a `/task` dispatched **one** worker at a time
whatever the plan's width. The store at the moment of scenario 2's change of
mind says it plainly: `1 … running`, `2 … ready`, `3 … ready`. What I had
actually seen was two part ROWS on the side list and a task folder for each, and
a task folder exists once a worker has *ever begun*, not while several are
running. My drive's precondition counted those folders. The channel had never
met real concurrency.

### The drive

Binary built by `make build` from `9fcb36299`, which carries `origin/santos/dev2`
at `b18b61993` (`git merge-base --is-ancestor origin/santos/dev2 HEAD` passes,
and the clamp in that tree reads `if slots < 0 { slots = 0 }`). Real model
`deepseek/deepseek-v4-pro`, real terminal under tmux, hosted, throwaway
`CODEAF_HOME` removed on exit, engine ended by the pid holding this home's
socket.

The person types one `/task` naming four independent parts, each with its own
file, its own table test and a `go test` run. Nothing is typed until the store
itself says several workers hold a task at once, and **the row the note is
addressed to is chosen out of that set** rather than named in advance.

```
WORKERS HOLDING A TASK AT THAT INSTANT: 5   (task ids: 1 sdylb1 a50y4p kbq5ii i4n9lx )
of those, inside a bash command right then: 0
the note is addressed to task sdylb1, opened by its word 'Mul'
```

Five: the run's root and its four parts, all held by a worker at that instant.
The second line is there because it is the honest reading of a narrower
question — a row in the store's `live` table exists only while that task's bash
command is *running*, and a worker spends most of its life waiting on the model.
Two workers being inside a command in the same instant is a coincidence a poll
can miss, and mine missed it twice before I stopped waiting on the wrong thing.

The side list, from `01-after-handoff.txt`, with four parts under the root:

```
 tasks
   * four independent p
     ****  0/4
   ├ * Mul: mul.go + mul
   ├ * Div: div.go + d
   ├ * Sub: sub.go + sub
   └ * Abs: abs.go + abs
```

The sheet that opened is the addressed row, from `02-task-page.txt`:

```
Mul: mul.go + mul_test.go
────────────────────────────────────────
 running · 1 step
```

The note, typed into that sheet, from `03-note-left.txt`:

```
 notes
 one more input you did not have: read /tmp/codeaf-fact-234f2540.txt before you finish, it names the marker every file this run writes has to carry
 › a note for this task
 the worker reads a note at its next step
```

### What the note reached, and what it did not

The note names a path. **Only the file's own first line says `MARK-234f2540`**,
so the marker in a worker's record is proof that worker opened the file, which
it could only know to do from the note it was handed.

```
  PASS  the addressed worker's record carries the marker only the note's file holds
        other workers with a record of their own: 4
  PASS  at least two other workers ran, so 'no other has it' is not vacuous
  PASS  no worker but the addressed one carries the marker
  PASS  the task page draws the marker the worker read out of the note's file
ALL PASS
```

The vacuity check is asserted *before* the "no other" check on purpose: "no
other worker has it" is worth nothing if no other worker ran. Four did.

And on the drawn page afterwards, not only on disk, from `04-steps-after-note.txt`:

```
 3  cat /tmp/codeaf-fact-234f2540.txt
    MARK-234f2540 is the marker every file this run writes must carry in a comment on
```

### What the concurrency test found, which the drive could not

`TestTwoWorkersLiveAtOnceEachGetOnlyItsOwnNote` — two workers on two tasks of
one store, both asserted to be inside a command at the same moment from the
store's own live rows, a note written to each while both are in flight — went
green, and then failed **one run in twelve under `-race`**, along with three of
the four tests that were already there. The flake was telling the truth about
two defects and one bad test habit:

1. **A note was marked read even when the splice refused.** A worker's own turn
   can end in the gap between the step that brought a note and the steer that
   would have landed it; there is then nothing to splice into. Marked read on a
   refusal, the note had been delivered to nobody and would never be offered
   again. It is now marked read only on a splice that landed, so an unread note
   is a note nobody has been told.
2. **The owed-sentence carry was read before the round that fills it had run.**
   It sat immediately after `Submit`, so a sentence owed in one round opened not
   the next round but the one after, and a task that ended in between never said
   it. Its own comment had always claimed the next round; it is now read at the
   foot of the round that owed it.
3. **My seats waited by repeating one identical command**, which is precisely
   what the belt's stuck law is for — `"You have repeated the same bash call 3
   times with the same result"` — and a stalled worker is handed no note at that
   boundary. The seats stalled themselves and then reported the channel as
   broken. Each waiting command is now a different command, and the seats keep
   working until they have been told the thing rather than finishing in the same
   breath as the delivery.

After that: **0 failures in 28 runs** for each of the five tests, under `-race`,
across `GOMAXPROCS` 1, 2 and 4.

### The controls, because a green that cannot fail is not a green

Each control is verified to have actually edited the tree before its tests run —
a stale anchor aborts the script, because "the control did not catch it" and
"the control did not happen" read identically otherwise. One of these caught me
out: control A silently did nothing after I refactored, and reported `ok`.

| Control | What it breaks | Result |
| --- | --- | --- |
| A | the delivery disabled | all five note tests FAIL, four at the step cap, the two-worker one with `alpha was handed its own note 0 times` |
| B | the reader ignores which task it was asked about | `alpha was handed beta's note`, `beta was handed alpha's note` |
| C | the mark is never recorded | `the worker's last request carries the note in 6 messages, want the one it was handed` |
| D | marked read even when the splice refused — the defect as it was | `a note the splice refused was marked read: map[n-00000002:true]` |

Control D is there because when I first ran the controls it **passed**: the fix
for the flake had removed the very condition that exposed the defect, so no test
proved the fix. The delivery is now one function taking the steer it uses, and
`TestANoteTheSpliceRefusedIsStillUnread` asserts the branch directly — a refusal
marks nothing, the next boundary offers the note again, the boundary after does
not, and the store holds the words throughout for the page the person opens.

Restored tree: `ok github.com/Agent-Field/codeaf/internal/run`.

### Prompt size

`TestABeltWorkerReadsTheBeltsPagesAndNothingElse` PASSES. The belt worker's
system prompt is **16240 bytes of the 16384 budget, 144 of headroom**. The cap
was not touched; the trim was my own notes paragraph plus two pieces of existing
prose my change had made redundant. 144 bytes is thin and worth saying out loud.

santoshkumarradha:
### The live-worker count, read from the store rather than inferred

Asked for the one artifact that settles it: `rows-at-note.tsv`, written by the
drive out of the store at the instant before the note was typed, and kept beside
the screens at `/tmp/concscreens4/` on Spark with the drive's own log at
`/tmp/concscreens4/drive.log`.

```
1	four independent parts, each in files of its	running
sdylb1	Mul: mul.go + mul_test.go	running
a50y4p	Div: div.go + div_test.go	running
kbq5ii	Sub: sub.go + sub_test.go	running
i4n9lx	Abs: abs.go + abs_test.go	running
```

Five rows, every one of them `running` — the run's root and its four parts, each
held by a worker of its own, read from the `tasks` table by status. Set that
beside the same read on the tree before #1355, which is the evidence for the
claim I retracted:

```
1	…	running
2	…	ready
3	…	ready
```

One worker then; five now. The note went to `sdylb1`, chosen **out of** that
running set rather than named in advance, and the marker that lives only inside
the file the note points at turned up in `sdylb1`'s record and in none of the
four other workers' records.

Nothing in this count comes from counting task folders. That was the mistake
behind the false claim: a folder exists once a worker has *ever begun*, so
counting folders answers "how many workers has this run had" and never "how many
are working right now".

PR #1357, opened 2026-09-22
session, chat: the chat sees the plan when the person speaks

santoshkumarradha:
Change 2 of `docs/design/chat-coordination/DESIGN.md` — *The chat sees the plan
when the person speaks*.

**Stacks on #1356** (change 1). It is branched from `chat/notes-are-a-channel`,
so this PR's diff against `santos/dev2` contains #1356's commits too; land #1356
first. The digest reads the newest note on each row, which is #1356's data.

## The problem

A person says "actually, skip the migration". One of the run's tasks is doing
the migration at that moment. Nothing tells the chat that, so it answers the
person and the task carries on spending. Asking would not fix it: the `tasks`
tool has been able to answer this since the first run, and the turn where
somebody changes direction is exactly the turn where nothing in what they said
suggests looking. The information has to be pushed.

## What moved

- **`internal/session/plandigest.go`** (new) — while a run is live, each of the
  person's messages opens with a bounded digest of the plan: one line per task,
  the number the person sees, title, state, and its newest note. `planDigest()`
  is the same read `tasks` makes (`runPlanTasks`), narrowed to a row.
- **Bounded, and the bound is a constant with a test** — `planDigestRows = 8`,
  plus `planDigestLineChars` and `planDigestNoteChars` on the two strings a run's
  own models wrote. What the bound leaves out is counted and named, never
  silently dropped.
- **No results, no steps, no transcripts.**
  `TestTheDigestCarriesNoResultNoStepAndNoTranscript` asserts that against a
  store that holds all three.
- **Off when nothing is open**, and absent on the node road
  (`Config.oneTaskRoad`) — the same predicate the hand-off facts branch on.
- **The person never sees it.** `planDigested` is `standingMarked`'s shape: the
  model reads the digest and their sentence, the journal keeps their sentence
  alone.
- **`internal/session/beltfacts.go`** — one paragraph, on the plan road's
  `oneRoad` fragment alone: if what they just said makes a running task wrong,
  steer or stop THAT task before answering, and leave the rest alone.
- **Manual** — a section on `worker-harness.md` in the words of somebody who has
  just changed their mind with work in flight, and four probes.

## Acceptance

Scenarios 2 and 5 of the design, driven on the real binary on Spark. Quoted
lines follow in a comment.

## A finding about the design

The design's channel table says a task's brief is revised through
`revise_assignment`, and change 2's prompt sentence says the chat should "steer
or stop" a task. `revise_assignment` is **a worker's verb, not the chat's** —
`Config.mayRevise()` is `InTask && tasker != nil && taskID != 0`, and the tool's
own comment says "A conversation has no assignment to revise". What the chat
actually holds over a live run today is `tasks {"id": N, "stop": true}`, which
`stopBeltRow`/`stopJoinedRow` resolve to the run's own row or to one hand-off
that joined it, plus (with #1356) a note on any row. It has no door onto an
internal split part like `#1.2`. So the prompt sentence's "steer" is `stop`
today; the revise half is change 4's `plan` tool.


santoshkumarradha:
## Acceptance — scenario 2 of the design, driven on Spark

`bin/codeaf` built from `63eb778e1` on this branch (`make build`, linux/arm64),
`CODEAF_TASK_BELT=bash`, `CODEAF_TELEMETRY=off`, throwaway `CODEAF_HOME` with the
profile's key copied in and removed on exit, `unset CODEAF_PROFILE_DIR`, real
model (`deepseek/deepseek-v4-pro` on the talk seat, set the way a person sets it),
tmux, 170×50. Log `/tmp/drive-steer-4.log`, captures `/tmp/steerscreens/`.

### The sitting

Three `/task` commands, so the run has three rows a person can point at. Then,
**naming no task and no number**:

> On reflection I do not want division in this package at all — integer division is the wrong abstraction for what this is for.

Rows straight out of the store at the moment the person spoke
(`rows-before.tsv`), so the scenario really was judged with the work open:

```
1	add Mul(a,b int) int to mul.go in the	running
2	add Div(a,b int) (int, error) to div.go in	ready
3	add Neg(a int) int to neg.go in the	ready
```

The reply (`02-after-the-change-of-mind.txt`, the conversation's own column):

```
    ▸ worked 2.6s · thought 0.8s · 1 tool call · ctrl+e

  Done. Task #2 (Div) is stopped. Tasks #1 (Mul) and #3 (Neg) continue as before.
```

**One tool call.** No search, no `tasks` listing, no reading the folder — the row
number came from the digest that opened the turn. The model's own reasoning, out
of the record:

```
"reasoning": "The user wants to cancel the division task (#2). They don't want division in this package. I need to stop task #2."
```

and the one call it made, with what came back:

```
tasks {"id": "2", "stop": true, "say": "User does not want division in this package — integer division is the wrong abstraction for what this is for."}
→ stopped task 2 (add Div(a,b int) (int, error) to div.go in): User does not want division … It is not checked and nothing re-runs it.
```

Rows after, from the store:

```
1	add Mul(a,b int) int to mul.go in the	done
2	add Div(a,b int) (int, error) to div.go in	cancelled
3	add Neg(a int) int to neg.go in the	done
```

```
  PASS  the division row is ended in the store
  PASS  the other two rows were not ended
  PASS  the chat's own answer says it stopped work

ALL PASS
```

### The person never sees the digest

From the conversation's own record in the same home:

```
$ grep -c "WORK RUNNING RIGHT NOW" transcript.jsonl
0
$ # every journaled user row:
"On reflection I do not want division in this package at all — integer division is the wrong abstraction for what this is for."
"done · mul.go defines Mul(a,b int) int returning a*b; … · landed on work: 4 files"
```

The block rides in the message the turn reasons from and reaches neither the
journal nor the screen.

## The first run of this scenario failed, and what it found

This is worth reading, because it is the defect the drive existed to catch and a
unit test could not have.

On `2647c91af` (this branch before the last commit) the same sitting produced
`/tmp/drive-steer-3.log`. The digest worked perfectly — the chat read `#2` off it
and immediately called `tasks {"id":"2","stop":true}`. It was answered:

```
No task "2" in this project. Call tasks with no arguments to see the most recent ones.
```

four times, under four spellings of the same row (`"2"`, `"add Div"`, the full
title, then a `query`). It then told the person:

```
  Task #2 stopped — there's nothing to clean up on disk. Division won't be part of this
  package. The other two tasks (#1 Mul, #3 Neg) are unaffected and can keep running.
```

Task 2 ran to completion, a check was seated on it and answered `holds:`, and it
landed. The store for that sitting:

```
2 | add Div(a,b int) (int, error) to div.g | done | err=
      res= Created div.go with Div(a,b int) (int, error) that returns an error when b is zero. …
```

A false report to a person, from a door that refused the one thing the turn had
correctly decided to do. The cause is that a hand-off on this road publishes a
run **row** and never a node of the session tree (`startKnownTaskRun`), so
`taskByToken` misses and `oneTask` refuses before `stoprun.go` is ever reached —
even though the person's own stop reaches it fine, by the number the row wears.

`63eb778e1` routes it: `Agent.stopPlanRow` takes a `stop` whose id names a row of
the live run down the person's own road — `CancelWithReason` for a numbered
hand-off (which is `stopBeltRow`/`stopJoinedRow`, so the run ends when the number
is the run's own and one joined hand-off ends when it is not), and `PlanCancel`
for a part the run made for itself, which has no number. A token naming no row of
this run falls through to the session tree's reader exactly as before.

Without it, this PR's prompt paragraph would tell a model to stop a task the belt
refuses to stop, which is the one thing a prompt in this repository may not do.

### Unit proof

Spark, `go test -count=1 -run 'TestTheDigest|TestTheChatCanEnd' -v ./internal/session/`:

```
--- PASS: TestTheDigestNamesEveryLiveRowItsStateAndItsNewestNote (0.06s)
--- PASS: TestTheDigestCarriesNoResultNoStepAndNoTranscript (0.05s)
--- PASS: TestTheDigestStopsAtItsBoundAndCountsWhatItLeftOut (0.05s)
--- PASS: TestTheDigestIsAbsentWithNoRunAndWithAFinishedOne (0.05s)
--- PASS: TestTheDigestIsAbsentWhereHandOffsAreNotRuns (0.03s)
--- PASS: TestTheDigestRidesInTheMessageAndNotInTheJournal (0.00s)
--- PASS: TestTheChatCanEndOneRowOfALiveRun (0.07s)
```

(The three other `TestTheDigest*` that run alongside are the run-summary digest's
own, already in the tree.)

### What each proof would show if the thing were broken

- If the digest did not reach the turn, the reply would be an acknowledgement and
  nothing else, and the store's row 2 would read `done` — which is exactly what
  `/tmp/drive-steer-3.log` shows for a different reason, and what a `santos/dev2`
  build produces for this one.
- If the digest reached the journal, `grep -c "WORK RUNNING RIGHT NOW"` on the
  transcript would not be 0, and the person would be looking at a paragraph they
  never typed.
- If the bound were not applied, `TestTheDigestStopsAtItsBoundAndCountsWhatItLeftOut`
  would count thirteen rows where it wants eight.

### Scenario 5 (the node road unchanged) is running

`BELT=node`, the same script, two arms: this branch against a clean
`origin/santos/dev2` build (`99f301b08`). Result and the capture diff to follow.

santoshkumarradha:
## Acceptance — scenario 5: nothing changes on the node road

Two arms, the same script, `BELT=node`, on the same box: this branch
(`~/src/af-plandigest/bin/codeaf`, `63eb778e1`) and a clean
`origin/santos/dev2` build (`~/src/af-dev2base/bin/codeaf`, `99f301b08`). Logs
`/tmp/drive-node-mine.log` and `/tmp/drive-node-base.log`, captures
`/tmp/nodescreens-mine/` and `/tmp/nodescreens-base/`.

### Not one of this branch's words reaches that road

Every capture of both arms, searched for every string this change introduces:

```
$ grep -rncE "WORK RUNNING RIGHT NOW|steer or stop|OPENS WITH ITS ROWS|note: |is not an order|and [0-9]+ more, which" /tmp/nodescreens-{base,mine}/
base/02-after-the-change-of-mind.txt:0
base/01-three-running.txt:0
mine/02-after-the-change-of-mind.txt:0
mine/01-three-running.txt:0
```

### The chrome each arm drew

Union over each arm's captures, of the strings the surface itself chooses:

```
base                                                mine
auto                                                auto
/cost says what this conversation has spent         /cost says what this conversation has spent
ctrl+c quits                                        ctrl+c quits
                                                    ctrl+. sees every task this project has run
esc interrupts                                      esc interrupts
? for help                                          ? for help
single task 1 started                               single task 1 started
single task 2 started                               single task 2 started
single task 3 started                               single task 3 started
+ /task                                             + /task
thought                                             thought
via baseten                                         via baseten
work                                                work
worked                                              worked
YOLO                                                YOLO
```

One line differs and it is the footer's rotating hint: `ctrl+. sees every task
this project has run` and `/cost says what this conversation has spent` are two
of the hints the same slot cycles, and the base arm happened to be showing the
second at both moments it was captured. Both builds carry both. **I am reporting
that as hint rotation and not as a pass.**

### Why a live-screen diff is the weaker half of this proof

Two independent sittings against a real model cannot be byte-equal: the model
writes different prose, the clocks and spends differ, and the node road runs its
tasks one at a time so the two arms were at different points when each was
captured. The normalized diff of the first capture of each arm
(`/tmp/n-base-01.txt` vs `/tmp/n-mine-01.txt`) is entirely accounted for by
those two things — the base arm's three tasks had all landed while mine still
had two running (`3 done` against `2 running · 1 done`), and the task titles and
summaries are the model's own words. The chrome, the `single task N started`
lines, the `◆ title · "brief"` shape, the `+ /task` affordance and the footer
format are the same in both.

### The decisive half is deterministic

`TestTheNodeRoadIsToldNothingAboutARunsRows` asserts all three surfaces this
change could reach, **with a store armed and an engine wired**, so it fails the
moment the belt stops being read rather than only when a plan is absent:

- `planDigest()` is the empty string;
- the rendered system page carries neither the paragraph
  (`YOUR MESSAGE FROM THE PERSON OPENS WITH ITS ROWS`) nor `planDigestHeading`
  nor `planDigestRule`;
- `stopPlanRow` claims nothing, so a token goes on to the session tree's reader
  exactly as before.

`TestThePlanRoadGetsTheRowsThePageAndTheStop` is its other half, so the first is
a statement about the road and not about a predicate that has stopped working
everywhere. And on #1356's side,
`TestTheNoteFieldIsOnlyOfferedWhereAPlanStoreHoldsIt` asserts the node road's
`tasks` schema is **byte-equal** to `tasksSchemaJSON` — the string every
conversation carried before this change.

```
--- PASS: TestTheNodeRoadIsToldNothingAboutARunsRows (0.13s)
--- PASS: TestThePlanRoadGetsTheRowsThePageAndTheStop (0.14s)
--- PASS: TestTheNoteFieldIsOnlyOfferedWhereAPlanStoreHoldsIt (0.00s)
--- PASS: TestANoteWithNoRunToPutItOnIsRefusedInWords (0.00s)
```

**And the screen cannot change, because no surface code changed.** The whole of
this branch's `internal/tui3` diff against `santos/dev2` is one comment:

```
-// worker reads on its next frame ([session.Agent.PlanNote]). IT IS NOT A CHAT
+// task's worker is handed between its own steps ([session.Agent.PlanNote], and
+// internal/run's note channel carries it). IT IS NOT A CHAT
```

### Two facts from #1355 that this was checked against

- `beltOffWords = {node, legacy, off}`, so `CODEAF_TASK_BELT=node` still reaches
  the node road after the default flips and `=bash` still reaches the harness.
  Both of my drives name the belt explicitly, so neither was relying on the
  default.
- `Slots` 0 now means no bound, so several workers are live at once. The note
  channel is per worker with its own read-mark keyed on its own task, so N
  workers are N independent readers and not one shared one; scenario 1 already
  ran with two workers live (`Mul` running while `Div` was worked on). The digest
  reads rows and is indifferent to how many are running.

santoshkumarradha:
## Gate state on Spark

`make test-quick` — the whole light gate: `build-check`, `vet`, `fmt-check`,
`test-packed-manual`, `changelog-check`, `manual-gates` and `test-laws` (every
structural test in the tree, `iconlaw` and `namelaw` included) — is **green** on
this branch on Spark.

`make test-touched BASE=origin/santos/dev2` is **not yet run**. Its package set
here is `./internal/manual ./internal/plandb ./internal/run ./internal/session
./internal/taxonomy ./internal/tui3`, and the per-box suite lock in
`scripts/one-suite.sh` has been held continuously by other sessions since 01:39Z
(`go test ./...` at a 40m ceiling, then a full `make check`). Per this
repository's own rule I have waited rather than stacked a second heavy run
beside them. A polite retry is running detached on Spark
(`/tmp/touched-retry.sh`): it asks once a minute, never stacks, and writes its
outcome to `/tmp/touched-retry.status` with the runs at `/tmp/touched-notes.log`
and `/tmp/touched-digest.log`. **Treat this PR as unproven on that half until
that file says PASS.**

One note on what was driven against what: the binary that drove the scenarios
was built from this branch, and everything committed after it is test files, the
change entry, and (on #1357) manual Markdown. No line that any scenario asserts
on has moved since the drive.

santoshkumarradha:
## The trunk merge, and what change 1's re-drive means for this one

This branch now carries `origin/santos/dev2` at `b18b61993` (#1355, which made
the bash belt the default again and removed the worker-slot clamp) and the whole
of #1356. Merge commits on this branch, newest last:

- `ed8a05a2f` — `origin/santos/dev2` into `chat/plan-digest-on-speak`
- `a79d3cd64`, `45dc11155`, `68e8d63c9`, `85d650a50` — `chat/notes-are-a-channel`
  as change 1 kept moving under it

#1359 was checked against this change and does not reach it. `TaskNotice.PlanTask`
now names the store task in `planStoreID`'s spelling (`t-41`), and **nothing here
reads that field**: the digest, the `tasks` listing and the note writer all
resolve a row through `planTaskLabels` / `planTaskID` over `PlanTaskRow.ID`,
which is already the store id in that spelling.

The merge was proved by running the work rather than by the absence of conflict
markers: `make test-laws` green on this branch (every structural law, including
`internal/session` and `internal/tui3`), and the prompt-size law measured on
**this** tree.

## The bound and its law

`planDigestRows = 8` is the named constant, `planDigestLineChars = 90` and
`planDigestNoteChars = 160` bound the line and the note inside it, and the
overflow line is `… and %d more, which \`tasks\` lists.` No results, no steps,
no transcripts — one line per task, and only while a run is live.

## Prompt size, on this tree

`TestABeltWorkerReadsTheBeltsPagesAndNothingElse` PASSES. The belt worker's
system prompt is **16240 bytes of the 16384 budget, 144 of headroom**. The cap
in PERF.md was not touched — the fix was trimming my own prose plus two pieces
of existing prose the change had made redundant. 144 bytes is thin and the next
lane to add a sentence there will hit it.

## Change 1's acceptance, since this stacks on it

Scenario 1 was re-driven on the real binary under real concurrency after the
merge: five workers holding a task at the instant the note landed, the note
addressed to a row **chosen out of that live set**, and the marker that only the
note's file carries found in the addressed worker's record and in none of the
four others'. The full evidence, the two defects the concurrency test found, and
the four negative controls are on #1356. I also retract a claim I made earlier
there — that scenario 1 had already run with two workers live. It had not, and
#1356's comment says why in full.

santoshkumarradha:
## `touched packages` red: `TestNothingArrivingAfterTheStopIsDrawn`

The gate failed on `internal/tui3` with

```
--- FAIL: TestNothingArrivingAfterTheStopIsDrawn (0.45s)
    stopping_test.go:136: the frame moved after the stop:
FAIL	github.com/Agent-Field/codeaf/internal/tui3	332.890s
```

**It is not this change, and I checked rather than assumed.** Two things say so.

First, the arms. I built the package on a clean detached worktree of
`origin/santos/dev2` and on this branch, and ran that one test 21 times on each
across `GOMAXPROCS` 1, 2 and 4:

| tree | result |
| --- | --- |
| `origin/santos/dev2` (`55dd97fbf`, which already contains #1356) | 0 of 21 failed |
| `chat/plan-digest-on-speak` (`af889bf8b`) | 1 of 21 failed |

Then 60 more runs on this branch at `GOMAXPROCS=4`: no failure. So **1 in 81**
on this branch and 0 in 21 on the trunk, which are not distinguishable rates —
and the trunk arm carrying #1356 means change 1 does not cause it either.

Second, the diff. This branch's entire delta inside `internal/tui3` is a **doc
comment** on `taskPlanNoteSend`. Nothing in change 2 can move a frame: the test
drives the surface with synthetic `streamEventMsg` values and a `fakeAgent`, so
the digest — which lives in `Agent.Submit` — is never on the path.

What the test is exposed to is real time. `stoppingApp` sets
`a.turnBegan = a.now()` and `newTestApp` installs no clock, so `a.now()` is
`time.Now()`; the test captures a frame, drives five more messages, and compares.
Any element rendered from elapsed time can tick over between the two captures on
a loaded box, which is what "the frame moved after the stop" reads as. That is a
hypothesis I could not confirm — I hunted a second failure for 60 runs
specifically to diff the two frames and print the line that differs, and it did
not reproduce, so I am **not** claiming the mechanism, only that the rate is the
same with and without this change.

I have not added it to `.github/known-red.txt`: that ledger only shrinks, and a
1-in-81 flake is a bug report rather than a known shape. Raising it as one is the
right next step and I have not done it yet — say the word and I will file it with
this measurement as the evidence, or leave it to whoever owns `internal/tui3`.

## The trunk moved and is merged

#1356 landed as `55dd97fbf` and is now `santos/dev2`; the branch was deleted, so
this branch merged the trunk instead. The squash moved the surrounding context
and conflicted in three files — `internal/manual/chat_test.go`,
`internal/manual/chat/worker-harness.md`, `internal/session/tools_tasks.go` —
each an add-on-my-side-only where the trunk and the merge base were both empty at
that spot. Resolved by keeping this branch's additions, then verified that
**both** changes are present rather than trusting the resolution: change 1's
`## Does a note actually reach the worker…` section, its `When it is not handed
over at all` bullet and `planNoteFromChat` all survive from the trunk, and change
2's `## Does the chat know what is running while I talk to it` and `stopPlanRow`
survive from here. No duplicated headings.

Change 2's diff against the trunk is now only its own nine files, 623 insertions.
Head `2795fd37a`.

PR #1384, opened 2026-09-22
feat(tui): share randomized working-logo animations across chat and tasks

santoshkumarradha:
## What changed
Codex Image Sep 22, 2026, 02_58_05 PM

While codeaf answers or waits, a single-line chevron-and-dot animation appears immediately below the latest submitted question, after its wrapped text and attachments. A two- or three-word developer caption is composed from compatible action/object families, avoids the last eight opening choices, and rotates every ten seconds through a deck shuffled once per operation. No phrase repeats until the deck is exhausted. Dim ink separates the playful caption from factual work updates. Work details and the answer grow below it; the indicator scrolls with the question and disappears on completion. The input area has no added activity row.

The component reserves nine terminal columns for every pose and two columns before following content. With its two-column inset, its label starts at column 14 relative to the content area. The caption has its own 28-column reservation plus a two-column gap before future content. A slow decoding ripple briefly replaces one letter at a time with two symbols over 160 ms, resolves for 80 ms, and pauses for 1.8 seconds between passes. Every substitution is one terminal cell. Accessible tiers keep static text. Token counts and rates do not affect this layout. The top-left header carries a stationary `>● codeaf` wordmark.

Chat, task transcripts and adaptive-run pages share `tokens.WorkActivity`. It offers ten motions with 28 timed poses each, explicit named or random selection, and operation-local state. Steering and provider retries keep the turn's choice. Frames are read-only and use the existing clock; there are no extra timers or background painting.

The animation stops for questions, inactive or disconnected tasks, and copy mode. ASCII, monochrome, linear and small-window views retain compact text. Includes manual, design and performance documentation, plus the numbered change entry.

## Validation

- Focused chat/task lifecycle, harness, approval, header and composer geometry checks pass locally.
- Caption grammar, length, recent-history exclusion, render stability and decoding readability and fallback checks pass.
- All 280 poses preserve the label column and anchor below a wrapped question. Streaming grows below the anchor without adding input chrome.
- Inspected actual app frame output with the shared indicator and wordmark.
- `make build` produced macOS arm64 revision `f4e1d60c5be24b8117cfa462e3dffffe40dfc78e`.
- Final Spark acceptance passed as job `20260922-184855-000785` on that exact revision: `make pr-ready`, build, size and benchmark.

## Checklist

- [x] Numbered change entry included.
- [x] The manual knows about it.
- [x] No new known-red entries.
- [x] Only this change's paths are staged.


AbirAbbas:
Taking over this PR to get it merge-ready. I ran it locally: I built the binary and drove it in tmux against a stubbed streaming model. The chat indicator, caption rotation, decode ripple, streaming below the anchor, cleanup on completion and failure, and the narrow-window fallback all behave as described. I'm pushing a few small follow-ups on top as normal commits, with no history rewrites.

---
_Generated by [Claude Code](https://claude.ai/code)_

AbirAbbas:
Pushed two follow-ups:

- **441385b**: the ball's gold was one hard-coded hue, about 2:1 against a white page. Light pages now get a deeper gold (`#8C6420`, at least 4.5:1). Which gold to use is decided from the active ladder's ink, so a measured background is covered as well as the theme setting. I also removed `workLogoRows` and `roomWorkLogoRows`, which had no production caller; the tests now call `activityRows` directly. New test: `TestTheWorkingBallIsDeeperGoldOnALightPage`.
- **245ef45**: this branch made `TestTwoCorrectionsStayInTheOrderAndThePlacesTheyWereSaid` flaky. It counts `› ` rows, and some logo poses (`‹•›`, `› •`) spell the same chevron, so the result depended on the pose the clock landed on. It failed 15 of 150 isolated runs; with the activity row excluded from the count, 0 of 150.

Local `make pr-ready` is green apart from three `tui3` permission tests. They fail on `dev` too, because this box runs as root.

Not changed (design calls, left for the author): shimmer is suppressed across the whole screen while any logo is visible, and a visible logo keeps the frame clock at full cadence even though poses change every 100 ms.

---
_Generated by [Claude Code](https://claude.ai/code)_

AbirAbbas:
Live-tested this on a real model today (deepseek-v4-flash over OpenRouter, tmux, 150×42, engine road and `--no-host`). The feature holds up: the row lands under the question within 0.04 s, animates in place, rotates captions every 10 s without repeats, decodes one letter at a time, disappears on completion, survives steering and interrupts, hides behind an approval card, draws its own motion on the task page, and stays off on 16-colour, narrow and short terminals.

Two defects were the PR's own, and I pushed fixes for both on this branch (four commits, `a440e8ef0..c81ea2784`):

- **Copy mode froze the row instead of hiding it.** `enterCopy` and `freezeRoom` snapshotted the page before setting `copy.on`, so a half-decoded caption sat frozen for the whole copy session. Both doors now lay the page out with copy mode already owning it; leaving copy mode brings the live row back. Re-verified live: the frozen page shows the question, one blank, then the tool row.
- **The anchor was not turn-scoped.** `questionActivity` took the last user/steer entry of any turn, so a turn that starts without a question of its own (a standing firing, an adopted in-flight turn) would draw the row under the previous, already answered question. `turnAnchor()` now gates both `workLogoVisible` and the chat branch; a turn without its own question keeps the compact waiting text, the shimmer and the spinner-stride cadence. Task pages stay anchored to their request across retries, with the reason in a comment. Five contract tests in `worklogo_test.go`, plus one existing freeze test re-measured against the frozen page.

Also recorded in PERF.md: while the row is visible the surface keeps the full paint cadence, measured at about 2.1 % of one core during a 30 s tool wait against 1.0 % without the row (two paired runs), daemon unchanged. Not changed, just written down.

One thing I first flagged and then withdrew: on `step icons = plain` the row still draws `●˙•`. That is the geometric floor doing its job, not a defect.

`make pr-ready BASE=origin/dev` is green on the head.

AbirAbbas (approved):

PR #1387, opened 2026-09-22
feat: add total token usage to anonymous session telemetry

santoshkumarradha:
## Change
- Count provider-reported input and output tokens per completed session, including hosted sessions, without sending model names or content.
- Add `total_tokens` to the existing allowlisted `session_ended` event and update the notice, docs, and focused tests.
- A daily sum chart is already attached to the PostHog `codeaf · Engineering` dashboard, filtered to release builds (`channel` excluding `dev` and `unknown`). It will populate when this change ships and new events arrive.

## Verification
- `git diff --check` passed.
- PostHog accepted the daily aggregation query and the dashboard tile is present. Current daily values are zero because existing events do not have `total_tokens`.
- Could not run Go tests or `make pr-ready`: this environment has no Go toolchain. CI should validate the change before merging.


AbirAbbas:
Reviewed, ran locally, and pushed one follow-up commit (e34fd8f).

**Verified end to end.** Built `bin/codeaf` from this branch against a stub OpenAI-compatible server that reports 1000 prompt + 42 completion tokens per call, then read the spooled `session_ended` event:

| run | model calls | `total_tokens` |
|---|---|---|
| `codeaf exec` | 1 | 1042 |
| `codeaf chat --once` (hosted, the default) | 4 | 4168 |
| `codeaf chat --no-host --once` (in-process) | 4 | 4168 |

Both counting paths (the provider chokepoint in-process, and the `EventTurnDone` tee for hosted sessions) agree, include auxiliary calls such as the namer, and do not double-count. `make pr-ready` passes, including fresh tests for `cmd/codeaf`, `internal/provider`, `internal/telemetry`.

**Pushed in e34fd8f:**
- `README.md` still printed the old notice line (`mode (chat or task), how many sessions, how many errors`); it now matches the `Notice` constant.
- The chat manual (`running-from-the-terminal.md`, the `codeaf telemetry` section) now says `session_ended` carries `total_tokens`, so the chat does not leave it out when asked what codeaf sends.

**Left as is, for a decision:**
- `total_tokens` is the first exact integer on `session_ended`; every other value there is a band or a fixed word. It carries no content, but an exact number is more distinguishing than a band. Keeping it exact makes the dashboard sum useful, so I left it. Banding it like `cost_usd` would be the alternative.
- The two paths can diverge slightly on edge cases: in-process counts tokens from every recorded non-start call (including a failed call that still reported usage), while hosted counts only sealed turns, so a voided retry or an errored turn can count differently. This is small and not a blocker.

---
_Generated by [Claude Code](https://claude.ai/code)_

